PR #17, opened 2016-09-20
Copy share directories into Windows binary package.

jeroenooms:
I think this will fix your windows build. This copies both the `./share` dirs from R winlib into the installation directory so they are bundled with the R binary package.


edzer:
Bingo!! Thanks a lot, Jeroen - another milestone!

PR #30, opened 2016-10-14
Update README.md

Nowosad:
r-consortium logo fixed


edzer:
Doesn't look really nice, does it?

PR #31, opened 2016-10-18
Update DESCRIPTION

mpadge:


edzer:
This was not a typo; renaming the package is a bit more involved than that.

mpadge:
Sorry Edzer, I meant it more just as a prompt to circumvent the potentially confusing steps of having to load `sfr` in order to access the `sf` package

Nowosad:
@edzer - I also have s similar concern. Why the repo for sf package has a "sfr" name?

edzer:
The proposal was called "simple features for R", as is the project that carries it out. The package it produces is called "sf", and is in the project directory. I can rename the repo or package, but what is the concern when in a short while most people will install from CRAN?

PR #33, opened 2016-10-18
Vignette

Robinlovelace:
Adds a benchmark and reading of .shp to the vignette (sorry for 2 overlapping PRs).

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -8,3 +8,5 @@ TODO
 README.md
 ^windows
 ^appveyor\.yml$
+^.*\.Rproj$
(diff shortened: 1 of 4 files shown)


edzer:
Please fix the shapefile / kmz examples.

Robinlovelace:
> Please fix the shapefile / kmz examples.

Is this still outstanding? Seems Travis is passing now - now sure why it was failing on this PR.

They work on my computer. I can make eval=TRUE but that would make the vignette substantially slower - not sure that's worth it. Maybe I should use smaller example files?

PR #46, opened 2016-11-03
update gdal_read/CPL_get_layers

mpadge:
The current version of `CPL_get_layers` simply returns list of layers identified by the appropriate OGR driver regardless of whether or not the file actually contains these named layers. Passing in `st_read()` the name of a layer not present in the actual file generates a fairly uninformative error:
```
Error in eval(substitute(expr), envir, enclos) : Opening layer failed.
```
This pull request modifies `CPL_get_layers` so that `st_list` returns a list of only those layers actually present in the given file. The above error will then only be generated when requesting layers not given in `st_list`.

-----

Note#1: Printing an `sf_layers` class object results in these layers being numbered consecutively rather than by their OGR driver layer numbers, but as layers can't be called by number with `st_read` this ought not be an issue. Easy to insert appropriate layer numbers if you think they're necessary ...

Note#2: I've changed your formerly pre-allocated `Rcpp::List`s to dynamically resized ones modified by the admittedly inefficient `Rcpp::List.push_back()`. Numbers of layers can nevertheless only be an absolute maximum of 17, so this inefficiency can likely be ignored. An alternative would be to re-code them as STL objects typecast once filled to equivalent Rcpp forms.

Note#3: Travis fails not only on my build, but on your own builds in my fork, so i'm guessing that's an issue with environment variables which you've set for your builds? (`_R_CHECK_FORCE_SUGGESTS_`?)

### DIFF
--- a/src/gdal_read.cpp
+++ b/src/gdal_read.cpp
@@ -79,6 +79,7 @@ size_t count_features(OGRLayer *poLayer) {
 		if (n == INT_MAX)
 			throw std::out_of_range("Cannot read layer with more than MAX_INT features");
 	}
+    poLayer->ResetReading ();
 	return(n);
 }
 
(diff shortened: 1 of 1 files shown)


mpadge:
Yep, ignore Note#3, travis passes

edzer:
Mark, could you provide me with a (link to a) data source and an the appropriate `st_list` and `st_read` commands that generates this problem?

edzer:
I think that you are too much thinking from a particular type of file you're working with, and not from the GDAL API we want to interface. If I want to list tables in a spatial database, I also want to list those that have no records. Your PR would skip those. 

I propose the following; we modify `st_list` such that it reports the number of fields and features in each layer; if this is NA (-1 in GDAL) we add an option to st_lists that replaces the missing value with 

    nrow(st_read("dsn", "layer", quiet =TRUE))

meaning that every layer gets read -- I don't think this would be good to do by default. Would that solve your problem?

mpadge:
Hmm... I've since done a fresh install (including GDAL2.1.2) and can't reproduce the problem exactly, but the error is simply:
```
> tst <- sf::st_read (system.file ("shape/storms_xyz.shp", package="sf"), layer="anything") 
Cannot open layer points
Error in eval(substitute(expr), envir, enclos) : Opening layer failed.
```
Your demo files only have single layers, and this is only really an issue for files with multiple layers such an an OSM file extracted thus:
```
> bbox <- "(51.5,-0.1,51.52,-0.08)"
> key <- '[highway]'
> query <- paste0 ('(node', key, bbox, ';way', key, bbox, ';rel', key, bbox, ';')
> url_base <- 'http://overpass-api.de/api/interpreter?data='
> dat <- httr::GET (paste0 (url_base, query, ');(._;>;);out;'))
> doc_xml <- xml2::read_xml (httr::content (dat, as="text", encoding="UTF-8"))
> xml2::write_xml (doc_xml, file="export.osm")
```
Then,
```
> sf::st_list ("export.osm")
Driver: OSM 
Available layers:
        layer_name       geometry_type
1           points               Point
2            lines         Line String
3 multilinestrings   Multi Line String
4    multipolygons       Multi Polygon
5  other_relations Geometry Collection
```
Previously---I guess with older GDAL versions?---`layer_name` included `polygons`, and attempts to `st_read` those generated the error. There are likely to be other OGR drivers where this problem still exists, and it arises simply because `GDALDataset::GetLayerCount()` counts all layers recognised by the OSM driver whether or not data actually exists for those layers. 

Even though the error is no longer prompted with any of these `layer_name`s, because the OSM driver now recognises `GDALDataset::GetLayerByName()` for all of the above `layer_name`s, the PR still prevents the following:
```
> dat <- sf::st_read ("export.osm", layer="multilinestrings")
Reading layer `multilinestrings' from data source `export.osm' using driver `OSM'
features:       0
fields:         4
proj4string:    +proj=longlat +datum=WGS84 +no_defs 
```
and then
```
> dat
Simple feature collection with 0 features and 4 fields
bbox:           xmin: NA ymin: NA xmax: NA ymax: NA
epsg (SRID):    4326
proj4string:    +proj=longlat +datum=WGS84 +no_defs
[1] osm_id     name       type       other_tags geometry  
<0 rows> (or 0-length row.names)
```
which is not much use. With my PR `st_list` only returns those layers actually containing data:
```
> sf::st_list ("export.osm")
Driver: OSM 
Available layers:
        layer_name       geometry_type
1           points               Point
2            lines         Line String
3    multipolygons       Multi Polygon
```
I would say this still makes this version of `st_list` more useful than simply returning names of all layers recognised by the OGR driver. Even more useful would be to include a feature count for each `layer_name`, but feature counting can of course be quite slow, so i'll leave that decision to you ...

PR #59, opened 2016-11-07
Cleaning up crs handling code

rundel:
I originally wanted to make some changed to `st_transform` but got bogged down in the crs code. I found it hard to understand all the checks and various assumptions about crs's so I've gone in and tried to revise everything to make more sense (at least to me). 

I've tried not to be too opinionated and leave the original functionality as much as possible, the core changes are as follows:

* I removed `check_replace`, as far as I can tell we only care about the case where the crs is changed and I think the new logic `st_crs<-.sfc` is cleaner and covers the necessary case.

* Allowing lists as a valid argument for `st_crs<-.sfc` seemed weird to me, I opted to just allow for crs class objects and then added a helper function to handle the list -> crs conversion needed by st_sfc (which is the only use case I found for a list)

* Switched epsgFromProj4 to epsg_from_proj4string for the sake of consistency (it seems likely this function will be used again internally at some point)

* ~st_read currently returns 0 for epsg for some objects, I'm fairly confident this is equivalent to no epsg - to avoid handling this additional special case with the crs code I opted to set the epsg attribute to NA_integer_ in st_read~

* Moved handling of epsg = 0 back to `st_crs<-.sfc` since some of the geometry functions can also produce this

There were a surprisingly large number of edge cases for these relatively small changes, hopefully I caught them all.

### DIFF
--- a/R/crs.R
+++ b/R/crs.R
@@ -57,73 +57,77 @@ is.na.crs = function(x) { is.na(x$epsg) && is.na(x$proj4string) }
 #'  sfc
 #' @export
 `st_crs<-.sfc` = function(x, value) {
-	trim <- function (x) gsub("^\\s+|\\s+$", "", x)
-	# init:
-	if (!is.na(value) && !(is.numeric(value) || is.character(value) || is.list(value)))
-		stop("crs should be either numeric (epsg), character (proj4string), or list")
-	check_replace(x, value)
-	attr(x, "proj4string") = NA_character_
-	attr(x, "epsg") = NA_integer_
-	if (is.list(value)) { # try to get value from the attribute list:
-		if (!is.null(value$epsg) && !is.na(value$epsg))
-			value = value$epsg
-		else if (!is.null(value$proj4string) && !is.na(value$proj4string))
-			value = value$proj4string
-		else 
-			value = NA_integer_
-	}
-	if (! is.na(value)) {
-		if (is.numeric(value)) {
-			value = as.integer(value)
-			if (value == 0) {
-				attr(x, "epsg") = NA_integer_
-				value = NA_character_
-			} else {
-				attr(x, "epsg") = value
-				value = CPL_proj4string_from_epsg(value)
-			}
-		} else 
-			attr(x, "epsg") = epsgFromProj4(value)
+	trim = function(x) gsub("^\\s+|\\s+$", "", x)
+	
+	# Make sure crs exists
+	if (is.null(attr(x, "proj4string")))
+		attr(x, "proj4string") = NA_character_
+	if (is.null(attr(x, "epsg")))
+		attr(x, "epsg") = NA_integer_
+
+	attr(x, "proj4string") = trim(attr(x, "proj4string"))
+	start_crs = st_crs(x)
+	
+	if (is.na(value)) {
+		# Do nothing
+	} else if (is.numeric(value) && value != 0) { # epsg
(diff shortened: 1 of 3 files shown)


R/crs.R
```diff
@@ -57,73 +57,77 @@ is.na.crs = function(x) { is.na(x$epsg) && is.na(x$proj4string) }
...
-		return() # trying to replace proj4string with identical value
-	if (!is.na(epsg) || !is.na(proj4string))  # possibly warn:
-		warning("st_crs: replacing crs does not reproject data; use st_transform for that")
+
+crs_from_list = function(x) {
+	trim = function(x) gsub("^\\s+|\\s+$", "", x)
+
+	if (is.null(x$proj4string) && is.null(x$epgs)) {
+		structure(list(epsg = NA_integer_, proj4string = NA_character_), class = "crs")
+	} else if ( is.null(x$proj4string) && !is.null(x$epgs)) {
```
edzer: did you mix up this and the next if block?

rundel: You are correct, should be correct now

edzer:
I see this breaks at least the appveyor build, with messages like
```

>  sfc = st_sfc(st_point(c(0,0)), st_point(c(1,1)))
Error in if (attr(x, "epsg") == 0) attr(x, "epsg") = NA_integer_ : 
  missing value where TRUE/FALSE needed
Calls: st_sfc -> st_crs<- -> st_crs<-.sfc
Execution halted
```
indicating that `attr(x, "epsg")` is `NULL` in your `if`. Did you run `R cmd check` on your PR?

rundel:
My fix was clearly too hasty, the are a number of other issues that pop up as I try to fix things - this is a bit of a rabbit hole. 

I'm still looking and will try to come back with a more complete solution.

edzer:
I agree it is messy, and share your wish to clean up. Would a cleaner solution need a `crs` object as attribute of an `sfc`, similar to `sp`? @rsbivand opionions?

rsbivand:
I can't read the R files because of the roxygen markup, which in vi is displayed in bold magenta. Is there a way of displaying just the R code in R files? The markup is visually dominant and pure noise.

In addition, the issues action is far too noisy, hundreds of message this weekend, and people are not talking together, so individual views are getting implemented without considering what the overall questions are. This is an overall question, so we need a forum where everybody who is contributing can discuss it together. These design decisions are most often not orthogonal. This is where a central repo and a mailing list is a much better solution.

I can't help until I can free the R files from roxygen markup, and work with no markup in the R folder.

edzer:
```
for f in `ls *R`
do
 grep -v "^#'" $f > x
 mv x $f
done
```

rsbivand:
But if I commit push this to my github copy, it will mean that any changes I make subsequently and wish to contribute will be polluted with the roxygen deletions. How do you or others edit files visually suppressing roxygen but not changing those lines in the underlying files? I must assume consistency between local and remote, otherwise LITS.

mpadge:
@rsbivand [auto fold roxygen comments in vim](https://gist.github.com/wrathematics/3e1d08a70f7b0eb57a88)

rsbivand:
Thanks, works for me and helps!

edzer:
I think the crs handling is now much cleaner in https://github.com/edzer/sfr/commit/e4de7e4f8935ee1e37795378d8dc05dcba29b0bf

rundel:
Looks like a nice set of changes, a crs attribute looks like a more robust approach. I'll play with the new version and close this PR.

PR #90, opened 2016-12-02
fix CPL_read_ogr when layers have no features

mpadge:
This example generates a file with recognised GDAL layers yet no data. This causes a bad allocation [here](https://github.com/edzer/sfr/blob/master/src/gdal.cpp#L168) because `g.size()=0`, with fatal consequences. The following code reproduces using the GDAL OSM driver, first by downloading a `.osm` file.
```
> query <- paste0 ("[out:xml];(node[\"building\"](51.51,-0.12,51.52,-0.11);",         
                  "way[\"building\"](51.51,-0.12,51.52,-0.11);",                     
                  "relation[\"building\"](51.51,-0.12,51.52,-0.11);",                
                  ");(._;>);out body;")                                              
> base_url <- "http://overpass-api.de/api/interpreter"                                
> dat <- httr::POST (base_url, body=query)                                            
> doc_xml <- xml2::read_xml (httr::content (dat, as="text", encoding="UTF-8"))        
> xml2::write_xml (doc_xml, file="export.osm") 
```
Five layers exist:
```
> sf::st_list ("export.osm")
Driver: OSM 
Available layers:
        layer_name       geometry_type features fields
1           points               Point       NA     10
2            lines         Line String       NA      9
3 multilinestrings   Multi Line String       NA      4
4    multipolygons       Multi Polygon       NA     25
5  other_relations Geometry Collection       NA      4
```
But this is a bad alloc:
```
sf::st_read ("export.osm", layer="multilinestrings") 
```
because [these lines](https://github.com/edzer/sfr/blob/master/src/gdal_read.cpp#L183-L188) first generate `nd=-1`, then transformed to `n_d=0` and ultimately the value of `n=0`. 

The easy fix is these two lines [here](https://github.com/mpadge/sfr/blob/master/src/gdal_read.cpp#L188-L189), but I suggest that's not optimal, because `sf` is very likely to be used to automatically extract any and all data from various sources. This will be done by counting numbers of layers, then iterating over them. There are bound to be countless cases where layers are identified by a driver yet have no features, and throwing a `std::error` in all such cases will prevent automatic extraction.

My suggested changes to `read.R` simply result in `NULL` being returned, along with the message on L60. Note that I can't check the travis build, but can confirm that this passes all tests.


edzer:
Nice catch! A few observations:
*  `x = st_read("/tmp/export.osm", layer = "multilinestrings", quiet = TRUE)`
returns
```
[1] osm_id     name       type       other_tags
<0 rows> (or 0-length row.names)
```
where `quiet=FALSE` (default) returns `NULL`; this is not consistent. 
* The `NULL` is returned by `message`, which doesn't make the intention of the code clear. 
* The message `Layer has no data` is not precise: it has no features, but it has field names.

I suggest that in line with the `read.table` family, in case of no features an `sf` object is returned with 0 rows.

mpadge:
Yeah, the inconsistent returns were my fault - that's fixed. And I agree with the rest, but must admit I can't figure out how to make an `sf` object with no rows because, for example,
```
> ch0 <- character (0) # void filler
> x <- data.frame ('a'=ch0, 'b'=ch0, 'c'=ch0, 'd'=ch0, 'geometry'=NULL) # 'geometry' of course does nothing there
> x [['geometry']] <- st_sfc (st_polygon (list ()))
Error in `[[<-.data.frame`(`*tmp*`, "geometry", value = list(list())) : 
  replacement has 1 row, data has 0
```
This happens because `sfc` objects are constructed as lists which have an implicit length of 1, and this is translated by `<-.data.frame` to 1 row, while `x` has 0 rows. In other words: `sf` objects are constructed by filling the geometry column of a `data.frame` with an `sfc` object. The `data.frame` can have zero rows, and the `sfc` object can be empty, but it nevertheless contains data and putting it in an empty `data.frame` is equivalent to attempting to fill the row of a zero-row `data.frame` with some non-NULL value. 

Either

1. Any easy suggestions? 

2. or, would you be okay with the ugly solution of filling the `sf df` with `NA` instead of `NULL` so the `sfc` geometry can be stored?

edzer:
after you crated data.frame x:
```
x$geom = st_sfc()
st_sf(x)
Simple feature collection with 0 features and 4 fields
bbox:           xmin: NA ymin: NA xmax: NA ymax: NA
epsg (SRID):    NA
proj4string:    NA
[1] a    b    c    d    geom
<0 rows> (or 0-length row.names)
```

mpadge:
Thanks Edzer, that was what I needed. All done now, so
```
> x <- sf::st_read ("export.osm", layer="multilinestrings")
Reading layer `multilinestrings' from data source `.../export.osm' using driver `OSM'
Layer has no features
Simple feature collection with 0 features and 4 fields
bbox:           xmin: NA ymin: NA xmax: NA ymax: NA
epsg (SRID):    NA
proj4string:    NA

> x
Simple feature collection with 0 features and 4 fields
bbox:           xmin: NA ymin: NA xmax: NA ymax: NA
epsg (SRID):    NA
proj4string:    NA
[1] osm_id     name       type       other_tags geometry  
<0 rows> (or 0-length row.names)
```
... and all info if there! 

You'll note, however, that I had to resort to `grep`-ing the GDAL output to find where the `geometry` was located. I can't see any way to avoid this, because the direct GDAL output is all you've got to go on. This nevertheless does always have a ''geometry'' component, so *should* be failsafe. Just in case, I've put a `stop` in there, but that can't be tested and so will irrevocably lower your codecov - maybe delete it?

edzer:
Thanks Mark! I think this fix fixes it at the place where it should; pls test; I see
```
> sf::st_read ("export.osm", layer="multilinestrings") 
Reading layer `multilinestrings' from data source `/tmp/export.osm' using driver `OSM'
Simple feature collection with 0 features and 4 fields
bbox:           xmin: NA ymin: NA xmax: NA ymax: NA
epsg (SRID):    NA
proj4string:    NA
```

mpadge:
Great solution Edzer - all good.

PR #102, opened 2016-12-06
st_read_db and st_write_db now handle and preserve the srid correctly

bleutner:
Hi, 

`st_read_db` and `st_write_db` are currently losing the projection information when transferring from/to a database.

Most changes needed to fix this are minor. However, in order to make it work for the WKB transfer mode, I had to use EWKB instead of plain WKB (line 115). I am unsure whether this could introduce a portability problem to DBs other than PostGIS. What do you think?

An example:
With the current version of `sf` I get:
```r
data(meuse)
sf <- st_as_sf(meuse, coords = c("x", "y"), crs = 28992)
cr <- st_crs(sf)
cr
#>$epsg
#>[1] 28992
#>
#>$proj4string
#>[1] "+proj=sterea +lat_0=52.15616055555555 +lon_0=5.38763888888889 +k=0.9999079 +x_0=155000 +y_0=463000 +ellps=bessel +towgs84=565.4171,50.3319,465.5524,-0.398957388243134,0.343987817378283,-1.87740163998045,4.0725 +units=m +no_defs"
#>
#>attr(,"class")
#>[1] "crs"

## Export to DB; with projection
st_write_db(con, sf, "meuse_tbl_txt", binary = FALSE)
st_write_db(con, sf, "meuse_tbl_bin", binary = TRUE)

## Import from DB; projection is gone
st_crs(st_read_db(con, "meuse_tbl_txt"))
#>$epsg
#>[1] NA
#>
#>$proj4string
#>[1] NA
#>
#>attr(,"class")
#>[1] "crs"

st_crs(st_read_db(con, "meuse_tbl_bin"))
#>$epsg
#>[1] NA
#>
#>$proj4string
#>[1] NA
#>
#>attr(,"class")
#>[1] "crs"

```

With my proposed patch it works as it should:
```r
## Export to DB; with projection
st_write_db(con, sf, "meuse_tbl_txt", binary = FALSE)
st_write_db(con, sf, "meuse_tbl_bin", binary = TRUE)

## Import from DB; projection is gone
crt <- st_crs(st_read_db(con, "meuse_tbl_txt"))
crb <- st_crs(st_read_db(con, "meuse_tbl_bin"))
all.equal(cr, crt)
#>[1] TRUE
all.equal(cr, crb)
#>[1] TRUE
crt
#>$epsg
#>[1] 28992
#>
#>$proj4string
#>[1] "+proj=sterea +lat_0=52.15616055555555 +lon_0=5.38763888888889 +k=0.9999079 +x_0=155000 +y_0=463000 +ellps=bessel +towgs84=565.4171,50.3319,465.5524,-0.398957388243134,0.343987817378283,-1.87740163998045,4.0725 +units=m +no_defs"
#>
#>attr(,"class")
#>[1] "crs"
```

### DIFF
--- a/R/db.R
+++ b/R/db.R
@@ -37,13 +37,13 @@ st_read_db = function(conn = NULL, table, query = paste("select * from ", table,
 			SRID = dbGetQuery(conn, paste0("select srid from geometry_columns where f_table_name = '", table, "';"))[[1]]
 			ret = dbGetQuery(conn, paste("select proj4text from spatial_ref_sys where srid =", SRID, ";"))
 			if (nrow(ret))
-				SRID
+				structure(list(epsg=SRID, proj4string=ret[[1]]), class = "crs")
 			else
 				NA_crs_
 		}
 	if (missing(EWKB))
 		EWKB = inherits(conn, "PostgreSQLConnection") || inherits(conn, "PqConnection")
-    tbl[[geom_column]] = st_as_sfc(structure(tbl[[geom_column]], class = "WKB"), EWKB = EWKB, crs = SRID)
+    tbl[[geom_column]] = st_as_sfc(structure(tbl[[geom_column]], class = "WKB"), EWKB = EWKB, crs = crs)
 	st_as_sf(tbl, ...)
 }
 
--- a/R/db.R
+++ b/R/db.R
@@ -37,13 +37,13 @@ st_read_db = function(conn = NULL, table, query = paste("select * from ", table,
 			SRID = dbGetQuery(conn, paste0("select srid from geometry_columns where f_table_name = '", table, "';"))[[1]]
 			ret = dbGetQuery(conn, paste("select proj4text from spatial_ref_sys where srid =", SRID, ";"))
 			if (nrow(ret))
-				ret[[1]]
+				SRID
 			else
 				NA_crs_
 		}
 	if (missing(EWKB))
 		EWKB = inherits(conn, "PostgreSQLConnection") || inherits(conn, "PqConnection")
-    tbl[[geom_column]] = st_as_sfc(structure(tbl[[geom_column]], class = "WKB"), EWKB = EWKB, crs = crs)
+    tbl[[geom_column]] = st_as_sfc(structure(tbl[[geom_column]], class = "WKB"), EWKB = EWKB, crs = SRID)
 	st_as_sf(tbl, ...)
 }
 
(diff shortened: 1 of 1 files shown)


R/db.R
```diff
@@ -37,13 +37,13 @@ st_read_db = function(conn = NULL, table, query = paste("select * from ", table,
 			SRID = dbGetQuery(conn, paste0("select srid from geometry_columns where f_table_name = '", table, "';"))[[1]]
 			ret = dbGetQuery(conn, paste("select proj4text from spatial_ref_sys where srid =", SRID, ";"))
 			if (nrow(ret))
-				ret[[1]]
+				SRID
```
edzer: I returned `ret[[1]]` here because SRID may well be not recognized as EPSG code: if you add a table with a SRS that is not recognized by PostGIS, it adds it to its spatial_ref_sys table, and creates a new SRID. Passing on this value will lead to errors later on.

bleutner: Ah, I wasn't aware of that and thought the EPSG would always be properly resolved.

I went this route because it would give me a complete "crs" with both EPSG and proj4string.
Would the newly committed version with constructing a complete "crs" beforehand be better then?

On the other hand, writing to the DB is currently only based on the SRID. Should this also be based on the proj4string instead then?

edzer: try returning `make_crs(ret[[1]])`, this should work in both cases (GDAL will try to resolve epsg from the proj4string, but leave it NA if it cannot find it). Also try the conditional example in st_read_db for an example.

bleutner:
Wow, already merged. You're fast.

A potential problem I still see is that during uploading to the db the projection is based on the EPSG/SRID only. If I see correctly (I'm new to postgis), then it cannot be based on the proj4string.
This means, however, that if I download the data to R, fiddle with it and upload again a few times, I will lose all projection information in cases where GDAL cannot find/define an EPSG code from the proj4string.

An example:
```r   
## 1st upload to db 
example(meuse, ask = FALSE, echo = FALSE)
meuse <- st_as_sf(meuse)
st_write_db(con, meuse, "meuse_tbl_txt", binary = T)
    
## 1st download from db
meuse2 <- st_read_db(con, "meuse_tbl_txt")
st_crs(meuse2)   # EPSG is NA, proj4 is correct
#>$epsg
#>[1] NA
#>
#>$proj4string
#>[1] "+proj=sterea +lat_0=52.15616055555555 +lon_0=5.38763888888889 +k=0.9999079 +x_0=155000 +y_0=463000 +ellps=bessel +towgs84=565.4171,50.3319,465.5524,-0.398957,0.343988,-1.8774,4.0725 +units=m +no_defs"
#>
#>attr(,"class")
#>[1] "crs"

## 2nd upload
dbRemoveTable(con, "meuse_tbl_txt")
st_write_db(con, meuse2[,-1], "meuse_tbl_txt",  binary = T)

## 2nd download -> proj4string lost
st_crs(st_read_db(con, "meuse_tbl_txt")) # Both epsg and proj4 are lost (NA)
#>    $epsg
#>[1] NA
#>
#>$proj4string
#>[1] NA
#>
#>attr(,"class")
#>[1] "crs"
```

etiennebr:
That wouldn't be easy, I guess. If I'm not mistaken postgis does not support the proj4 string directly. The only way I see it could work would be to insert a new srid row to `spatial_ref_sys` and fetch it on write, but it seems complex. There should be a warning when `!is.na(st_crs(geom))` and `SRID` is set to `0`.

PR #108, opened 2016-12-10
Try adding n_points

Robinlovelace:
Almost certainly not the best way to do this but I think allowing the user to specify `n_points` (could also be called `n` or, following `sample`, `size`) will be useful for many applications, not least my own. 

Let me know if it's deliberately excluded or of ideas how to fix this hacky 1st try.

Note: I'm getting this error with `st_lines_sample()`:

```
st_line_sample(st_transform(ls, 3857), density = 1/1000) # one per km
Error in st_length(x) : could not find function "CPL_length"
```

Any ideas?

### DIFF
--- a/R/geom.R
+++ b/R/geom.R
@@ -363,7 +363,7 @@ st_sym_difference = function(x, y) {
 st_line_sample = function(x, density = NA, type = "regular", n_points = NA) {
 	if (isTRUE(st_is_longlat(x)))
 		stop("st_line_sample for longitude/latitude not supported")
-  invalid_args = !is.na(n_points) & !length(density) > 1
+  invalid_args = !is.na(n_points) & length(density) > 1
   if(invalid_args)
     stop("Error: provide either density or n_points arguments but not both")
   l = st_length(x)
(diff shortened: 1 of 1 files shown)


R/geom.R
```diff
@@ -358,11 +359,17 @@ st_sym_difference = function(x, y) {
...
 #' st_line_sample(st_transform(ls, 3857), density = 1/1000) # one per km
 #' st_line_sample(st_transform(ls, 3857), density = c(1/1000, 1/10000)) # one per km, one per 10 km
-st_line_sample = function(x, density, type = "regular") {
+#' st_line_sample(st_transform(ls, 3857), n_points = 5) # one per km, one per 10 km
+st_line_sample = function(x, density = NA, type = "regular", n_points = NA) {
 	if (isTRUE(st_is_longlat(x)))
 		stop("st_line_sample for longitude/latitude not supported")
-	l = st_length(x)
-	n = round(rep(density, length.out = length(l)) * l)
+  invalid_args = !is.na(n_points) & length(density) > 1
```
edzer: This doesn't catch the case where n_points is given and length(density) == 1

edzer:
Thanks @Robinlovelace! Pls test; the error you got shouldn't happen -- recompile?

PR #113, opened 2016-12-12
Add examples showing how n works in st_line_sample

Robinlovelace:
I think if `length(n) == 1` the *total* sample size should be `n`, not `n * length(x)`.

To implement this is started writing:

```r
	  if(length(n) == 1)
	    n = n * l / sum(l)
```

But then was going to add another conditional in case rounding `n` caused problems but then realised it was getting a little complex! Any ideas? It will definitely useful to be able to set n and then distribute them according to line length. 

Happy to say it works now and this is sufficient to solve my problem + like that it gives the user low level access to the number of lines sampled per line.

### DIFF
--- a/R/geom.R
+++ b/R/geom.R
@@ -368,6 +368,8 @@ st_sym_difference = function(x, y) {
 #' ls = st_sfc(st_linestring(rbind(c(0,0),c(0,1))),
 #'   st_linestring(rbind(c(0,0),c(.1,0))), crs = 4326) 
 #' try(st_line_sample(ls, density = 1/1000)) # error
+#' st_line_sample(st_transform(ls, 3857), n = 5) # five points for each line
+#' st_line_sample(st_transform(ls, 3857), n = c(1, 3)) # one and three points
 #' st_line_sample(st_transform(ls, 3857), density = 1/1000) # one per km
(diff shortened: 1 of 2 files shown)


Robinlovelace:
BTW this is what that function to sample from the lines could look like if you don't get the right sample size 1st time:

```r
  if(sum(n_vec) != n) {
    n_diff = n  - sum(n_vec)
    probs = n_vec / n - l_lengths_rel
    probs[probs < 0] = 0
    sel = sample(length(l), size = abs(n_diff), prob = probs)
    n_vec[sel] = n_vec[sel] + sign(n_diff)
  }
```

Robinlovelace:
BTW II I've created and tested a function that could be useful for this:

```r
n_sample_length = function(n, l_lengths, weights) {
  # generate length-adjusted weights equal to 1
  l_lengths_rel = l_lengths * weights / (sum(l_lengths * weights)) 
  
  n_vec = round(n * l_lengths_rel)
  if(sum(n_vec) != n) {
    n_diff = n  - sum(n_vec)
    probs = abs((n_vec / n) - l_lengths_rel)
    sel = sample(length(l), size = abs(n_diff), prob = probs)
    n_vec[sel] = n_vec[sel] + sign(n_diff)
  }
  n_vec
}
```

Issue #125, opened 2016-12-20
Failure: guess_driver works (@test_read.R#25)


etiennebr:
Lots of things going on when running `devtools::test()`! I'd like to expose the `ext_map` and `prefix_map` and I think this would facilitate testing the driver support / matching. 

`ext_map`is useful to e.g. scan directories and try to guess which files might be readable.

PR #127, opened 2016-12-21
Tests drivers

etiennebr:
This is a larger p/r, @edzer tell me what you think, especially about  85b08e5, which might be more equivocal. The other commits are mostly adding `quiet = TRUE` to read and write, and catch outputs.

I could also merge all commits into one, or split them apart.

In the current state, this commits breaks the tests. I expect to make some changes before integrating in master. 

- `CPL_geos_is_valid works (@test_geos.R#17) - Self-intersection at or near point 0.75 0` here I'm not sure what to do. `CPL_geos_is_valid` returns a warning in addition to `FALSE`. Would returning `FALSE` be sufficient, and leave the warning to the caller ? I haven't looked into it really, so I'm not sure what it does.
- `Failure: st_crs works (@test_crs.R#20)` this one is about precedence. What warning should epsg:0 return ? `GDAL Error 1: Failed to lookup UOM CODE 0` or `GDAL Error 6: EPSG PCS/GCS code 0 not found in EPSG support files.  Is this a valid EPSG coordinate system?`. The former is a bit cryptic for my tastes, so I would go for the second one if possible.


edzer:
Regarding the warnings from GDAL: quite frankly, I'm happy with any relevant warning / error shown right now. At the C++ level we can make them silent (push/pop error handlers before/after), I don't think we can manipulate them at the R level; lower priority?

etiennebr:
I agree on GDAL, let's keep it that way if it's complex. It's just that I was expecting the same error for epsg:0 than epsg:1000. 

Regarding the lists, is it possible to expose the lists with `sf::prefix_map`? I'll give it a try. Indeed, it could be a straight list. I think my original intent was to add an option, I feel like a function offers more flexibility for future ideas, but I don't have a specific reason.

R/read.R
```diff
@@ -180,61 +180,92 @@ st_list = function(dsn, options = character(0), do_count = FALSE) {
...
+
+guess_driver_can_write = function(dns, drv = guess_driver(dns)) {
+  if(is.na(drv)) {
+    stop("Could not guess driver for ", dns, call. = FALSE)
+  }
+  if(!is_driver_available(drv)) {
+    stop(paste(unlist(drv), " not available in supported drivers, see `st_drivers()'"), call. = FALSE)
+  }
+  if(!is_driver_can(drv, operation="write")) {
+    stop(paste("Driver ", drv, " cannot `", operation, "`, see `st_drivers()'"), call. = FALSE)
```
edzer: see travis logs: operation does not exist as a local object, here. Did you cover this with a test? (I want to maintain 100% coverage of the R code)

etiennebr: I'll look into it, I agree on code coverage. R checks passes, at least locally, I'll look at the Travis log. I wasn't sure we could safely rely on minimal driver support, that's why there is no specific test for it. I could add a test relying on `shp` support since it is used implicitly to load data for different tests. I think it's a reasonable assumption, and no `shp` support would mean many other problems.

edzer: travis breaks on the missing .Rd file; you need to commit changes to them too.

etiennebr:
Forgot to mention it fixes #125

edzer:
Looks like there's still an error on [windows](https://ci.appveyor.com/project/edzer/sfr/build/1.0.369#L882).

etiennebr:
Beware, `c:/` is a drive, not a driver ! I sometimes get confused.

etiennebr:
I believe it happens in read.R#L188 when `guess_driver()` fails to match extension, but why does it fail on CRAN, but passes on appveyor? Seem to be related to local configuration as the tempfile() is generated locally and is unique. Do you have any strategy for debugging on CRAN? 

My hypothesis would be that `tools::file_ext` fails to parse `D:\temp\Rtmpqg1fuW\file287051112926.shp` because backslashes are often a mess.

PR #130, opened 2016-12-21
import Spatial from sp

### LINKED ISSUE #129: `setAs()` has no definition for class `Spatial`
When sourcing `sp.R` which contains `setAs("Spatial", "sf", ...)`, `setAs("Spatial", "sfc", ...)`, `setAs("sf", "Spatial", ...)` and `setAs("sfc", "Spatial", ...)`. 

```
in method for ‘coerce’ with signature ‘"Spatial","sf"’: no definition for class “Spatial”
in method for ‘coerce’ with signature ‘"Spatial","sfc"’: no definition for class “Spatial”
in method for ‘coerce’ with signature ‘"sf","Spatial"’: no definition for class “Spatial”
in method for ‘coerce’ with signature ‘"sfc","Spatial"’: no definition for class “Spatial”
```
It complains on `devtools::load_all(".")` and `devtools::document(roclets=c('rd', 'collate', 'namespace'))`, which both source these functions (well, all `.R` files).

etiennebr:
fix #129


etiennebr:
This is more complex then I thought, because it requires to move `sp` from `Suggests` to `Imports`. There's probably a way that I can't think of, to keep `sp` in the `Suggests`, but still offer `setAs` methods (and namespace) when the package is available.

PR #135, opened 2016-12-27
rough st_read1 prototype

mdsumner:
Unexported st_read1 function to read just one feature, the iFeature by default is 1 (the first), this is converted to the OGR 0-based index. An error is triggered if iFeature  exceeds the feature count.


edzer:
Closing, as this was a proof-of-concept only.

Issue #131, opened 2016-12-22
Error on st_union() in summarise()


etiennebr:
`st_union()` seem to work fine, but in `summarise()` it creates a weird geometry that it can't read. I tried to dump the geometry to wkt (quite large, so I won't put it here), and it can display on [wicket](http://arthur-e.github.io/Wicket/sandbox-gmaps3.html), but vertex order is slightly wrong. I think it might be how it handled the multi parts polygons.

```r
library(dplyr)
library(tibble)

nc <- st_read(system.file("shape/nc.shp", package="sf"))

nc %>%
  st_area() %>%
  sum
# 127031757146 m^2

as(nc$geometry, "Spatial") %>% 
  rgeos::gUnionCascaded(nc$SID79) %>%
  st_as_sf() %>% 
  st_area %>% 
  sum
# 127031757076 m^2

x <- lapply(unique(nc$SID79), function(i) {
  nc %>%
    dplyr::filter(SID79 == i) %>% 
    summarise(geog = st_union(geometry), area = st_area(geog))
})
# It would be nice to have something like row_bind(x) %>% st_area() %>% sum()
sum(sapply(x, function(x) x[["area"]]))
# [1] 127031757146

nc %>%
  ungroup() %>%
  summarise(geog = st_union(geometry), area = st_area(geog))
#                             geog             area
# 1 MULTIPOLYGON(((-76.54427337... 127031757146 m^2
```

But it fails with summarise
```r
nc %>%
  group_by(SID79) %>% 
  summarise(geog = st_union(geometry), area = st_area(geog))
# Error in summarise_impl(.data, dots) : 
# no applicable method for 'st_geometry' applied to an object of class "list" 
```

To dump the geometry: 
```r
nc %>% as_tibble %>%
  group_by(SID79) %>% 
  summarise(geog = st_union(geometry)) %$%
  st_as_text(geog) %>%
  cat(file="geog.wkt")
```

etiennebr:
I see, this is taken care of in [`summarise_.sf`](https://github.com/edzer/sfr/blob/master/R/tidy.R#L113).

edzer:
If you 

    nc %>% as_tibble %>% st_as_sf %>% class()
    # [1] "sf"         "tbl_df"     "tbl"        "data.frame"

you get an object of class `sf`, `tbl_df` etc. Is that a good idea? Does that help for anything?

etiennebr:
Seems to be related to the geometries.

Works :
```r
nc %>% 
[...]

PR #139, opened 2017-01-02
Coerce sfc

etiennebr:
This adds geometry coercion to `st_sfc()`. Summary:

* Removed sfr.R#63 (`attr(lst, "classes") = sapply(lst, class)[2L,] # Rcpp forces me to do this. Or is it me, allowing a mix?`); doesn't seem to break any test, but might have some effect somewhere I'm not aware. I just couldn't find the purpose, so I guessed it might be dead wood.
* Had to carry some attributes, I'm just unsure if I got them all or the correct ones. There's an `idx` attribute from `st_intersect` and I had to carry it over for the `st_interpolate_aw` example.
* `st_multi()` is named for consistency with PostGIS, although it might have differences in execution, it seemed like it had the same intent.
* If you could have particular scrutiny for the `cls == "GEOMETRYCOLLECTION"` part in `st_multi()` as I'm not sure if other cases could come up.
* I did not expose `st_multi()` as it's setters are incomplete and it's better to use `st_sfc()` (at least for now).
* Removed the warning expectation in `tests/cast.R#30` because `st_sfc()` does the coercion automatically. 

Related to #131


edzer:
I don't see which practical problem this solves. A problem it introduces is that it is no longer possible to create an `sfc` object that has a mix of geometry types:
```
> class(st_sfc(st_point(c(0,1)), st_linestring(matrix(1:10,,2))))
[1] "sfc_GEOMETRYCOLLECTION" "sfc" 
```
 We now have a working `st_cast` to change between the types, at all levels (`sfg`, `sfc`, `sf`). `GEOMETRYCOLLECTION` `sfc`s are not structurally easier to deal with than `GEOMETRY` `sfc`s.  What is the advantage over `st_cast`? Why not make all this optional?

edzer:
My mistake that I didn't update the docs, sorry about that. `st_read` by default coerces to multi in case of a mix, but can be turned off. `st_union`, `st_intersection` and `st_[sym_]difference` return mixes. We might equip those with a default to cast to multi, and it as an option in `st_sfc`.

`st_cast()` without arguments tries to `st_multi`, but not to force into `GEOMETRYCOLLECTION`:
```
> st_sfc(st_point(c(0,1)), st_linestring(matrix(1:10,,2))) %>% st_cast() %>% class()
[1] "sfc_GEOMETRY" "sfc"         
```

etiennebr:
Ok, no worries. I agree that casting might be more useful. I got confused by the doc and the coerce_multi() function that was commented just below. However I can't figure what you'd want from 
```
st_sfc(st_point(c(0,1)), st_linestring(matrix(1:10,,2))) %>% st_cast() %>% class()
```
I guess with S3, the best strategy is to check attributes before performing an operation, then.

PR #146, opened 2017-01-04
rbind.sf doesn't preserve the crs of its input

faridcher:


R/sf.R
```diff
@@ -249,7 +249,7 @@ print.sf = function(x, ..., n =
 #' @export
 rbind.sf = function(..., deparse.level = 1) {
 	ret = base::rbind.data.frame(...)
-	st_geometry(ret) = do.call(st_sfc, st_geometry(ret))
+	st_geometry(ret) = do.call(st_sfc, c(st_geometry(ret),crs=st_crs(st_geometry(ret))$epsg))
```
edzer: using `crs = st_crs(reg)` makes it more compact and more general.

faridcher: `crs = st_crs(reg)`  doesn't work. throws another error. `crs = st_crs(reg)$epsg` works fine.

edzer: I meant `crs = st_crs(ret)`.

faridcher: I see. that's a typo. `st_sfc` is supposed to received class type `crs` which is evident from `inherits(crs, "crs")`. But this line `u = unique(sapply(lst, class)[1,])` raises an error!

faridcher: try is with `nc` dataset:
```
nc <- st_read(system.file("shape/nc.shp", package="sf"), "nc", crs = 4267)
ret = base::rbind.data.frame(nc)
st_geometry(ret) = do.call(st_sfc, c(st_geometry(ret),crs=st_crs(ret)))
```

edzer: replace `c()` with `list()`.

Issue #154, opened 2017-01-06
st_cast GC loses CRS


mdsumner:
I need clarification before a PR for this. 

```R
library(sf)
example(st_read)

## two examples where the GEOMETRYCOLLECTION casting drops CRS
st_cast(st_triangulate(nc))
st_cast(st_cast(nc$geometry, "GEOMETRYCOLLECTION"))
```

Don't we also need to carry through the precision as well as the crs of the sfc? 

The return values are kind of idiomatic and sparse, and there's several places that will require precision re-added as well. (Or is precision to be explicitly dropped here due to the unwrapping?)

etiennebr:
I'll try to fix this in #157 because I get:
```r
st_cast(st_triangulate(nc))
# Geometry set for 100 features 
# geometry type:  GEOMETRYCOLLECTION
# dimension:      XY
# bbox:           xmin: -84.32385 ymin: 33.88199 xmax: -75.45698 ymax: 36.58965
# epsg (SRID):    4267
# proj4string:    +proj=longlat +datum=NAD27 +no_defs
# First 5 geometries:
# GEOMETRYCOLLECTION(POLYGON((-81.7027969360352 3...
# GEOMETRYCOLLECTION(POLYGON((-81.3452987670898 3...
# GEOMETRYCOLLECTION(POLYGON((-80.9657745361328 3...
# GEOMETRYCOLLECTION(POLYGON((-76.3302536010742 3...
# GEOMETRYCOLLECTION(POLYGON((-77.8988571166992 3...
# Warning message:
# In st_triangulate(nc) :
#   st_triangulate does not correctly triangulate longitude/latitude data.
st_cast(st_cast(nc$geometry, "GEOMETRYCOLLECTION"))
# Geometry set for 100 features 
# geometry type:  GEOMETRYCOLLECTION
# dimension:      XY
# bbox:           xmin: -84.32385 ymin: 33.88199 xmax: -75.45698 ymax: 36.58965
# epsg (SRID):    4267
# proj4string:    +proj=longlat +datum=NAD27 +no_defs
# First 5 geometries:
# GEOMETRYCOLLECTION(MULTIPOLYGON(((-81.472755432...
# GEOMETRYCOLLECTION(MULTIPOLYGON(((-81.239891052...
# GEOMETRYCOLLECTION(MULTIPOLYGON(((-80.456344604...
# GEOMETRYCOLLECTION(MULTIPOLYGON(((-76.008972167...
[...]

Issue #144, opened 2017-01-03
Improve st_cast() guesswork


etiennebr:
As discussed in #139, I would expect `st_cast_default()` to coerce geometry to MULTI* when presented a mix of types: 
```r
st_sfc(st_point(c(0,1)), st_multipoint(matrix(1:10,,2))) %>% st_cast() %>% class()
# [1] "sfc_GEOMETRY" "sfc"  
```
Would it make sense to make this the default:
```r
st_sfc(st_point(c(0,1)), st_multipoint(matrix(1:10,,2))) %>% st_cast("MULTIPOINT") %>% class()
# [1] "sfc_MULTIPOINT" "sfc"  

```
I'd be happy to integrate the #139 (`st_multi()`) logic :) to `st_cast()` as I think it would be convenient, but I might overlook something, or maybe there's another function for that purpose ?

edzer:
Yes, that makes sense.

etiennebr:
This might be in conflict with #110, #138. I think there should be a conservative way to `st_cast` that would pop an error when going to a higher dimension point > line > polygon and only allow polygon > line > point.

Issue #87, opened 2016-11-28
group_by - a problem with mixed POLYGON/MULTIPOLYGON object


Nowosad:
`group_by` works well with groups of more than one elements (all of elements are MULTIPOLYGON then):

```R
library(sf)
library(tidyverse)
nc_1 <- st_read(system.file("shape/nc.shp", package="sf"), "nc", crs = 4267)
nc_1$group <- c(rep('one', 50), rep('two', 50))
nc_1 <- nc_1 %>% 
        group_by(group) %>% 
        summarise(sum_area=sum(AREA), nr=n()) 
plot(nc_1)
st_write(nc_1, 'nc_1.gpkg', overwrite=TRUE)
```

It also works well with one-element groups  (all of elements are POLYGON):

```R
nc_2 <- st_read(system.file("shape/nc.shp", package="sf"), "nc", crs = 4267)
nc_2$group <- c(rep('one', 50), rep('two', 50))
nc_2 <- nc_2[c(1, 100), ]
nc_2 <- nc_2 %>% 
        group_by(group) %>% 
        summarise(sum_area=sum(AREA), nr=n()) 
plot(nc_2)
st_write(nc_2, 'nc_2.gpkg', overwrite=TRUE)
```

However, the result goes wrong when `group_by` and `summarise` create object with mixed POLYGON/MULTIPOLYGON elements:

```R
nc_3 <- st_read(system.file("shape/nc.shp", package="sf"), "nc", crs = 4267)
nc_3$group <- c('one', rep('two', 49), rep('three', 50))
nc_3 <- nc_3 %>% 
        group_by(group) %>% 
        summarise(sum_area=sum(AREA), nr=n()) 
plot(nc_3)
st_write(nc_3, 'nc_3.gpkg', overwrite=TRUE)
```

I've tried to fix it by using `st_cast` function. While it fixed problem with plotting, there is still a problem with writing a file:

```R
nc_3b <- nc_3 %>% st_cast(., "MULTIPOLYGON")
plot(nc_3b)
st_write(nc_3b, 'nc_3b.gpkg', overwrite=TRUE)
```

edzer:
This now works: 
```
> st_write(nc_3, 'nc_3.gpkg', overwrite=TRUE)
Writing layer `nc_3.gpkg' to data source `/home/edzer/git/nc_3.gpkg' using driver `GPKG'
GEOMETRY: returning class wkbUnknown (0)
features:       3
fields:         3
geometry type:  Unknown (any)
```

PR #157, opened 2017-01-06
Add support for as() GEOMETRY

### LINKED ISSUE #131: Error on st_union() in summarise()

etiennebr:
See 93a5aff for a simple failing test. `st_area()` could not transform a GEOMETRY to `Spatial` because there was no `sfc_GEOMETRY` definition. `st_area()` had to call geosphere to compute area from angular coordinates (hence the conversion to `Spatial`). 

Worth mentioning:
* Add a precision setter and getter, they're very basic but I find it convenient, if that's ok.
* `setOldClass(c("sfc_GEOMETRY", "sfc"))` to enable as(geom, "Spatial"), when geom is a GEOMETRY. It then goes through `st_cast_default()`. 
* Replace `st_cast_default()` to coerce geometries to MULTI* when possible.

Closes #131 and #144, mention to #110, #138. @mdsumner, could you check if this breaks anything on your side, I believe you are working on #154. I'd be happy to edit the PR if needed.


etiennebr:
Also related to #87, but might not fix it yet.

R/cast_sfc.R
```diff
@@ -88,19 +88,45 @@ get_lengths = function(x) {
...
+      
+    } else if (all(cls %in% c("POLYGON", "MULTIPOLYGON"))) {
+      x <- lapply(x,
+                  function(x) if (inherits(x, "POLYGON")) POLYGON2MULTIPOLYGON(x) else x)
+      attributes(x) <- attributes(.x)
+      class(x) <- c("sfc_MULTIPOLYGON", "sfc")
+    } else {
+      class(x) <- c("sfc_GEOMETRY", "sfc") 
+    }
+  } else if (cls %in% c("GEOMETRY", "GEOMETRYCOLLECTION") ) {
```
edzer: cls will never be `GEOMETRY`, only `sfc`'s can be; I will take care of this.

PR #160, opened 2017-01-07
geom operations within a data frame

mdsumner:
Return a sf dataframe with the geometry column modified, not just the geometry.

Will fix the failing tests.

I wanted this for segmentizing, so I've applied it all all of `st_buffer`, `st_boundary`, `st_convex_hull`, `st_simplify`, `st_triangulate`, `st_polygonize`, `st_linemerge`, `st_centroid`, `st_segmentize`. 

I made each a generic function, removed `st_geometry(x)` from `st_fun.sfc` and applied `st_geometry(x) <- st_fun(st_geometry(x))` for each.


edzer:
Maybe start saying what this PR tries to do/to solve? Will you address the failing checks?

mdsumner:
Abandoning this in current form, will revisit.

PR #162, opened 2017-01-08
geom methods for sf, sfc , sfg

mdsumner:
This PR is to return a sf dataframe with the geometry column modified, not just the geometry for

`st_buffer`, `st_boundary`, `st_convex_hull`, `st_simplify`, `st_triangulate`, `st_polygonize`, `st_linemerge`, `st_centroid`, `st_segmentize`. 

Currently, the `st_buffer` ... `st_segmentize` functions accept `sf`, `sfc` or `sfg` objects, but always return `sfc`. 

There's a kind of implicit coercion currently where `st_geometry` elevates an `sfg` up and lowers an `sf` down, both to the `sfc` for the implementation itself. 

This PR converts those nine geom-modifying functions all to S3 generics, with  `.sfg` `.sfc` and `.sf` methods. 

* the .sfg method elevates with `st_sfc` and calls the .sfc method. 

* the .sf method modifies the geometry column with `st_geometry<-` 

* the .sfc method is the implementation as before, but without st_geometry(x). 

I didn't realize about the implicit coercion up that was happening for raw `sfg` when I started, hence the reboot on this PR.

It does mean that the affine example required updating in the vignette, otherwise st_centroid(nc) as `sf` cannot be used in the Ops for subtract. I realize this may well clash with the overall plan of how those Ops would work. 

I was motivated by wanting a copy of the original `sf` data frame, with the geometry modified in place. 

Finally, the man page says "Geometric operations on (pairs of) simple feature geometries" which I think means we should separate these nine into a separate Rd from the x,y functions here, and probably the same with "st_is_valid", "st_area", "st_is_simple", "st_length", "st_dimension", "st_combine" and "st_line_sample".


mdsumner:
Hooray, checks passed. This was tough, it's giving me an excellent education - I hope it's a fit for sf!

edzer:
Looks like you changed the indentation style completely, any particular reason for that? Or is this default behavior of rstudio?

mdsumner:
Ah, not for any reason sorry - I got into a muddle with the first PR changing too much, so I copied geom.R and re-branched.

tests/testthat/test_geos.R
```diff
@@ -92,11 +92,12 @@ test_that("geom operations work on sfg or sfc or sf", {
   expect_silent(st_simplify(pnc, FALSE, 1e4))
   expect_silent(st_simplify(gpnc, FALSE, 1e4))
   expect_silent(st_simplify(gpnc[[1L]], FALSE, 1e4))
-  
-  expect_silent(st_triangulate(pnc))
-  expect_that(st_triangulate(gpnc), is_a("sfc_GEOMETRYCOLLECTION"))
-  expect_that(st_triangulate(gpnc[[1]]), is_a("sfc_GEOMETRYCOLLECTION"))
-  
+
+  if (sf:::CPL_geos_version() >= "3.4.0") {
```
etiennebr: Check out `testthat::skip_if_not()` that allows for an informative message when skipping.

Issue #103, opened 2016-12-07
limits on geometric-dimension mixing?


mdsumner:
The rules for which topological types are allowed within a feature are quite clear, pretty much anything goes including empty GEOMETRYCOLLECTION but only one GEOMETRYCOLLECTION per feature. 

What about a mix of "XY" and "XYZ" geometric-dimensions within a feature? Within an sf object? 

It seems a bit perverse to allow either, with an implication that any explict XY would implicitly be XYZ but with a missing value for Z. 

I see that mixing is not allowed when collecting geometries: 

```R
st_sfc(st_point(c(1,2,3)), st_point(c(1,2), "XY"))
Error in st_sfc(st_point(c(1, 2, 3)), st_point(c(1, 2), "XY")) : 
  found multiple dimensions: XYZ XY
```

But with points we can mix them with implicit casting to multipoint, and with a warning recycling occurs:
 
```R
 c(st_point(c(1,2,3)), st_point(c(.001,2), "XY"))
MULTIPOINTZ(1 2 3, 0.001 2 0.001) 
Warning message:
In (function (..., deparse.level = 1)  :
  number of columns of result is not a multiple of vector length (arg 2)
```

This is not allowed for matrix based geometries: 

```R
m <- rbind(c(0,3),c(0,4),c(1,5),c(2,5))
c(st_linestring(m), st_linestring(cbind(m, 1)))
Error in MtrxSet(x, dim, type = "MULTILINESTRING", needClosed = FALSE) : 
  matrices having unequal number of columns

c(st_multipoint(m), st_multipoint(cbind(m, 1)))
Error in (function (..., deparse.level = 1)  : 
  number of columns of matrices must match (see arg 2)
```

I considered that NA should be inserted rather than recyling, but I also expected that NA would trigger an error and it does not. 

```R
st_point(c(1, NA))
```

It seems XYZ wins over XYM despite input order: 

```R
st_sfc(c(st_point(c(1, 2, 3), "XYM"), st_point(c(1,2,3))))
```

I'm unclear on whether this kind of mixing is desirable, I think it definitely shouldn't be within a feature, but I'm not sure about in one sf object, across features. 

Also, I'm not anticipating an influx of rogue mixing-constructors here, just unsure what the policy is, and I think ` c(st_point(c(1,2,3)), st_point(c(.001,2), "XY"))` needs to error out like linestring does when faced with a mix of input dimensions .

edzer:
[...]

PR #172, opened 2017-01-11
XYZ transformation tests

mdsumner:
PR to add a test suite for XYZ transformations, with two items that include partially failing tests . 
 
## 1. Z is preserved only for MULTI-types
  
Tests: 

* planar transformations preserve Z
* geocentric transformations preserve Z
 
Expectation:  XYZ geometries should preserve the Z dimension for planar  (4326 to laea) or geocentric (4326 to geocent)

Result:   Z is dropped for  the non-MULTI geometries. 

##  2.  Automatic elevation to XYZ for geocentric transformations
 
 This was mentioned in #103 as a possible improvement. 
 
Tests: 

 * geocentric transformations ADD Z
 
 Expectation: transformation to geocent should convert XY to XYZ
 
 Result:  XY geometries stay as XY. 
 
 
 
See also discussion here
 
 https://github.com/edzer/sfr/issues/103


tests/testthat/test_xyz.r
```diff
@@ -0,0 +1,92 @@
...
+pol <- st_polygon(pol1)
+polz <- st_polygon(pol1z)
+
+mpol <- st_multipolygon(list(pol1, pol1))
+mpolz <- st_multipolygon(list(pol1z, pol1z))
+
+## worker functions to apply transformations
+## do we need to test inverse versions of these as well? vertical? (see below)
+do_planar <- function(x) st_transform(st_sfc(x, crs = 4326), crs = "+proj=laea")
+do_geo <- function(x) st_transform(st_sfc(x, crs = 4326), crs = "+proj=geocent")
```
edzer: I think that projecting from 4326 to geocentric as implemented by Proj.4 only considers longitude and latitude, and outputs geocentric 3D points _on the 4326 ellipsoide_. This means that for this to be meaningful, if present a z coordinate has to be zero (which it is in this test: by accident or on purpose?) or otherwise should be rejected (by `st_sfc`?) since it is ignored by Proj.4.

For these examples to be meaningful (i.e. be sensitive to non-zero z coordinates on long/lat), I think you should look into using a [vertical datum](https://github.com/OSGeo/proj.4/wiki/VerticalDatums). I'd be happy to help out, but believe these are not the tests we need.

mdsumner: Do you not think the single/multi inconsistency is important?  For me those behaviours need clarification before going further. I made XY and XYZ geometries explicitly to tease out the distinctions.  ( I haven't yet checked what GDAL does with analogous tasks. None of this is pressing, but I see this is as a building block towards the vertical stuff).

edzer: What do you mean by single/multiple inconsistency? Do you understand, and agree to my comment?

mdsumner: I mean that I expect that XYZ geometries should preserve the Z dimension for planar or geocentric transformations, but that Z is dropped for the non-MULTI geometries. 

I don't know if I understand your comment, it doesn't appear so. I think you are telling me that this is not a test you need for transformations involving vertical datums, and I absolutely agree with that because it's not for that, it's a pre-requisite clarification I would need to go further, and a record of existing behaviour to regress against future changes.  If you are telling me this test is not useful because it's not meaningful to you without being a test for non-zero z within vertical datums then that's fine, I'll put it elsewhere, write the observations down elsewhere, and close this.

edzer: OK. I agree that long/lat -> geocentric should yield 3D coords, which right now it doesn't. I disagree that long/lat/zero is a useful representation of points on a non-vertical, 2D datum. I think that long/lat/alt should be accepted for vertical datums only, and geocent should always require x/y/z. So there is work to be done here, but your tests need modified somewhat.

mdsumner: Ok that's good, it clarifies about what I need to do to go further. As it is I'm happy to ignore this PR since I just won't be able to get to vertical datums, or to tease out what I found here, for a while. I don't agree about the "vertical datums only" part but I can't support that with any great weight now. As long as it's not precluded then I don't have a problem.

PR #187, opened 2017-01-25
Add RStudio project

hadley:
This makes life a little easier for those of us who use RStudio

### DIFF
--- a/sf.Rproj
+++ b/sf.Rproj
@@ -1,0 +1,21 @@
+Version: 1.0
+
+RestoreWorkspace: No
+SaveWorkspace: No
+AlwaysSaveHistory: Default
+
+EnableCodeIndexing: Yes
+UseSpacesForTab: Yes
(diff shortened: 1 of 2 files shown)


sf.Rproj
```diff
@@ -0,0 +1,21 @@
+Version: 1.0
+
+RestoreWorkspace: No
+SaveWorkspace: No
+AlwaysSaveHistory: Default
+
+EnableCodeIndexing: Yes
+UseSpacesForTab: Yes
+NumSpacesForTab: 2
```
edzer: To be consistent with the rest of the code, these would be "No" and 4, respectively. Do you see a problem there?

hadley: No, that's fine (it just inherited my default settings). 

Do you want me to change it?

edzer: I'll take care of it; thanks!

PR #189, opened 2017-01-25
Preserve crs objects

hadley:
e.g. so you can use `st_is_latlon()` directly on a crs object.


hadley:
I just noticed that I forgot to re-roxygenise to update the NAMESPACE 😞

edzer:
No probs, will do.

Issue #163, opened 2017-01-08
RTriangle alternative to GEOS for triangulate?


mdsumner:
We could use RTriangle as an alternative to st_triangulate, to avoid the GEOS 3.4.0 requirement.   (The license might be a problem: 	CC BY-NC-SA 4.0). 

There are many other Delaunay triangulators on CRAN that we could use, but this one has the added benefits over all the others in having constrained triangulations, breaking the criterion where needed to mesh input line segments, and the ability to apply other requirements on triangles like maximum area, minimum internal angles, and more. 

With st_triangulate we could break multipolygons up into parts and triangulate those separately, and filter out triangles from holes and concavities with point in poly test, but there still will often be edges that aren't in the Delaunay mesh. RTriangle gives an exact decomposition to triangles for any multipolygon, and there aren't many available implementations for that, this is the only one in R that I know of. 

I've done a basic example here: 

https://github.com/mdsumner/sfr/blob/constrained-triangulation/R/rt_triangulate.R

That relies on `dplyr` just for bind_rows, but that can be replaced with base. 

It's a bit quicker to short-circuit the extra checks done by `st_polygon`.

I use this in mesh-indexed form elsewhere, but GEOMETRYCOLLECTIONs of triangle polygons could be pretty useful. Perhaps it's of interest in sf itself.

edzer:
That license is a no-go. Remember `gpclib`?

mdsumner:
A no-go for what, CRAN?  (I'm confused)  https://CRAN.R-project.org/package=RTriangle

edzer:
I got that one wrong; of course, gpclib is also still there. 

In any case, we don't want an NC clause tainting the rights to use `sf`; you'd have to do it in a package deriving from sf.

mdsumner:
Ok no problem.

mdsumner:
implemented here: https://github.com/r-gris/sfct

PR #203, opened 2017-02-03
include MIT and GPL2 licences

etiennebr:
From reading #163, I realized license texts were missing (required by both MIT and GPL2). I think this is the right way to include multiples licenses.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,2 +1,368 @@
+-----------------------------------MIT------------------------------------------
 YEAR: 2016-2017
 COPYRIGHT HOLDER: Edzer Pebesma
+
+License: MIT + file LICENSE
+
+Copyright (c) 2016-2017, Edzer Pebesma
+
+Permission is hereby granted, free of charge, to any person obtaining
+a copy of this software and associated documentation files (the
+"Software"), to deal in the Software without restriction, including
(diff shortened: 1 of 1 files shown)


edzer:
What's the advantage of adding this when DESCRIPTION and the CRAN package site points to the right licenses, and FILE? (see also: "Writing R extensions").

nuest:
In other languages having a `LICENSE` file is common, in R it is actually not. If there is a license file, it normally contains the full text of the license. I find it confusing to see both. Since MIT is much less limiting than GPL, you could just put MIT in there.

etiennebr:
My understanding is that both MIT and GPL2 require the text to be
distributed with the software. Although I agree that MIT would probably be
sufficient, but I'm far from an expert on legal matters.
Le mar. 7 févr. 2017 à 07:22, Daniel Nüst <notifications@github.com> a
écrit :

nuest:
Fair point, and no worries, nobody is a lawyer here..

edzer:
See [WRE](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Licensing): I think the way we do it now is in accordance with what is written there; the license files are in every R distribution, it is not needed to repeat those in every package using them.

PR #209, opened 2017-02-07
speed up bulk point creation from data frame

mdsumner:
The bulk point creation in st_as_sf.data.frame is a bit slow. 

I've short-circuited the creation of the POINT geometries, and also created the sfc structure directly which makes this quite a bit faster. 

For the benchmark below with 1e4 points I see 2 seconds (vs 9 seconds) by vectorizing the `st_point` build and then to < 1 second by further avoiding checks in `st_sfc`

```R
n <- 1e4
set.seed(1)
d <- data.frame(x = rnorm(n), y = rnorm(n), z = rnorm(n), m = rnorm(n), a = seq_len(n))

library(sf)
system.time({
  aa <- st_as_sf(d, coords = c("x", "y"), crs = "+init=epsg:4326")
  bb <- st_as_sf(d, coords = c("x", "y", "z"))
  cc <- st_as_sf(d, coords = c("x", "y", "m"), dim = "XYM")
  dd <- st_as_sf(d, coords = c("x", "y", "z", "m"), dim = "XYZM")
  
})

class(st_geometry(aa)[[1]])
class(st_geometry(bb)[[1]])
class(st_geometry(cc)[[1]])
class(st_geometry(dd)[[1]])

```

### DIFF
--- a/R/sf.R
+++ b/R/sf.R
@@ -39,8 +39,15 @@ st_as_sf.data.frame = function(x, ..., agr = NA_agr_, coords, wkt,
 		else
 			x$geometry = st_as_sfc(as.character(x[[wkt]]))
 	} else if (! missing(coords)) {
-		x$geometry = do.call(st_sfc, c(lapply(seq_len(nrow(x)), 
-				function(i) st_point(unlist(x[i, coords]), dim = dim))))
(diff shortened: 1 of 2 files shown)


edzer:
Thanks, makes sense!

PR #213, opened 2017-02-10
Check for views and materialized views when reading and writing from db

### LINKED ISSUE #212: st_read_db() from view or materialized view
When I query a postgres view or materialized view with `st_read_db()` I get an error saying:
`Error: 'public.view' does not exist.`
However, I can read tables without problems. Also, `RPostgreSQL::dbReadTable()` works fine in all cases. I haven't found related info in the documentation so I wonder if I have missed a switch somewhere...
All relations are in the public schema.

etiennebr:
Search for tables, views and materialized views when checking read or write.

I couldn't find a generic DBI procedure for that, so it's very PostGIS specific and thus makes it less generic and will probably fail for other databases that don't support views, materialized views or that store them differently. I've also asked the DBI package, just in case.

fix #212

### DIFF
--- a/R/db.R
+++ b/R/db.R
@@ -24,7 +24,7 @@ st_read_db = function(conn = NULL, table = NULL, query = NULL,
     
     if (!is.null(table)) {
         table <- schema_table(table)
-        if (!DBI::dbExistsTable(conn, table)) {
+        if (!db_exists(conn, table)) {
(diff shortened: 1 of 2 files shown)


balazsdukai:
Thanks for the super quick fix!

etiennebr:
You're welcome! If you find anything else, don't hesitate.

PR #224, opened 2017-02-20
Try to document plot.sf

Robinlovelace:


edzer:
Shouldn't we rather give `plot.sf` its own manual page? The thing is very clobbered now. 

To keep in line with the rest of the documentation: if something is a method, we should call it a method.

Robinlovelace:
Yes that sounds good.

Not sure what you mean by 'the thing is very clobbered now'. It seemed rather 'clobbered' before, although that word is open to interpretation at the best of times, especially in a software development context.

Should I create a new man page? If so any suggestions about how to do that and, if possible, good examples from other well-documented packages, are very much appreciated.

Robinlovelace:
This now conflicts with https://github.com/edzer/sfr/commit/2550bf92108370b02a8968b97e7d71359ab77470 - @edzer should I re-add the extra content (with how `plot.sf` differs from `plot` in there) to you new one?

edzer:
I think the way forward would be to give `plot.sf` its own manual page. However, I'd like it to be minimal; if we want to educate people how base plot works too, a vignette would be more appropriate.

A plotting vignette would not be a bad idea anyway, if it also covers ggplot.

Robinlovelace:
Sweet - could build on this: https://github.com/Robinlovelace/Creating-maps-in-R/blob/master/vignettes/vspd-base-shiny.md I'm up for that so any comments before a move forward on it appreciated.

Robinlovelace:
Seemingly out of date now. @edzer let me know if you'd like any input on the docs for plotting and I can try to salvage some of this PR. Think were was something useful in there.

PR #227, opened 2017-02-22
Fix st_segmentize example in sf3.Rmd

JoshOBrien:
The current code amounts to just doing `points(0)` on both the line segment and the polygon.


edzer:
Thanks, Josh!

PR #240, opened 2017-03-02
basic st_poly_sample and C++ code cleanup

virgesmith:


Robinlovelace:
What are your thoughts on this @edzer? We were just talking about sf today - the project is going great guns and we hope to help, from Leeds to Munster!

edzer:
A large PR, consisting of lots of micro stuff and one undocumented function. Have you looked at the alternative `st_sample`? It doesn't (yet) return _exactly_ `size` points, yet, but at least doesn't have `st_intersects` in a potentially massive inner loop.

etiennebr:
I think that the while loop should have a condition to exit, like a time or iteration limit. Imagine a multipolygon with two tiny parts spread apart, producing a very large bbox compared to the polygons; it would take forever to have a point falling within them.

I think it could also be worth writing a test to show how much more efficient or faster it is than running st_sample until n is reached.

Also, a lot of that PR is end-of-line whitespace removed. It should probably be looked at with `git diff --ignore-space-change`.

edzer:
Since nothing seems to be moving here, I'm closing this PR.

Issue #274, opened 2017-03-26
Overwrite argument in st_write function.


Nowosad:
```R
library('sf')
nc = st_read(system.file("shape/nc.shp", package="sf"), quiet=TRUE)

st_write(nc, 'my_file.gpkg') # save for the first time - works

# I made some changes to nc object

st_write(nc, 'my_file.gpkg') # error - "Dataset already exists."
st_write(nc, 'my_file.gpkg', update = TRUE) # error "Layer creation failed."
st_write(nc, 'my_file.gpkg', layer_options = 'OVERWRITE=YES') # error - "Dataset already exists."
st_write(nc, 'my_file.gpkg', layer_options = 'OVERWRITE=YES', update = TRUE) # works
```

Above you can find my struggles with overwriting a file.  An alternative to that is to delete the file from a hardrive and try again.

Is it possible to implement a simple "overwrite" argument, such as: ?
`st_write(nc, 'my_file.gpkg', overwrite = TRUE)
`

edzer:
The story is a bit more complicated, since you can overwrite dsn's or layers, and an `overwrite` option would not make clear what to overwrite; in your commands above, at least it is clear that `overwrite` is an option that concerns the layer. For instance,
```r
st_write(nc, 'my_file.gpkg', "second_nc", update = TRUE)
```
adds a layer `second_nc` to `my_file.gpkg`; would you have expected `my_file.gpkg` to be overwritten with `overwrite=TRUE`, or layer `second_nc` in `my_file.gpkg` to be overwritten, in case it were already there?

See also `?writeOGR` which has more sensible arguments. We're looking at a single interface for nearly 80 drivers, many of which are not simply files.

https://github.com/edzer/sfr/commit/2e2b49c131ff2aabacd5e396bdd6db275acbda29 adds GPKG as one of the drivers for which `update` is `TRUE` by default.

Nowosad:
I thought about the `overwrite` option as "delete an old file and create a new one". I was not aware of any writable drivers, which are not files. Are there any?

edzer:
All the database drivers; gpkg is both: database and file, and so are spatialite, sqlite.
```r
> unlist(sf:::prefix_map)
       couchdb        db2odbc           dods            gft          mssql 
     "CouchDB"      "DB2ODBC"         "DODS"          "GFT" "MSSQLSpatial" 
         mysql            oci           odbc             pg            sde 
[...]

PR #288, opened 2017-04-02
Overwrite argument

Nowosad:
#274 Changes made:
1. Overwrite argument added to the st_write function
2. Overwrite argument from the st_write_db function renamed to drop

@edzer Please let me know what do you think about these changes.


edzer:
For how many drivers did you test this? What if `dsn` is `.`, or another directory? What if a data source consists of multiple files (like a shapefile, or another)? 

I feel rather uneasy taking responsibility for code that looks OK and works for a few drivers but may have unforeseen consequences for others. Why not ask GDAL do delete the dsn? In that case the driver implementer decides what should happen, and if it is reasonable to do this.

Nowosad:
http://rpubs.com/j_n/st_write - there you have some tests.

edzer:
What happens if you do
```r
st_write(meuse_sf, system.file("shape", package="sf"), "nc", overwrite = TRUE)
```
?

Nowosad:
> st_write(st_as_sf(meuse), system.file("shape", package="sf"), "nc", overwrite = TRUE)
Error: Could not guess driver for /home/username/R/x86_64-redhat-linux-gnu-library/3.3/sf/shape

edzer:
```r
st_write(st_as_sf(meuse), system.file("shape", package="sf"), "nc", driver = "ESRI Shapefile", overwrite = TRUE)
```

Nowosad:
```
 Error in CPL_write_ogr(obj, dsn, layer, driver, as.character(dataset_options),  : 
  Dataset already exists. 
```

However, (correct me if I'm wrong) - it should failed, but with a different error message. Let's try these examples:

```
nc = st_read(dsn = system.file("shape", package="sf"), layer = "nc", driver = "ESRI Shapefile") #failed
nc = st_read(dsn = system.file("shape", package="sf"), driver = "ESRI Shapefile") #failed
nc = st_read(system.file("shape/nc.shp", package="sf")) #works
```

Therefore, if we are unable to read a file this way, we should be also unable to write it this way.

edzer:
Sorry Jakub, I appreciate your efforts very much, but I don't think we should go this way. In sf, we interface to GDAL for reading and writing data, for good reasons, and we should use this same interface if we want to delete data sources. We can then thank (or blame) the driver implementer when it doesn't do what we had in mind.

Nowosad:
Understood. I will think of a gdal way to do it.

PR #293, opened 2017-04-04
Document about st_ prefix

yutannihilation:
Hi,

I saw several people (including me 😃 ) got confused why the prefix is `st_`, not `sf_`. I guess this is just because I'm a newbie to GIS and it will be OK for most of GIS guys, who are familiar with the `ST_` prefix of PostGIS, right? So, I don't complain about the prefix, but could you consider add some sentence about what `st` stands for.

(This PR is an example and please close this anytime. I'm not confident with my English...)


edzer:
Thanks, great idea, I added a somewhat different explanation, found [here](http://doesen0.informatik.uni-leipzig.de/proceedings/paper/68.pdf).

yutannihilation:
Oh, spatial and temporal... Sorry for my misunderstanding :P

Anyway, thanks for the clear explanation. Cool!

PR #294, opened 2017-04-04
Initial implementation of overwrite argument in st_write

Robinlovelace:
With commit from @Nowosad in there and example of use case with .shp file.

WIP. Comments / suggested changes welcome...


edzer:
Robin, please also see https://github.com/edzer/sfr/pull/288 and https://github.com/edzer/sfr/issues/204. I decided that if we want to delete a datasource, we should ask gdal to do so, not R, since R has no idea what a datasource means. Also, update is something else as delete. GDAL has no concept of overwrite (afaics), but you can try to open and then delete before you write.

edzer:
See also [crud](https://en.wikipedia.org/wiki/Create,_read,_update_and_delete).

Robinlovelace:
Fair points well made - I fully agree with your approach of making **sf** as 'close to the metal' of GDAL as possible and I think your current implementation does that. Fair enough not having a nice `overwrite` argument. What I dislike is how this is [documented](https://github.com/edzer/sfr/blob/master/R/read.R#L137):

```r
#' @param update logical; if \code{TRUE}, try to update (append to) existing data source;this is only supported by some drivers
#' (e.g. GPKG), for other drivers the layer may simply be overwritten
```

That is not precise enough - the file does get overridden in most cases - 'may' introduces ambiguity to the user. Further, 'simply be overwritten' is not good English - superfluous adverb usage!

How about I put in another PR just focussing on improving the documentation of `update`?

edzer:
Yes, please do.

Issue #229, opened 2017-02-22
Convenience function to access non-spatial fields


sboysel:
Is there a convenience function to extract the non-spatial fields of a feature collection akin to the data slot in `sp` (e.g. `x@data`)?  Something that explicitly drops any geometry data from the feature collection and returns a simple `data.frame`:

```r
st_just_data <- function(x) {
  df <- as.data.frame(x)
  df_sfc_cols <- sapply(df, function(i) inherits(i, "sfc"))
  df[, !df_sfc_cols]
}
```

Is this something useful?  Is there a better way of doing this?

tim-salabim:
I also wrote my own function for this. What I found crucial though is to retain the data frame structure in the output, hence I would suggest the following:
```r
geompos <- which(names(x) == attr(x, "sf_column"))
return(data.frame(x)[, -geompos, drop = FALSE])
```
In general, I think this could be a useful function.

edzer:
You can use `st_geometry(nc) <- NULL`.

edzer:
(the more elegant way if you want to do it "by hand", which I learned very recently, is to use `setdiff`:

    nc[setdiff(names(nc), attr(nc, "sf_column"))]

note that all this assumes there is only _one_ geometry list-column, which is the current assumption of `sf` objects)

tim-salabim:
Very elegant. I assume `setdiff` will preserve the `data.frame` class?
If so, I will update my function accordingly, thanks!

edzer:
Sorry, the right incantation would be

    x = as.data.frame(nc)[setdiff(names(nc), attr(nc, "sf_column"))]

since otherwise `[.sf` is used, which makes the geom sticky. But my first suggestion is shorter, and uses the `sf` api.

edzer:
The original proposal of @sboysel btw removes _all_ geometry columns.

edzer:
Multiple geom columns are discussed in https://github.com/edzer/sfr/issues/183

Robinlovelace:
+1 for having this as a convenience function.

PR #297, opened 2017-04-08
st_attr returns non-spatial attributes of sf object

sboysel:
Addresses #229 (convenience function to extract non-spatial fields from `sf` object) using the method suggested by @edzer.  Perhaps the function could benefit from some renaming for the sake of clarity.  Any thoughts?


edzer:
I don't think this problem is large enough to get a function on its own. In addition, sf now allows for more than one geometry list-column, and st_attr doesn't deal with that, afaics.

PR #300, opened 2017-04-13
add sample argument to st_line_sample

joethorley:
so that user can specify points to sample


edzer:
What happens if sample is not numeric, or contains values outside [0,1]?

joethorley:
Values outside the range are set as 0 or 1
```
> identical(st_line_sample(st_transform(ls, 3857), sample = c(0, 1)), st_line_sample(st_transform(ls, 3857), sample = c(-1, 2)))
[1] TRUE
```
Integers are fine
```
> identical(st_line_sample(st_transform(ls, 3857), sample = c(0, 1)), st_line_sample(st_transform(ls, 3857), sample = c(0L, 1L)))
[1] TRUE
```

Characters cause error to be thrown
```
> st_line_sample(st_transform(ls, 3857), sample = c("0", "1"))
 Error in `*.default`(sample, l[i]) : 
  non-numeric argument to binary operator 
```

Seems reasonable. 

Note I went minimal on input checking as this seemed to be the style of the other arguments - let me know if you want any modifications.

edzer:
Thanks for checking!

PR #305, opened 2017-04-16
Add libudunits2-dev to list of dependencies for ubuntu in installation instructions (README)

loicdtx:


edzer:
make sense. Thanks!

PR #310, opened 2017-04-23
logo link fixed

Nowosad:


edzer:
Seems the logo moves with every  update they do. I made a copy to a more permanent area.

Issue #349, opened 2017-05-18
liblwgeom support on macOS


pat-s:
Related: #347, #266 

I did `brew install liblwgeom && brew link --force liblwgeom` and during `sf` source install `liblwgeom` is found. However, ` lwgeom_version in -llwgeom` is not found. 

```
checking liblwgeom.h usability... yes
checking liblwgeom.h presence... yes
checking for liblwgeom.h... yes
checking for lwgeom_version in -llwgeom... no
```

`sf` v0.4.3

Any ideas where to look here? I do not know where `-llwgeom` is referring to..

edzer:
My guess is that the version of lwgeom is too old, but I haven't tried or looked into.

pat-s:
* brew: 2.1.5_3 (Dec 2013) 
* latest postgis: 2.3.2
* Debian 9: 2.3.0 (Sep 2016)

It seems patching the brew formula might not be that easy -> `brew edit liblwgeom` 10-13:

> Strip all the PostgreSQL functions from PostGIS configure.ac, to allow
>     building liblwgeom.dylib without needing PostgreSQL
>     NOTE: this will need to be maintained per postgis version
>     Somehow, this still works for 2.1.5, which is awesome!

edzer:
I see, on ubuntu 16.04:
```r
> library(sf)
Linking to GEOS 3.5.1, GDAL 2.1.3, proj.4 4.9.2, lwgeom 2.3.2 r15302
```
(seems that lwgeom versions follow postgis version)

pat-s:
> (seems that lwgeom versions follow postgis version)

jep, its dev repository is within the postgis one: https://github.com/postgis/postgis/tree/svn-trunk/liblwgeom

edzer:
So we can close this one, right?

pat-s:
I would like to leave it open until there is a working solution for the `liblwgeom` install - the current brew version does not work (probably its too old as you suggested). I will later try to install from source / patch the brew formula. 

ping @bhaskarvk

edzer:
I like to see open issues as issues that need work on the side of `sf`. As this is an issue with brew and MacOS, I'm closing here, but would welcome it if you report back here when you've settled that.

pat-s:
`brew install postgis` solves the problem as it does the linking to `-llwgeom`

```
checking liblwgeom.h usability... yes
checking liblwgeom.h presence... yes
checking for liblwgeom.h... yes
[...]

PR #350, opened 2017-05-18
Update lwgeom.cpp - CPL_make_valid() error message

pat-s:
#349


edzer:
Thanks, I can see these are all well intended, but I don't think such instructions belong to error messages; I used them to improve the package README. In case you would want such messages in errors, you could make them OS dependent in the R code, so that windows users would not be bothered by them. On many linux distributions your suggested error message will not be of much help.

PR #397, opened 2017-06-24
change src/geos.cpp int *.size() -> unsigned *.size()

mpadge:
clang output on `-Wall -pedantic` compile of latest version (0.5-1):
```
geos.cpp: In function ‘Rcpp::List CPL_geos_binop(Rcpp::List, Rcpp::List, std::__cxx11::string, double, std::__cxx11::string, bool, bool)’:
geos.cpp:276:23: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
     for (int j = 0; j < tree_sel.size(); j++)
                       ^
geos.cpp:297:24: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
      for (int j = 0; j < tree_sel.size(); j++)
                        ^
geos.cpp:315:24: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
      for (int j = 0; j < tree_sel.size(); j++)
                        ^
```
This fixes that. Note also that travis checks for dependency packages require that they pass `clang/clang++ -Wall -pedantic` (plus other flags), and so travis would refuse to install this as a dep for another package, even though travis builds of the repo itself will pass. (I learnt this the hard way: [see this example](https://travis-ci.org/ropensci/osmplotr/jobs/244886520#L3759-L3790)).

### DIFF
--- a/src/geos.cpp
+++ b/src/geos.cpp
@@ -273,7 +273,7 @@ Rcpp::List CPL_geos_binop(Rcpp::List sfc0, Rcpp::List sfc1, std::string op, doub
 				std::vector<size_t> tree_sel, sel;
 				if (! GEOSisEmpty_r(hGEOSCtxt, gmv0[i]))
 					GEOSSTRtree_query_r(hGEOSCtxt, tree1, gmv0[i], cb, &tree_sel);
-				for (unsigned j = 1; j < tree_sel.size(); j++)
+				for (size_t j = 0; j < tree_sel.size(); j++)
 					if (chk_(GEOSRelatePattern_r(hGEOSCtxt, gmv0[i], gmv1[tree_sel[j]], pattern.c_str())))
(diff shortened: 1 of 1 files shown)


edzer:
using `unsigned` without type specifier is not my style (and considered sloppy by some, leaving the type implicitly defaulting to int); please use `size_t` instead as used by the rest of the code. 

Changing the 0 into a 1 on lines 276 and 297 caused errors that took me pretty long time to find - your PR comments only mentions the type, I obviously overlooked the 0 to 1 change the first two times. Is it intentional, and if yes, what for?

mpadge:
sorry about wasting your time with the "1" values there - i've no idea how that happened, and should have seen in it the PR in the first place. Agreed that "unsigned" is lazy and sloppy; I had semi-retrained myself to "unsigned int" because Armadillo uses that instead of "size_t". Non-issue here, but all quirks that reveal themselves at some stage ...

PR #402, opened 2017-06-29
Makes DBI dependency specification stricter (correct)

demorenoc:
The `DBI::dbExecute()` method [appeared in v0.5 of `DBI`](https://cran.rstudio.com/web/packages/DBI/news.html), so `sf` fails to install if using previous `DBI` versions.


edzer:
Thanks!

Issue #421, opened 2017-07-15
graticule not working with ggplot of world2 wrapping around longitude [0,360]


bbest:
Hi @edzer,

Thank you for the wonderful sf package! The new @hadley tidyverse paradigm applied to spatial analysis is a lifesaver, especially with huge speed improvements from [spatial indexes](http://r-spatial.org/r/2017/06/22/spatial-index.html). I noticed one little issue with ggplot2 not quite working  as anticipated with the Pacific-centric world2 (longitude of [0,360] vs [-180,180]):

```r
library(ggplot2)
library(maps)
devtools::install_github('r-spatial/sf')
library(sf) 

wrld2 = st_as_sf(map('world2', plot=F, fill=T))
ggplot() +
  geom_sf(data=wrld2, fill='lightgray') +
  coord_sf(xlim=c(140,295), ylim=c(5,60))
````

![image](https://user-images.githubusercontent.com/2837257/28239014-d8a3ee50-6915-11e7-9b9c-cfea8bb2e7eb.png)

There's a problem with the graticules disappearing. Solution forthcoming...

**Comment by bbest** [comment]
Just sent pull request https://github.com/r-spatial/sf/pull/422 with the fix...

```r
library(ggplot2)
library(maps)
devtools::install_github('bbest/sf')
library(sf)

wrld2 = st_as_sf(map('world2', plot=F, fill=T))
ggplot() +
  geom_sf(data=wrld2, fill='lightgray') +
  coord_sf(xlim=c(140,295), ylim=c(5,60))
```

![image](https://user-images.githubusercontent.com/2837257/28239044-423f962e-6917-11e7-9100-2cb284aa6662.png)

edzer:
Great work, thanks!

PR #422, opened 2017-07-15
fix for graticule not working with ggplot of world2 wrapping around longitude [0,360]

bbest:
Fix for graticule not working with ggplot of world2 wrapping around longitude [0,360] https://github.com/r-spatial/sf/issues/421.

```r
library(ggplot2)
library(maps)
devtools::install_github('bbest/sf')
library(sf)

wrld2 = st_as_sf(map('world2', plot=F, fill=T))
ggplot() +
  geom_sf(data=wrld2, fill='lightgray') +
  coord_sf(xlim=c(140,295), ylim=c(5,60))
```

image

vs before...

```r
library(ggplot2)
library(maps)
devtools::install_github('r-spatial/sf')
library(sf) 

wrld2 = st_as_sf(map('world2', plot=F, fill=T))
ggplot() +
  geom_sf(data=wrld2, fill='lightgray') +
  coord_sf(xlim=c(140,295), ylim=c(5,60))
````

image

### DIFF
--- a/R/graticule.R
+++ b/R/graticule.R
@@ -118,17 +118,22 @@ st_graticule = function(x = c(-180,-90,180,90), crs = st_crs(x),
 
 	bb = st_bbox(box_ll)
 	if (is.null(lon)) {
-		lon = if (bb[1] < -170 && bb[3] > 170) # "global"
-				seq(-180, 180, by = 60)
-			else
-				pretty(bb[c(1,3)], n = 6)
+	  if (bb[3] <= 180 && bb[1] < -170 && bb[3] > 170){
+	    lon = seq(-180, 180, by=60)
+	  } else {
+	    lon = pretty(bb[c(1,3)], n = 6)
+	  }
 	}
 	if (is.null(lat))
 		lat = pretty(bb[c(2,4)], n = 6)
 
 	# sanity:
 	if (isTRUE(st_is_longlat(datum))) {
-		lon = lon[lon >= -180 & lon <= 180]
+	  if (max(lon)>180){
+		  lon = lon[lon >= 0 & lon <= 360]
+	  } else {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: R/graticule.R [hunk 1]
```diff
@@ -183,19 +188,21 @@ graticule_attributes = function(df) {
...
 	bb = st_bbox(df)
-    selE = df$type == "E" & df$y_start < min(df$y_start) + 0.001 * (bb[3] - bb[1])
+  selE = df$type == "E" & df$y_start < min(df$y_start) + 0.001 * (bb[3] - bb[1])
 	selN = df$type == "N" & df$x_start < min(df$x_start) + 0.001 * (bb[4] - bb[2])
 	df$plot12 = selE | selN
 	df
 }
 
-trim_bb = function(bb = c(-180, -90, 180, 90), margin) {
+trim_bb = function(bb = c(-180, -90, 180, 90), margin, wrap=c(-180,180)) {
```
bbest: Borrowing `wrap`=c(-180,180) default argument from [maps::map()](https://www.rdocumentation.org/packages/maps/versions/3.2.0/topics/map):

- **wrap**:  Boolean or a numeric vector. If TRUE, lines that cross too far across the map (due to a strange projection) are omitted. If wrap is a vector of length 2 or more, it is interpreted as the longitude range to be used for a global map, e.g. c(-180,180) or c(0,360). This wrapping even works when fill=TRUE and is performed before any projection (so the range must always be in degrees). However, the wrapping is performed after xlim, ylim are applied, so these options should probably never be combined. If there is a third component, this signifies the latitude at which Antarctica will be "closed". The default value is -89.9. Special values are NA (don't draw Antarctica at all) and 0 (draw the line at the latitude of the extremal points, not at a fixed lower latitude).

edzer:
Great, thanks a lot!

PR #441, opened 2017-07-28
Remove trailing spaces

karldw:
It'd be nice to remove all trailing spaces from the code (partly for style, partly because Atom keeps doing it automatically), but feel free to reject this PR if it's overly aggressive.

The python programmer in me also wants to replace the mix of tabs and spaces with all spaces, but that's an even more dramatic change.

### DIFF
--- a/CONDUCT.md
+++ b/CONDUCT.md
@@ -1,6 +1,6 @@
 # Contributor Code of Conduct
 
-As contributors and maintainers of this project, we pledge to respect all people who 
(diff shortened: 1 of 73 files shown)


edzer:
I know your intentions are good, but trailing spaces are as invisible as they are harmless. I'd be happy if you'd replace the spaces with tabs, see the "contributing" section in the README. I'd be MUCH more happy with meaningful contributions, though.

karldw:
Sounds good, thanks!

PR #481, opened 2017-09-13
add a docs example for st_area

jsta:
I apologize for the line ending change on the `st_centroid` example. I cannot figure out how to suppress it.


edzer:
Thanks!

PR #482, opened 2017-09-13
st_collectionextract

ateucher:
I needed the equivalent of postgis' [`ST_CollectionExtract`](https://postgis.net/docs/ST_CollectionExtract.html) the other day, so I thought it might be worth offering it up. Definitely no hard feelings if you're not interested.  

I made it using only exported functions, so if there are more elegant/performant ways of doing this by going lower-level, let me know. I got the basics from [this stackoveflow question](https://stackoverflow.com/questions/45723775/how-to-select-certain-geometries-from-a-geometrycollection-after-st-intersect).


ateucher:
I just found a case I didn't have a test for that doesn't work... working on a fix

R/collectionextract.R
```diff
@@ -0,0 +1,122 @@
...
+#' st_collectionextract(cc, "POLYGON")
+#'
+st_collectionextract = function(x, type = c("POLYGON", "POINT", "LINESTRING"), warn = FALSE) {
+	UseMethod("st_collectionextract")
+}
+
+#' @name st_collectionextract
+#'
+#' @export
+st_collectionextract.sfg = function(x, type = c("POLYGON", "POINT", "LINESTRING"), warn = FALSE) {
```
edzer: In line with function naming, please rename into `st_collection_extract` -- I think PostGIS users will understand, as we don't use camel case.

ateucher: Ok, will do

ateucher:
I think that's better now

R/collection_extract.R
```diff
@@ -0,0 +1,138 @@
+#' Given a (multi)geometry, return a (multi)geometry consisting only of elements of the specified type.
+#'
+#' Similar to st_collection_extract in postgis. If there are no sub-geometries of the specified type, an empty geometry is returned.
```
edzer: Similar to `ST_CollectionExtract` in PostGIS.

R/collection_extract.R
```diff
@@ -0,0 +1,138 @@
...
+#'
+#' @param x an \code{sf(c/g)} object that has mixed geometry
```
edzer: an object of class \code{sf}, \code{sfc} or \code{sfg} that has...

With mixed geometry, do you also mean MULTIPOLYGON or only GEOMETRYCOLLECTION?

R/collection_extract.R
```diff
@@ -0,0 +1,138 @@
+#' Given a (multi)geometry, return a (multi)geometry consisting only of elements of the specified type.
```
edzer: Also GEOMETRYCOLLECTION?

R/collection_extract.R
```diff
@@ -0,0 +1,138 @@
...
+#'
+#' Similar to st_collection_extract in postgis. If there are no sub-geometries of the specified type, an empty geometry is returned.
+#'
+#' @param x an \code{sf(c/g)} object that has mixed geometry
+#' @param type One of "POLYGON", "POINT", "LINESTRING"
```
edzer: type character; one of

R/collection_extract.R
```diff
@@ -0,0 +1,138 @@
...
+#' @param warn logical; if \code{TRUE}, warn if attributes are assigned to sub-geometries when casting (see \code{\link{st_cast}})
+#'
+#' @return An sf(c/g) object with (multi)geometries consisting only of elements of the specified type. For sfg objects, an sfg object is returned if there is only one geometry of the specified type, otherwise the geometries are combined into an sfc object of the relevant type. If any subgeometries in the input are MULTI, then all of the subgeometries in the output will be MULTI.
```
edzer: An object having the same class as \code{x}, with ...

sfg -> \code{sfg}

R/collection_extract.R
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	# Find the geometries of the specified type and extract into a list
+	matches = vapply(x, st_is, types, FUN.VALUE = logical(1))
+	x_types = x[which(matches)]
+	if (length(x_types) == 0L) {
+		## Not sure if this should be a warning and return an empty sfg (current)
+		## or error?
+		warning("x contains no geometries of specified type")
+		return(x_types)
```
edzer: better return `typed_empty(type)` here, so at least an `sfg` is returned?

R/collection_extract.R
```diff
@@ -0,0 +1,138 @@
...
+	} else if (length(x_types) == 1L) {
+		# Get the contents of the first (only) list element which is an sfg
+		return(x_types[[1]])
+	} else {
+		# turn list into an sfc, and cast it to single type. Will be multi
+		# if any are multi
+		return(st_cast(st_sfc(x_types), warn = warn))
```
edzer: Yeah, so we either return an sfg or an sfc from this method; I guess there's no better option.

ateucher:
Great comments @edzer, thanks. I'll start working on them.

R/collection_extract.R
ateucher: Not that I could find, I don't think running `st_multi()` on the list of `sfg` objects to force an `sfg` would be desirable.

R/collection_extract.R
ateucher: As it's set up it works on only `GEOMETRYCOLLECTION` for `sfg` and `GEOMETRY` or `GEOMETRYCOLLECTION` for `sfc` and `sf`. It doesn't work to extract `POLYGON` from `MULTIPOLYGON` etc. I *think* this is consistent with `ST_CollectionExtract`. I'll try to make that clearer in the documentation.

R/collection_extract.R
ateucher: Thanks. Didn't know about `typed_empty`

edzer:
Thanks, great work!

ateucher:
Fantastic, thanks!

Issue #180, opened 2017-01-18
Compare CRS


etiennebr:
What's the best option to compare crs's ? For instance:

```r
epsg <- "+proj=tmerc +lat_0=0 +lon_0=0 +k=0.9999 +x_0=304800 +y_0=0 +datum=NAD83 +units=m +no_defs"
st_crs(epsg)$proj4string; epsg
# [1] "+proj=tmerc +lat_0=0 +lon_0=0 +k=0.9999 +x_0=304800 +y_0=0 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs"
# [1] "+proj=tmerc +lat_0=0 +lon_0=0 +k=0.9999 +x_0=304800 +y_0=0 +datum=NAD83 +units=m +no_defs"
```

Using `==` like it's currently the case seems too strict.

```r
epsg <- "+proj=tmerc +lat_0=0 +lon_0=0 +k=0.9999 +x_0=304800 +y_0=0 +datum=NAD83 +units=m +no_defs"
y <- x <- st_crs(epsg)
y$proj4string <- epsg  # this is cheating, but it's for illustration; it happened in a "natural" way via st_read
x == y
# FALSE
```

I see `sp::identicalCRS()`, `sf` doesn't seem to have an equivalent, yet. Any thoughts?

edzer:
If you don't cheat, `x == y` is `TRUE`. What is your point?

etiennebr:
Sorry if the application wasn't clear. I have shapefiles coming from `st_read()` with `x` and I need to merge them with shapefiles coming from `st_transform()` with `y` (ref. to the example above). But they won't merge because they have different `crs` strings. I'd be fine merging them because I know they're the same (nad83 uses GRS80). Is there a way to do it automatically? Apparently proj4 knows it's equivalent because it made the switch.

edzer:
If you `st_transform` _y_ into the `st_crs` of _x_, do they still have different `st_crs`?

etiennebr:
Yes, (sorry for giving context piece by piece, I was attempting to simplify the problem). I did something like:
```r
shapefiles <- lapply(lf, st_read)
crs <- st_crs(shapefiles[[1]])
shapefiles_ident_crs <- lapply(shapefiles, function(.) st_transform(., crs))
```
Those that didn't need reprojection were left with `epsg`, but the others have `st_crs(epsg)`.
```r
sapply(shapefiles_ident_crs, st_crs)
[...]

PR #485, opened 2017-09-19
add test for CRS comparison (#180)

etiennebr:
Awesome @edzer! I didn't know about GDAL isSame. I think adding a test might be useful for maintenance.


edzer:
Thanks, Etienne!

PR #488, opened 2017-09-24
check PostGIS driver in dontrun blocks

rsbivand:
Most of the changes here were to let me run check --run-dontrun on GDAL 2.2.2 w/o PostGIS; now the code checks that the PostGIS driver is present before trying to use it. I don't know if this is the only check needed, as the db daemon also needs to be running, I think.

The mess with roxygen2 was caused by the serious breakdown in roxygen2 that roxygen-dev (or similar) is not available, so I followed HWs advice and used devtools::document() instead, but all that needed changing was the example block in man/ (didn't touch that, would have been much less invasive).

Final extra issue: Dirk found yesterday that recent autoconf (he and I have 2.69) treat AC_CHECK_HEADERS more aggressively, and try to compile a stub program with the header. I've commented out the gdal.h check in configure.ac and run autoconf. The build passed check --run-dontrun OK. I haven't looked further, but maybe this will bite PROJ.4, GEOS and liblwgeom. Dirk's advice is trust *-config if it exists; a work-around for PROJ.4 w/o config is on R-Forge for rgdal.


edzer:
Thanks; I now solved this through https://github.com/r-spatial/sf/commit/20ade90c3c6e828ba16827d59c5b527b3815ac47  ; https://github.com/r-spatial/sf/commit/4d749cbd3c1157184a42d88701d35c01ab61e1ec provides a test environment.

PR #512, opened 2017-10-10
Add stringsAsFactors parameter to sf's rbind method

gergness:
Would it be possible to include the stringsAsFactors parameter for rbind.sf? It'll help tide me over until the issues around bind_rows are sorted out.

Thanks for the great work!

(Apologies about the extra parts of the diff - it looks like I have a setting in either Rstudio or git that got rid of spaces at the end of lines automatically. Let me know if you'd prefer I resubmit.)


gergness:
Sorry, I found a bug in my code - I'm not sure this is actually needed.

PR #519, opened 2017-10-16
as_Spatial() could support sf objects

etiennebr:
I find it confusing that `as(x, "Spatial")` can support `sf` objects, but not the `as_Spatial(x)` so this is my proposal to add support for it. It might be more elegant to make `as_Spatial()` generic or something else. I went for minimal impact to initiate the discussion.


R/sp.R
```diff
@@ -170,32 +173,64 @@ setAs("sf", "Spatial", function(from) {
...
 #' @rdname coerce-methods
 #' @aliases coerce,sfc,Spatial-method
 setAs("sfc", "Spatial", function(from) as_Spatial(from))
 
-# setAs("sfg", "Spatial", function(from) as(st_sfc(from), "Spatial"))
-##  doesn't work for:
-## as(st_point(0:1), "Spatial")
+# create empy class
+setClass("XY")
+setAs("XY", "Spatial", function(from) as(st_sfc(from), "Spatial"))
```
edzer: why not use `setAs("sfg", "Spatial",...etc)`  here? This would then also work for `XYM` points.

PR #520, opened 2017-10-17
fixed typo

Zedseayou:
geomtry -> geometry


edzer:
Changing it here doesn't make sense, as the .Rd files are auto generated from the .R files; please modify R/st_geometry.R instead.

PR #526, opened 2017-10-21
Add a `pal` argument to `plot.sf` to change the palette.

hughjonesd:
This is an attempt to make plot.sf more flexible by letting the user choose a palette, e.g. `heat.colors`. I have left `sf.colors` alone as it is publicly documented, but potentially this function could be simplified as now some of the work is done in the private `get_colors` function.

Tell me if you think this is the right approach.

I also threw in code to plot quantiles of numeric variables rather than just equally spaced intervals. Again, this makes sense to me but YMMV.

### DIFF
--- a/R/plot.R
+++ b/R/plot.R
@@ -3,12 +3,11 @@
 #' @param x object of class sf
 #' @param y ignored
 #' @param ... further specifications, see \link{plot_sf} and \link{plot}
-#' @param ncol integer; default number of colors to be used
-#' @param max.plot integer; lower boundary to maximum number of attributes to plot (defaults to 9)
 #' @param main title for plot (\code{NULL} to remove)
 #' @param pal palette function, similar to \link[grDevices]{rainbow}; if omitted, \link{sf.colors} is used
-#' @param nbreaks number of colors breaks
+#' @param nbreaks number of colors breaks (ignored for \code{factor} or \code{character} variables)
 #' @param breaks either a numeric vector with the actual breaks, or a name of a method accepted by the \code{style} argument of \link[classInt]{classIntervals}
+#' @param max.plot integer; lower boundary to maximum number of attributes to plot; the default value (9) can be overriden by setting the global option \code{sf_max.plot}, e.g. \code{options(sf_max.plot=2)}
 #' @param pch plotting symbol
 #' @param cex symbol size
 #' @param bg symbol background color
@@ -85,7 +84,7 @@
 #' gc = st_sf(a=2:3, b = st_sfc(gc1,gc2))
 #' plot(gc, cex = gc$a, col = gc$a, border = rev(gc$a) + 2, lwd = 2)
 #' @export
-plot.sf <- function(x, y, ..., ncol = 10, col = NULL, max.plot = 9, main, pal = NULL, nbreaks = 10, breaks = "pretty") {
+plot.sf <- function(x, y, ..., col = NULL, main, pal = NULL, nbreaks = 10, breaks = "pretty", max.plot = ifelse(is.null(n <- options("sf_max.plot")[[1]]), 9, n)) {
 	stopifnot(missing(y))
 	dots = list(...)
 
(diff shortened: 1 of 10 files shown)


R/plot.R
```diff
@@ -521,7 +521,7 @@ get_colors = function(pal, n, xc) {
 			if (all(is.na(x)) || all(range(x, na.rm = TRUE) == 0))
 				rep(1, length(x))
 			else
-				cut(x, n)
+				cut(x, breaks = quantile(x, seq(0, 1, 1/n), na.rm = TRUE))
```
edzer: I can see that this would be useful to some, but why would it be a better default? (In the end, I guess we need a color legend, and could make this optional)

R/plot.R
```diff
@@ -106,10 +108,10 @@ plot.sf <- function(x, y, ..., ncol = 10, col = NULL, max.plot = 9, main) {
 		}
 		# col selection may have changed; set cols again:
 		cols = setdiff(names(x), attr(x, "sf_column"))
-		invisible(lapply(cols, function(cname) plot(x[, cname], main = cname, col = col, ...)))
+		invisible(lapply(cols, function(cname) plot(x[, cname], main = cname, col = col, pal = pal, ...)))
 	} else {
 		if (is.null(col) && ncol(x) == 2)
-			col = sf.colors(ncol, x[[setdiff(names(x), attr(x, "sf_column"))]])
+			col = get_colors(pal, ncol, x[[setdiff(names(x), attr(x, "sf_column"))]])
```
edzer: Why not use `pal(ncol, x[[setdiff(names(x), attr(x, "sf_column"))]])` here?

hughjonesd:
Good question. Hmm, pros and cons:

* Pro: by dividing the data up equally, it uses the full range of colours.
That should
make maps easier to interpret.
* Pro: it is easier to look at two maps and see if the underlying data are
correlated.
* Con: it is harder to look at two maps and make an absolute comparison of
values.

But I am not sure about that "con", because the cutpoints of two different
numeric variables will be different anyway, even without quantiles.

For comparison, I will attach maps of two variables in Trafford,
Manchester, one with and one without using quantiles. I think the version
with quantiles makes it easier
to see the (significant) correlation in observed data.

Cheers,
David

R/plot.R
hughjonesd: Not all palettes take two arguments in this way. For example, `heat.colors` does not.

hughjonesd:
no-quantiles
with-quantiles

rsbivand:
Would it not make sense to use the classInt package like tmap does? I'd be happy to migrate classInt to github.

edzer:
Yes, as that generalizes the quantile idea; I think we should allow for 

* a parameter `pal`, which is a pallette function similar to heat.colors
* a parameter `nbreaks` specifying the number of colors wanted
* a parameter `breaks` specifying the actual breaks if numeric, or the `style` parameter to be passed to `classInt::classIntervals` (including `quantile` and many others)

Clearly, `sf.colors` now does too much!

rsbivand:
Shall I put classInt on github? If so where? In r-spatial as a fork from my (as yet unmigrated) repo, or just straight into r-spatial?

edzer:
Migration will make your life easier.

tim-salabim:
This is related to an r-spatial discuss issue I have just opened (I have been meaning to do this for a while now) https://github.com/r-spatial/discuss/issues/16

edzer:
I implemented your PR in a slightly different way. Thanks!

rsbivand:
classInt migrated to r-spatial.

edzer:
Right now the default `breaks` value is `"pretty"`; I chose that because I think that by default we do want a color scale on the side (or bottom) of a map with colors. We're not yet there...

hughjonesd:
Tell me if there's other stuff that needs doing.

edzer:
@hughjonesd the fifth and sixth plot [here](https://edzer.github.io/sp/#other-spatial-objects-in-base-plot) have color keys; this code is in sp, in the last 150 lines of https://github.com/edzer/sp/blob/master/R/SpatialGridDataFrame-methods.R. I think that if we plot single sf maps, a color key should be displayed.

hughjonesd:
OK, I've finished putting the curtain rails up, so I'll give it a shot.

-- 
Sent from Gmail Mobile

PR #527, opened 2017-10-23
Pass a bare matrix to st_polygon() and a list of matrix to st_multipolygon()

yutannihilation:
IIUC, in terms of the semantics of Simple Features, a bare matrix is not the equivalent of `POLYGON` and a list of matrix is not of `MULTIPOLYGON`. But, it doesn't seem a big deal to assume each matrix is a single polygon without holes and wrap them with list. This may be useful for the data like this:

``` r
library(GISTools, quietly = TRUE)
#> Checking rgeos availability: TRUE
#> rgeos version: 0.3-25, (SVN revision 555)
#>  GEOS runtime version: 3.6.1-CAPI-1.10.1 r0 
#>  Linking to sp version: 1.2-5 
#>  Polygon checking: TRUE
data("georgia")

str(georgia.polys, list.len = 5L)
#> List of 159
#>  $ : num [1:125, 1:2] 1292287 1292654 1292949 1294045 1294603 ...
#>  $ : num [1:99, 1:2] 1263206 1264530 1263799 1263714 1263095 ...
#>  $ : num [1:53, 1:2] 1267843 1270202 1271306 1272213 1272824 ...
#>  $ : num [1:124, 1:2] 1120149 1120164 1119692 1119295 1118798 ...
#>  $ : num [1:116, 1:2] 1175513 1177004 1192290 1192624 1192486 ...
#>   [list output truncated]
```

It would be nice if we can create a multipolygon with

```r
x <- sf::st_multipolygon(georgia.polys)
```

instead of manually wrapping them with list:

```r
x <- sf::st_multipolygon(lapply(georgia.polys, list))
```

But I'm not quite sure my assumptions are correct. What do you think...?

### DIFF
--- a/R/sfg.R
+++ b/R/sfg.R
@@ -135,9 +135,9 @@ st_linestring = function(x = matrix(numeric(0), 0, 2), dim = "XYZ") Mtrx(x, dim,
 #' @name st
 #' @export
 st_polygon = function(x = list(), dim = if(length(x)) "XYZ" else "XY") {
-	if (identical(x, 1))
-		st_polygon(list(rbind(c(0,0),c(1,0),c(1,1),c(0,1),c(0,0))))
-	else MtrxSet(x, dim, type = "POLYGON", needClosed = TRUE)
+	if (is.matrix(x)) x = list(x)
+	if (identical(x, 1)) x = list(rbind(c(0,0),c(1,0),c(1,1),c(0,1),c(0,0)))
+	MtrxSet(x, dim, type = "POLYGON", needClosed = TRUE)
 }
 #' @name st
 #' @export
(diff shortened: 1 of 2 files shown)


edzer:
I see `st_polygon` and `st_multipolygon` as mostly having didactical purposes: they allow users to build simple features from scratch. If we allow them to take all kind of formats, this effect goes away.

In your case, you want to modify them not for didactical purpose, but to make import from another package easier. Had this package thought a bit more about its users, it would have provided polygons in a form that is understood by R users (e.g. as `SpatialPolygons`) or (if there was a need) it would have provided polygons in a class they defined themselves. 

Assuming the package would only deal with polygons consisting of single outer rings and without holes, the package should provide such objects as a list of matrices with a class, say `GISTools_polygons`, in which case you could have written an `st_as_sf` method (`st_as_sf.GISTools_polygons`) that does exactly what you now try to do above by modifying `st_polygons` etc.

So if you want to get this going, contact the authors of `GISTools` and offer help there. Given the date of the last release and the lack of reverse dependencies, I'm not sure this package should still be considered alive.

yutannihilation:
Oh, reasonable. Thank you very much for clarification!

PR #528, opened 2017-10-23
Add color key

hughjonesd:
Here is the beginning of a color key. It is very basic and doubtless has bugs, but I just want to check you think the approach is OK. The drawing code is copied over from `sp` and simplified a bit.

I don't try to plot a key if there are multiple columns. It could be done, but it would require changing from `par(mfrow)` and/or `layout` to `split.screen` (which claims to nest nicely).


edzer:
Thanks a lot! I tried to make the flow more logical, and added checks for corner cases such as
```
demo(nc)
nc$const = 1
nc$const[1:10] = NA
plot(nc["const"])
```
and
```
nc$f = factor(nc$const)
plot(nc["f"])
```
where, now, in the first case you don't get a (continuous) color key, in the second you do. Also, I added key.size as a parameter, to get things like
```
nc$new = cut(nc[[1]], 5)
plot(nc["new"], key.size = lcm(5))
```
plot reasonably. Still, if key.pos = 1 this case plots terrible; maybe we shouldn't want that casse.

PR #536, opened 2017-11-03
st_minimum_bounding_circle

rundel:
Added functionality for calculating minimum bounding circle using lwgeom library. Replicates ST_MinimumBoundingCircle from PostGIS (https://postgis.net/docs/ST_MinimumBoundingCircle.html)


edzer:
`is.na(sf_extSoftVersion()["lwgeom"])` can be used to make example/tests conditional.

rundel:
Opps missed that check, should be there now

edzer:
There are two considerations here to discuss:
* should we boundlessly add `st_do_whatever` and end up with a huge name space by the end of the day, similar to what PostGIS has? Or can this command be grouped with others, somehow?
* should we strive for limited name lengths, such that typing `st_ <tab> <tab>` does not give endless white space? (this one adds five to the currently longest command)

rsbivand:
Is there any Rcpp mechanism to handle "linking to"? I'd also be concerned that exposing lots of st_rabbits in sf may be a maintenance problem later on. I can also see that there is a static-linking issue too; with dynamic linking, a stub in sf could be linked to in an a package linking to sf, but with static linking, all of the lwgeom functionality would be brought in anyway simply by creating a pointer to an external function not otherwise used. 

I did have to unroll Colin's nice work with function pointers in rgeos because it got across CRAN checks IIRC.

tim-salabim:
maybe it's time for sf.extras?

edzer:
Maybe it's not such a bad idea to start a separate pkg, (`sf_lwgeom`?) with all the lwgeom stuff in it. I don't think any linkingto is required, for that. It would make it easier to understand for users that the lwgeom stuff is not in CRAN binaries. We could try to develop that package such that it has lwgeom actually included (like @rubak did with s2), to get it to CRAN.

Nowosad:
And what's with the `lwgeom` functions names? Is it still be `st_name` or something else, such as `lw_name`?

tim-salabim:
Having lwgeom actually shipped w the binary would be nice from a windows point of view. How many functions would then have to migrate from sf to lwgeom?

edzer:
@Nowosad liblwgeom is the geometry wrapper library used in PostGIS; it is easiest to mirror their naming conventions. It does much of things what we now do in sf with GEOS, which we won't repeat, but it has quite a few additional features. So yes, `st_name` and upper camel case becomes `_` separated.

edzer:
Ah, @rsbivand I now see the need for LinkingTo: ...

rundel:
I take the point at st_minimum_bounding_circle is super unwieldy, the goal was to match the existing functions which match PostGIS's naming conventions.

It seems reasonable that a separate lwgeom package is probably warranted, the actual library is fairly self contained and should be possible to include it directly in the package instead of requiring linking to an outside library. Moving the existing lwgeom dependent sf functionality would likely require that some sf stuff get moved into header files which are then moved into inst/include or something similar so that `LinkingTo` will work.

edzer:
https://github.com/r-spatial/lwgeom

edzer:
(I now simply copied over src/wkb* and src/bbox* from sf; this needs some work in the LinkingTo: department!)

PR #537, opened 2017-11-03
lwgeom split

rundel:
Based on the discussion in PR #536. This is a quick attempt to remove all the lwgeom dependent functionality from sf and move it into r-spatial/lwgeom (see the pull request there). I've been fairly crude in hacking pieces out and moving everything in wkb.cpp and bbox.cpp into shared header files. Everything is inlined to avoid symbol duplication issues during linking (there may be a cleaner approach to this?).

On my system both packages work and pass R CMD CHECK but I'm interested to see if it works for anyone else.


edzer:
Great work I now did this manually in https://github.com/r-spatial/sf/commit/5b144f36be51aee88c5a5b01a5ba29c54ed9c5ff and https://github.com/r-spatial/sf/commit/9341cff41654e275a9b9aec39889a39255e481de, and managed to also export the CPL_read/write_wkb stuff, which needed some deeper cutting. See https://github.com/r-spatial/lwgeom/pull/1 for further discussion.

PR #543, opened 2017-11-07
Fix st_length for mixed geometries

### LINKED ISSUE #542: st_length broken for mixed polygon / multipolygon (sfc_geometry)
Currently st_length gives an error when used with a geometry column of type sfc_geometry, e.g.

```r
outer = matrix(c(0,0,10,0,10,10,0,10,0,0),ncol=2, byrow=TRUE)
hole1 = matrix(c(1,1,1,2,2,2,2,1,1,1),ncol=2, byrow=TRUE)
hole2 = matrix(c(5,5,5,6,6,6,6,5,5,5),ncol=2, byrow=TRUE)

poly = st_polygon(list(outer, hole1, hole2))
mpoly = st_multipolygon(list(
  list(outer, hole1, hole2),
  list(outer + 12, hole1 + 12)
))

class(st_sfc(poly, mpoly))
## [1] "sfc_GEOMETRY" "sfc"  
st_length(st_sfc(poly, mpoly))
## Error: inherits(x, "sfc_LINESTRING") || inherits(x, "sfc_MULTILINESTRING") is not TRUE
```

Currently [st_length](https://github.com/r-spatial/sf/blob/master/R/geom.R#L78) only checks for point/multipoint and polygon/multipolygon, anything else triggers the `stopifnot`. sfc_GEOMETRY can't just be added to the polygon check condition since it might be hiding badness like points or more esoteric types. 

A better solution seems like iterating through the individual geometries and the doing the checks and calculations at that level?

rundel:
Fixes #542

### DIFF
--- a/R/geom.R
+++ b/R/geom.R
@@ -66,38 +66,69 @@ ll_length = function(x, fn, p) {
 	}
 }
 
+
+geom_length = function(x, dist_fun, longlat, crs) {
+	stopifnot(inherits(x, "sfg"))
+
+	if (inherits(x, "POINT") || inherits(x, "MULTIPOINT"))
+		return(0.0)
+
+	x = st_cast(x, "MULTILINESTRING")
+
+	if (isTRUE(longlat)) {
+		if (is.na(crs))
+			stop("crs must be defined for longitude/latitude data")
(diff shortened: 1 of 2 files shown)


edzer:
Thanks, Colin!

PR #558, opened 2017-11-17
Enable database connection for st_read and st_write

etiennebr:
This PR enables database (i.e. postgis) connection for `st_read` and `st_write`. It also updates tests and  documentation.

I've moved `st_*_db()` to a new `deprecated.R`. I didn't make any effort to reconcile the arguments for `st_read.character` and `st_read.PostgreSQLConnection)`, which might be a little sloppy. I'm open to trying to reconcile arguments as much as possible since this is an opportunity, if you have any preference, please let me know.


R/db.R
```diff
@@ -98,10 +97,10 @@ st_read_db = function(conn = NULL, table = NULL, query = NULL,
 #'   library(RPostgreSQL)
 #'   try(conn <- dbConnect(PostgreSQL(), dbname = "postgis"))
 #'   if (exists("conn") && !inherits(conn, "try-error"))
-#'     st_write_db(conn, sf, "meuse_tbl", drop = FALSE)
+#'     st_write(conn, sf, "meuse_tbl", drop = FALSE)
 #' }
-#' @details st_write_db was written with help of Josh London, see \url{https://github.com/r-spatial/sf/issues/285}
-st_write_db = function(conn = NULL, obj, table = deparse(substitute(obj)), geom_name = "wkb_geometry",
+#' @details database reading was written with help of Josh London, see \url{https://github.com/r-spatial/sf/issues/285}
+st_write.PostgreSQLConnection = function(conn = NULL, obj, table = deparse(substitute(obj)), geom_name = "wkb_geometry",
```
edzer: the arguments before ... in the generic need to have identical names to that of the generic, which is: `obj, dsn, layer`; similar for `st_read`.

R/deprecated.R
```diff
@@ -0,0 +1,26 @@
...
+#'   \code{st_read_db} \tab now a synonym for \code{\link{st_read}}\cr
+#'   \code{st_write_db} \tab now a synonym for \code{\link{st_write}}\cr
+#' }
+#'
+st_read_db <- function(...) {
+	.Deprecated("st_read")
+	st_read(...)
+}
+
+st_write_db <- function(...) {
```
edzer: here, the argument order needs to change: `st_write_db(conn, obj, layer)` should become `st_write(obj, conn, layer, ...)` but then with correct names `obj, dsn, layer`.

R/db.R
etiennebr: The problem is that if we put the connection (`dsn`) in second, then we'll need to use S4 generic. Otherwise we'll only catch `st_write.sf()`. It also seems more common to see the connection as the first argument. S4?

edzer: Better to branch in `st_write.sf` - then I'm pragmatic and would like to avoid S4. 

For `write_xxx` function it's more common to see the object to write as first argument, I'd say.

etiennebr:
There is something I can't explain yet when reading the crs from the database.It might be a problem of version, but on my personal computer EPSG 28992 is different than from postgis for some decimals of the `towgs84`parameter.
``` r
#  crs from the database
st_crs(st_read("PG:dbname=postgis", "meuse"))  
#> Coordinate Reference System:
#>   EPSG: 28992 
#>   proj4string: "+proj=sterea +lat_0=52.15616055555555 +lon_0=5.38763888888889 +k=0.9999079 
#> +x_0=155000 +y_0=463000 +ellps=bessel +towgs84=565.4171,50.3319,465.5524,
#> -0.398957,0.343988,-1.87740,4.0725 +units=m +no_defs"
# local crs
st_crs(st_read(pg, "meuse"))
#> Coordinate Reference System:
#>   EPSG: 28992 
#>   proj4string: "+proj=sterea +lat_0=52.15616055555555 +lon_0=5.38763888888889 +k=0.9999079 
#> +x_0=155000 +y_0=463000 +ellps=bessel +towgs84=565.2369,50.0087,465.658,
#> -0.406857,0.350733,-1.87035,4.0812 +units=m +no_defs"
```
Where `pg` is a connection to a postgis database. See [tests](https://github.com/r-spatial/sf/pull/558/files#diff-833e0176c5034738e94e18dd9bcf9038R50) for details.bh

edzer:
Yes, this indicates changes to the EPSG database, present in different proj.4 versions; this particular one is well known. National geodesy institutes update this when they improve the mapping from their national system to (in this case) the global datum WGS84.

etiennebr:
To clarify the PR:

* Uses  `st_write()` and `st_read()` rather than `st_read_db()` and `st_write_db()` (both deprecated)
* Uses generic `dbWriteTable()`, which, thanks to `DBI` package makes it very simple to add support for other (postgis enabled) databases.
* Adds support for `DBIObjects` in addition to `RPostgreSQL`
* Makes `RPostgreSQL` required in the `Imports` :-1: . Maybe there's another way I'm not aware of?
* Now that `RPostgres` is on CRAN, it would probably be possible to remove `RPostgreSQL` support., but speaking for myself, I'd need a little transition period. 
* Adds tests, namely support for `odbc` connections (on postgres database). It works in most ordinary cases, but there are some glitches such as `overwrite` and `append` outside of default (public) schema fails. [But it seems to relate to the `odbc` package](https://github.com/r-dbi/odbc/issues/140).
* Provides an option to create invalid crs with `st_crs(valid = FALSE)` to synch with database (basically to replace `structure(list(epsg=..., proj4string=...), class="crs")`).

To do:
* [ ] Fix Travis-ci build
* [ ] Reorder arguments with deprecated functions
* [ ] Add package `odbc` to `Suggests` for tests

Probably other things. Happy to discuss any of these and modify the pull request before it gets merged.

edzer:
Cool that RPostgres is out! Let me know when you have travis running here; looking good. Release sf 0.5-6 planned early Jan.

dpprdan:
Err, sorry to barge in here, but #592 might be a problem? Or would that be solved with the `dbWriteTable`? Still there would be an inconsistency w.r.t. allowed field types on a PostgreSQL backend with `st_write` vis-a-vis `dbWriteTable`?

etiennebr:
@dpprdan, I'll add explicit test for logical type. It's the best way to convince everyone :)

R/deprecated.R
```diff
@@ -0,0 +1,26 @@
+#' Deprecated functions in `sf`
+#'
+#' These functions are provided for compatibility with older version of `sf`.
+#' They may eventually be completely removed.
+#' @md
+#' @rdname sf-deprecated
+#' @name sf-deprecated
+#' @param ... Parameters to be passed to the modern version of the function
```
edzer: Pleaase remove this line to make travis pass check.

edzer: No, sorry, please ignore; seems to be OK.

edzer:
Is this ready for merge?

etiennebr:
I can't wait to say yes!... but no :( But very close!

edzer:
No hurries!

etiennebr:
@edzer, thanks! Do I really need to export it? I thought exporting the generic sufficient.

edzer:
If you want to let users use the method, you need to export it, not only the generic.

etiennebr:
This PR is ready to merge :tada: ! however the commits are messy. @edzer, would you like me to clean up the commits before merging?

edzer:
When testing the RPostgres tests, I see
```r
> pg <- DBI::dbConnect(RPostgres::Postgres(), host = "localhost", dbname = "postgis")
Error in connection_create(names(opts), as.vector(opts)) : 
  fe_sendauth: no password supplied
```
any idea? the `postgis` db doesn't have a password set.

edzer:
When I remove the `host = "localhost"` argument, things work. Should I modify my postgres configuration? Do you need the `host` setting for your tests?

etiennebr:
Check your [`~/.pgpass`](https://www.postgresql.org/docs/current/static/libpq-pgpass.html). Your credentials should be there, if not add them. It should solve your problem.  

I thought I could do without the `host`, but it seems more robust to provide it. For instance to connect locally to a Docker instance it need the `localhost` specification, and it seems to work fine on Travis. Same thing with `PG:host=localhost`.

edzer:
Thanks, that works now. Using `localhost`, the users is apparently forced to have a password.

etiennebr:
It seems to be driver dependent. For instance, ODBC was less demanding. 

By the way, ODBC is not fully supported, there are still some issues, but it works for most cases. So the test is just skipped for now.

tests/testthat/test_postgis_RPostgres.R
```diff
@@ -0,0 +1,320 @@
...
+})
+
+test_that("can write to other schema", {
+    skip_if_not(can_con(pg), "could not connect to postgis database")
+    try(DBI::dbSendQuery(pg, "CREATE SCHEMA sf_test__;"), silent = TRUE)
+    q <- "SELECT schema_name FROM information_schema.schemata WHERE schema_name = 'sf_test__';"
+    suppressWarnings(could_schema <- DBI::dbGetQuery(pg, q) %>% nrow() > 0)
+
+    skip_if_not(could_schema, "Could not create schema (might need to run 'GRANT CREATE ON DATABASE postgis TO <user>')")
+    expect_error(st_write(pts, pg, Id(schema = "public", table = "sf_meuse__")), "exists")
```
edzer: I'm getting
```r
Error: `st_write(pts, pg, Id(schema = "public", table = "sf_meuse__"))` threw an error with unexpected message.
Expected match: "exists"
Actual message: "unable to find an inherited method for function ‘dbWriteTable’ for signature ‘\"PqConnection\", \"SQL\", \"missing\"’"
```
here -- any idea?

etiennebr: Make sure you are running the latest `DBI` and `RPostgres`.

edzer: I did, but that didn't resolve it; sessionInfo below; postgresql 9.6.8. 

When I outcomment these test, next thing I get stuck with is 
```r
> st_write(pts, pg, Id(schema = "CAP__", table = "Meuse_tbl"))
Error in (function (classes, fdef, mtable)  : 
  unable to find an inherited method for function ‘dbWriteTable’ for signature ‘"PqConnection", "SQL", "missing"’
```
What am I missing?
<details>

```r
> sessionInfo()
R version 3.4.3 (2017-11-30)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 17.10

Matrix products: default
BLAS: /usr/lib/x86_64-linux-gnu/openblas/libblas.so.3
LAPACK: /usr/lib/x86_64-linux-gnu/libopenblasp-r0.2.20.so

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] sp_1.2-7       testthat_2.0.0 DBI_0.8       

loaded via a namespace (and not attached):
 [1] bit_1.1-12      compiler_3.4.3  magrittr_1.5    R6_2.2.2       
 [5] hms_0.4.2       tools_3.4.3     Rcpp_0.12.16    bit64_0.9-7    
 [9] grid_3.4.3      blob_1.1.1      pkgconfig_2.0.1 rlang_0.2.0    
[13] RPostgres_1.1.0 lattice_0.20-35
```
</details>

etiennebr: This error is because the test uses the new way of naming tables in `DBI`. It uses `DBI::Id()` (prior was `DBI::SQL()`). Also `Rpostgres` was just recently updated (https://github.com/r-dbi/RPostgres/issues/179) could you try `devtools::install_github("r-dbi/rpostgres")`. I was surprised that travis ran the test without complaining.

Looking at the travis logs (`devtools::session_info()`)
```
DBI              0.8       2018-03-02 cran (@0.8)
```
Can't find `RPostgres`, but it must be 1.1.0. (I'm using `RPostgres 1.1.0 2018-04-05 from Github (r-dbi/RPostgres@ef2ab88)`)

dpprdan: This is probably not causing this, but in general, shouldn't `RPostgres` be in `Suggests` in `DESCRIPTION`, (preferably with minimal version number)? Or am I missing something?

edzer:
Cool -- now merged in master!! I had to install DBI from github to get the thing going, in the end, and remove one `host=localhost` from an st_read test that used the gdal driver.

tests/testthat/test_postgis_RPostgres.R
etiennebr: Good question. I don't know. I think the whole point of using DBI is to only require DBI, but it's true that we sould still suggest RPostgres, but then should we suggest other drivers?

etiennebr:
Yay! Writing the code was nothing compared to having the tests run successfully in another environment :)

tests/testthat/test_postgis_RPostgres.R
dpprdan: If there are tests for them, they ought to be added, right?

etiennebr: For our own education, I searched a bit for other examples and I found that `dplyr` uses [`data.table` in tests](https://github.com/tidyverse/dplyr/blob/892cf876ec8044f1ec6f5ca7904f939fd9281262/tests/testthat/test-group-by.r#L134), but it's nowhere in the [`DESCRIPTION`](https://github.com/tidyverse/dplyr/blob/master/DESCRIPTION). I still think it's a good idea to suggest it, though. Maybe @krlmlr knows the reason?

krlmlr: We moved the data.table backend to *dtplyr* a while ago. The tests are gone now: https://github.com/tidyverse/dplyr/blob/master/tests/testthat/test-group-by.r.

edzer:
This now also requires DBI and RPostgres versions from gitnub to pass travis checking. 

@etiennebr could you please restore code coverage [here](https://codecov.io/gh/r-spatial/sf/src/master/R/db.R) with either adding tests or nocovs? And could you pls complete the NEWS entries? Also: would you be tempted to draft an r-spatial.org blog post on all the new capabilities? I'd be happy to help here.

tests/testthat/test_postgis_RPostgres.R
etiennebr: Ah, thanks (I got confused in the commits)! Makes sense. So if I understand correctly we should `Import` `DBI` and `Suggest` all drivers we explicitly test.

etiennebr:
@edzer yes and yes!

etiennebr:
When is the next release?

edzer:
A month from now (2 months between releases is required by CRAN), or perhaps earlier if 3.5.0 burns the whole thing down before that (see [here](https://cran.r-project.org/web/checks/check_results_sf.html)).

edzer:
I think 2. was caused by #718. Can that be a problem? We could, in principle, move it to last before writing.

dpprdan:
@edzer I cannot say for sure, but I think I noticed that already earlier than two days ago, i.e. before you merged the fix for #718. 

Re: 1. To be more precise, GDAL renames the geometry column to `wkb_geom`, DBI does not and leaves the name as-is (right?).

etiennebr:
Thanks @dpprdan, indeed, `st_write` copies the table to the database as closely as it can. It uses the column names and order from the data frame. I'll make sure to mention that, but I don't think it's an issue (or a surprise). GDAL creates more surprises and indeed shouldn't be encouraged.

PR #570, opened 2017-11-27
Add unnest method for sf class

karldw:
Changes:
- Create unnest.sf method, automatically adding the geometry column(s) to the list of list columns that will *not* be unnested.
- Add tests.
- Increase suggested version of tidyr to > 0.7.2 to be able to use `.preserve`. (Note, the CRAN and Github versions are both at 0.7.2, so the tests currently fail.)
- Unindent a few lines in `gather.sf` and `spread.sf` for clearer expression (not part of `if` block).
- Auto-remove a bunch of trailing spaces.

Do you want to hold off merging this until tidyr does a release?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: sf
 Version: 0.5-6
 Title: Simple Features for R
-Description: Support for simple features, a standardized way to encode spatial vector data. Binds 
+Description: Support for simple features, a standardized way to encode spatial vector data. Binds
     to GDAL for reading and writing data, to GEOS for geometrical operations, and to Proj.4 for projection
 	conversions and datum transformations.
 Authors@R: c(person("Edzer", "Pebesma", role = c("aut", "cre"), email = "edzer.pebesma@uni-muenster.de", comment = c(ORCID = "0000-0001-8049-7069")),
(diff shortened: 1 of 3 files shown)


karldw:
Update: this should work with the development version of tidyr now.

tests/testthat/test_tidy.R
```diff
@@ -31,9 +31,22 @@ test_that("sample_n etc work", {
...
 test_that("st_intersection of tbl returns tbl", {
- nc = read_sf(system.file("shape/nc.shp", package="sf")) 
+ nc = read_sf(system.file("shape/nc.shp", package="sf"))
  nc = st_transform(nc[1:3,], 3857)
  st_agr(nc) = "constant"
  expect_is(nc, "tbl_df")
  expect_is(st_intersection(nc[1:3], nc[4:6]), "tbl_df")
 })
+
+test_that("unnest works", {
```
edzer: If you would make the test conditional on the tidyr version, it might pass tests and could be merged now.

DESCRIPTION
```diff
@@ -15,34 +15,34 @@ Authors@R: c(person("Edzer", "Pebesma", role = c("aut", "cre"), email = "edzer.p
...
 	tmap,
-    maptools, 
+    maptools,
 	RSQLite,
     RPostgreSQL,
     tibble,
 	rlang,
 	dplyr (>= 0.7-0),
-    tidyr (>= 0.7-1),
+    tidyr (> 0.7-2),
```
edzer: Should be left as it was, here.

edzer:
Thanks! This is now ignored with CRAN versions; FYI I tried testing with tidyverse GH versions of tidyr, tidyselect and rlang, but that broke on `rename`, for reasons probably not caused here.

Issue #572, opened 2017-11-28
st_bbox() could return a geometry


etiennebr:
This is a suggestion derived from https://github.com/r-spatial/stars/issues/12.

Postgis uses [`ST_Envelope`](https://postgis.net/docs/ST_Envelope.html) to return the geometry of a bounding box (it's also the expression in the OGC spec), but since we're not very fond of adding new functions, I'd like to express my preference for having `st_bbox()` return a geometry by default (polygon with five vertex). User could use `st_as_text()` when in need of a legible version. 

While the current option to rather output a custom bbox class better expresses the max and min, there aren't as many methods to handle `bbox` objects than there are to handle geometries. For instance union of two bbox would require a transformation. I like the idea to keep the output in the most usable class. My two cents.

Robinlovelace:
Agree this would be useful for many users. I wrote a function ([`stplanr::geo_bb`](https://github.com/ropensci/stplanr/blob/master/R/geo-functions.R#L221)) that does this for `Spatial` and `sf` objects as I found myself wanting to clip and do other operations on an object using the bb of another. Borrowing from the very useful `tmaptools::bb` function it also allows you to scale it. Not sure if that is adding too much complexity but my view is that such things are useful enough to go into 'core' geo libraries like sf so v. happy at this suggestion (and up for helping if I can in any way).

edzer:
```r
> methods(class = "bbox")
[1] is.na     print     st_as_sfc st_crs   
see '?methods' for accessing help and source code
> demo(nc, ask = FALSE, echo = FALSE)
> (bb = st_as_sfc(st_bbox(nc)))
Geometry set for 1 feature 
geometry type:  POLYGON
dimension:      XY
bbox:           xmin: -84.32385 ymin: 33.88199 xmax: -75.45698 ymax: 36.58965
epsg (SRID):    4267
proj4string:    +proj=longlat +datum=NAD27 +no_defs
POLYGON ((-84.3238525390625 33.8819923400879, -...
```
so you'd need
```r
st_envelope = function(x) st_as_sfc(st_bbox(x))
```
and, yes, we could add this function, as we can add many more, and in PostGIS this is indeed the only thing you can do. 

Regarding scaling: try Ops:
```r
bb * 1.1 + 0.5
```

edzer:
[...]

PR #573, opened 2017-11-28
bbox docs

Robinlovelace:
Still figuring-out best way of scaling it without moving the centre.

### DIFF
--- a/R/bbox.R
+++ b/R/bbox.R
@@ -62,7 +62,10 @@ bbox.MtrxSetSetSet = function(obj) {
 #' @examples
 #' a = st_sf(a = 1:2, geom = st_sfc(st_point(0:1), st_point(1:2)), crs = 4326)
 #' st_bbox(a)
-#' st_as_sfc(st_bbox(a))
+#' b = st_as_sfc(st_bbox(a))
+#' b2 = st_as_sfc(st_bbox(a) * 1.5)
+#' plot(b2)
+#' plot(b, add = TRUE)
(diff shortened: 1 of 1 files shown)


R/bbox.R
```diff
@@ -62,7 +62,12 @@ bbox.MtrxSetSetSet = function(obj) {
 #' @examples
 #' a = st_sf(a = 1:2, geom = st_sfc(st_point(0:1), st_point(1:2)), crs = 4326)
 #' st_bbox(a)
-#' st_as_sfc(st_bbox(a))
+#' b = st_as_sfc(st_bbox(a))
+#' b_larger = b * 1.1
```
edzer: this functionality has nothing to do with st_bbox: why introduce it here?

Robinlovelace: To show people how you can work with bbs to meet a common need: to increase or decrease the 'field of view'. In short: user-friendliness.

Robinlovelace: But agree this example may not best demonstrate that (hence it being work in progress - just an idea).

Robinlovelace:
OK I've updated it to actually demonstrate scaling and shifting the bb which is surely useful for at least some users... Any constructive review comments, or event just 'not interested ditch it' welcome.

ratnanil:
Has this PR been ditched? Working with bounding boxes should def. be introduced somewhere. It's a very common task and somewhat obscurely documented (I keep looking up https://github.com/r-spatial/sf/issues/572#issuecomment-347602717). Or am I missing something?

Issue #611, opened 2018-01-10
parallel processing


jeffreyhanson:
I've been working with some large-ish data sets and I was thinking it would be great if the _sf R_ package could use parallel processing. I noticed that there was an _openmp_ branch, but I can't tell if much progress has been made on it, and I was wondering if anyone was working on this? If not, I have a little experience with _Rcpp_ and _OpenMP_ and I'm keen to take a shot at implementing this.

From what I can tell, for the functions that utilise _geos_, the trick seems to be creating a copy of the handler object (`GEOSContextHandle_t`) for each thread. For instance, implementing something like this:

```
// [[Rcpp::export]]
Rcpp::List CPL_geos_op(std::string op, Rcpp::List sfc, 
    Rcpp::NumericVector bufferDist, int nQuadSegs = 30,
    double dTolerance = 0.0, bool preserveTopology = false,
    int bOnlyEdges = 1, double dfMaxLength = 0.0, std::size_t threads = 1) {

  int dim = 2;
  std::vector<GEOSContextHandle_t> hGEOSCtxt(threads);
  for (std::size_t i = 0; i < threads; ++i)
    hGEOSCtxt[i] = CPL_geos_init();

  std::vector<GEOSGeom> g = geometries_from_sfc(hGEOSCtxt[0], sfc, &dim);
  std::vector<GEOSGeom> out(sfc.length());

  if (op == "buffer") {
    if (bufferDist.size() != (int) g.size())
      Rcpp::stop("invalid dist argument"); // #nocov
    #pragma omp parallel for num_threads(threads)
    for (size_t i = 0; i < g.size(); i++)
      out[i] = chkNULL(GEOSBuffer_r(hGEOSCtxt[omp_get_thread_num()], g[i], bufferDist[i], nQuadSegs));
  } else if (op == "boundary") {
    #pragma omp parallel for num_threads(threads)
    for (size_t i = 0; i < g.size(); i++)
      out[i] = chkNULL(GEOSBoundary_r(hGEOSCtxt[omp_get_thread_num()], g[i]));
  } else if (op == "convex_hull") {
    #pragma omp parallel for num_threads(threads)
    for (size_t i = 0; i < g.size(); i++)
      out[i] = chkNULL(GEOSConvexHull_r(hGEOSCtxt[omp_get_thread_num()], g[i]));
//  } else if (op == "unary_union") { // -> done by CPL_geos_union()
//    for (size_t i = 0; i < g.size(); i++)
[...]

PR #613, opened 2018-01-11
scaffold + parallelize st_is_simple

jeffreyhanson:
After I started working on this PR, I realised that I would need to set up some scaffolding for OpenMP. So, I thought it might be a good idea to make an initial pull request that sets up the scaffold for parallel processing and implements parallel processing for a single function. This way we can verify that that the code/strategies used here actually work. Additionally, if anyone can think of better naming conventions for functions or files, this can be easily changed now before I start rolling out parallel versions of other functions.

So, this pull request attempts to add parallel processing to the `st_is_simple` function. While it is unlikely that this function would substantially benefit from parallel processing, it provides a self-contained minimal example that uses the _GEOS C_ API. 

The specific features of this pull request include:
1. `st_is_simple` and `CPL_is_simple` functions now have a `threads` arguments. Additionally, the example for `st_is_simple` now includes an example for using two threads.
2. The file _R/parallel.R_ is provided to define helper functions related to parallel processing. This file contains the functions:
    - `is_parallel_available()` which returns a logical value indicating if parallel functionality is available.
    - `number_processors()` which returns the number of logical processors on the system (similar to `parallel::detectCores(TRUE)`). If OpenMP is not available, then this throws an error.
    - `is_valid_thread_number(threads, stopiffalse = TRUE)` which checks if an object passed to the argument `threads` is a suitable integer for setting threads for use in `st_is_simple`. This is a hidden function that is not exposed to the user.
3. The _src/openmp.h_ and _src/openmp.cpp_ files contain the functions `CPL_number_processors` (used by `number_processors`) and `CPL_openmp_version` (returns OpenMP version). Additionally, because many systems do not provide OpenMP, it defines the `omp_get_thread_num` and `omp_get_max_threads` functions if OpenMP is not available, so that the package will offer single-threaded functionality when OpenMP is not available (which can be simulated by commenting out lines 638--639 from _configure.ac_ and running `autoconf` in the terminal).
4. The _configure.ac_ file now includes instructions for determining the compiler flags for enabling OpenMP (since this can differ among compilers). Thanks @edzer for raising the point about compilation environments in issue #661. [The CRAN manual](https://cran.r-project.org/doc/manuals/R-exts.html#OpenMP-support) encourages the use of a  _configure_ file for C++11 projects, so I have copy and pasted the relevant parts from the [_configure.ac_ file](https://github.com/zhangh12/ARTP2/blob/master/configure.ac) in the [ARTP2 R package](https://cran.r-project.org/package=ARTP2). This appears to work but I do not have any prior experience with _autoconf_, so it would be great if someone else could verify that this is correct. Also, I am not sure what the preferred way to acknowledge copy and pasted code is in this project, so I have left this for others to address.
5. The `.onAttach` function now prints the linked version of OpenMP (using `CPL_openmp_version`) to help with debugging the OpenMP-related issues that users will inevitably encounter.
6. The `sf_extSoftVersion` function now includes the linked version of OpenMP (using `CPL_openmp_version`).
7. The _tests/testthat/test_parallel.R_ file contains unit tests for `number_processors` and `is_parallel_available`. It also contains an additional test that compares an output from a single-threaded run of `st_is_simple` to a multi-threaded run of `st_is_simple` (if parallel processing is available).

Let me know if anyone has any questions or suggestions for improving this.


jeffreyhanson:
Ok, I think I'm happy with this PR for the moment.

Although it doesn't currently pass CRAN checks on Travis due to set-up issues that aren't related to this PR (i.e. it fails to install MongoDB due to problems with apt-get), this PR passes CRAN checks locally on my computer (Ubuntu Xenial; where OpenMP is available) and also AppVeyor (where OpenMP is not available). 

If it is agreed that OpenMP is worth pursuing (see #611), then it would be great if one of the package maintainers could take a look at this and see what they think. Otherwise, if it is agreed that OpenMP is not worth pursuing, then please feel free to close this PR.

PR #615, opened 2018-01-12
Parallel processing using Intel TBB

jeffreyhanson:
Since the _OpenMP_ library may not be a viable solution for adding parallel processing to _sf_, this PR adds the scaffolding for parallel processing using the [_Intel TBB_ library](https://www.threadingbuildingblocks.org). It also implements a parallelized version of `st_is_simple` to provide a minimal example of the usage. Many thanks to @kendonB for suggesting this idea.

Briefly, this PR has added the [_RcppParallel R_ package](http://rcppcore.github.io/RcppParallel/index.html) to the [Imports and Linking fields in the DESCRIPTION to make the TBB library available to c++ code](http://rcppcore.github.io/RcppParallel/index.html#r_packages). It has also updated the _configure.ac_, _configure_, and _Makevars.win_ files to add the neccesscary compiler flags. This PR also includes `CPL_tbb_version()`, `CPL_number_processors()`, `number_processors()`, `is_parallel_available()`, and `is_valid_thread_number` helper functions (similar in spirit to the helper functions defined in GH-613).

The main _TBB_ function of interest is `tbb::parallel_for(std::size_t start, std::size_t end, lambda expression)` ([see here for more information](https://software.intel.com/en-us/node/506153)). Since [_TBB_ does not provide a function to obtain a sequential integer thread identifier](https://software.intel.com/en-us/forums/intel-threading-building-blocks/topic/303617) (e.g. like [`omp_get_thread_num()` in _OpenMP](https://gcc.gnu.org/onlinedocs/libgomp/omp_005fget_005fthread_005fnum.html)), we must instead use [`std::this_thread::get_id()`](http://en.cppreference.com/w/cpp/thread/get_id) to obtain a unique thread identifier. To help with managing thread-specific instances of `GEOSContextHandle_t` objects, I have implemented a `GEOSContextHandleManager` class  which can be used to instatiate, query, and destroy thread-specific `GEOSContextHandle_t` objects (defined in _src/geos.cpp_). Additionally, [since _TBB_ may not available for all systems](http://rcppcore.github.io/RcppParallel/tbb.html#portability), this PR includes extra code so that it should work when _TBB_ is not available. Specifically, I have provided a wrapper `parallel_for_loop` function which uses `tbb::parallel` when _TBB_ is available ([defined using the `RCPP_PARALLEL_USE_TBB` macro](http://rcppcore.github.io/RcppParallel/tbb.html#portability)) or `std::for_each` when _TBB_ is not available (code implemented in _src/tbb.h_).

I would consider this PR an alterative to GH-613 which uses OpenMP, as I don't think it would make much sense to support multiple parallelization libraries.


edzer:
I can install this, but with a small test I don't see any speedup:
```r
library(sf)
# Linking to GEOS 3.5.1, GDAL 2.1.2, proj.4 4.9.3, TBB 8000
demo(nc, ask = FALSE, echo = FALSE)
# Reading layer `nc.gpkg' from data source `/home/edzer/R/x86_64-pc-linux-gnu-library/3.4/sf/gpkg/nc.gpkg' using driver `GPKG'
# Simple feature collection with 100 features and 14 fields
# Attribute-geometry relationship: 0 constant, 8 aggregate, 6 identity
# geometry type:  MULTIPOLYGON
# dimension:      XY
# bbox:           xmin: -84.32385 ymin: 33.88199 xmax: -75.45698 ymax: 36.58965
# epsg (SRID):    4267
# proj4string:    +proj=longlat +datum=NAD27 +no_defs
l = vector("list", length = 100000)
for (i in 1:100000)
	l[[i]] = st_geometry(nc)
gg = do.call(c, l)
length(gg)
# [1] 10000000
object.size(gg)
# 10676802248 bytes
system.time(x <- st_is_simple(gg))
#    user  system elapsed 
#  64.768   2.860  67.868 
system.time(x <- st_is_simple(gg, threads=2))
#    user  system elapsed 
#  75.984   1.440  71.170 
system.time(x <- st_is_simple(gg, threads=4))
#    user  system elapsed 
#  90.948   1.412  71.417 
```
What am I overlooking? Ubuntu 17.04; `number_processors()` gives 4.

jeffreyhanson:
I don't think you've overlooked anything. I'm getting similar results on my computer (also with 4 cores), and I've double checked that multiple threads are being generated, so I think `GEOSisSimple_r` might just be really fast so that it doesn't substantially benefit from parallel processing (unless perhaps lots of cores are available such that the overheads for managing threads don't exceed the benefits)?

I ran this code for the OpenMP PR, and it looks like the elapsed time increases with the number of threads used. I think the set difference functions (`st_intersection`, `st_difference`, `st_sym_difference`)  or `lwgeom::st_make_valid` would provide more motivating examples because they can take much longer to run, but I think I'll wait until it's decided whether OpenMP or TBB (or another parallel library?) is the solution moving forward before I start paralellizing other functions.

```
devtools::load_all()
# Linking to GEOS 3.5.1, GDAL 2.2.2, proj.4 4.9.2, OpenMP 201307
demo(nc, ask = FALSE, echo = FALSE)
# Reading layer `nc.gpkg' from data source `/home/jeff/GitHub/sf/inst/gpkg/nc.gpkg' using driver `GPKG'
# Simple feature collection with 100 features and 14 fields
# Attribute-geometry relationship: 0 constant, 8 aggregate, 6 identity
# geometry type:  MULTIPOLYGON
# dimension:      XY
# bbox:           xmin: -84.32385 ymin: 33.88199 xmax: -75.45698 ymax: 36.58965
# epsg (SRID):    4267
# proj4string:    +proj=longlat +datum=NAD27 +no_defs
n = 10000
l = vector("list", length = n)
for (i in 1:n)
  l[[i]] = st_geometry(nc)
gg = do.call(c, l)
length(gg)
# [1] 1000000
object.size(gg)
# 1067682248 bytes
system.time(x <- st_is_simple(gg))
#   user  system elapsed 
#  6.524   0.148   6.675 
system.time(x <- st_is_simple(gg, threads=2))
#   user  system elapsed 
#  7.572   0.048   7.111 
system.time(x <- st_is_simple(gg, threads=4))
#   user  system elapsed 
#  9.736   0.104   8.375 
```

PR #616, opened 2018-01-15
Parallel cl

rsbivand:
This indicates the code that generated the output in the issue, but isn't finished - st_union.sfg probably shouldn't be there - the st_union.sfc shows the bare-bones parallel approach.


rsbivand:
Changed case to binary union. Question - how to tell roxygen to add parallel to Suggests:?

LDalby:
@rsbivand: If you are using devtools I think this line should do it:
`devtools::use_package("parallel", "Suggests")`

edzer:
Currently, `DESCRIPTION` is manually maintained.

rsbivand:
Corrected socket parallel using trick from boot::boot to realize the y promise. Added internal timing - the parallel approaches using mclapply and parLapply are viable, but for the test case (see issue #611) with a 0.5GB output object and 80000 union objects, catenating the output is the main source of time cost.

PR #645, opened 2018-02-21
Fix #644 - failing binary ops on tbl_dfs with common column names

### LINKED ISSUE #644: binary operations fail on tibble sf objects when they have common column names
When using a binary operation such as `st_interesction` on `sf` objects that are `tibbles` fails if the two objects have column names in common:

``` r
library(sf)
#> Linking to GEOS 3.6.2, GDAL 2.2.3, proj.4 4.9.3

pol1 <- st_sfc(st_polygon(list(cbind(c(0,3,3,0,0),c(0,0,3,3,0)))))
pol2 <- pol1 + 1
sf1 <- st_sf(id = 1, pol1)
sf2 <- st_sf(id = 2, pol2)

st_intersection(sf1, sf2) 
#> Warning: attribute variables are assumed to be spatially constant
#> throughout all geometries
#> Simple feature collection with 1 feature and 2 fields
#> geometry type:  POLYGON
#> dimension:      XY
#> bbox:           xmin: 1 ymin: 1 xmax: 3 ymax: 3
#> epsg (SRID):    NA
#> proj4string:    NA
#>   id id.1                           pol1
#> 1  1    2 POLYGON ((1 3, 3 3, 3 1, 1 ...

# Convert to tibbles
sf1 <- st_as_sf(tibble::as_tibble(sf1))
sf2 <- st_as_sf(tibble::as_tibble(sf2))

st_intersection(sf1, sf2)
#> Warning: attribute variables are assumed to be spatially constant
#> throughout all geometries
#> Error: Column `id` must have a unique name
```

This is happening in `geos_op2_df()` here: https://github.com/r-spatial/sf/blob/2de29bf9c810be314c226d2519d6f60459ba8b2e/R/geom.R#L748, where `dplyr::tbl_df()` throws an error when the `data.frame` has duplicated column names. (Incidentally `dplyr::tbl_df()` should probably be replaced by `tibble::as_tibble()` as the former is deprecated).

I think it could be fixed by replacing `df = cbind(df, y[idx[,2], , drop = FALSE])` with `df = data.frame(df, y[idx[,2], , drop = FALSE])` since `cbind.data.frame()` is just a wrapper for `data.frame(..., check.names = FALSE)`. Using `data.frame()` instead of `cbind()` would fix the column names before it's passed into `dplyr::tbl_df()`.

If you think that's a good route to go I can submit a PR with a new test and a fix.

ateucher:
I also took the liberty of replacing `dplyr::tbl_df()` with `tibble::as_tibble()`.

Apologies for the extraneous whitespace changes - that's RStudio automatically stripping trailing whitespace - didn't notice until I made the PR. I can revert them if you like.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,8 @@
 
 * add `st_as_sfc.raw` method
 
+* Bugfix: binary operations (`st_intersection`, `st_difference`, etc) no longer fail when operating on data frames of class `"tbl_df"` with common column names; #644
+
 # version 0.6-0
 
 * add `pillar` to Imports: to provide method for printing WKT geometries in tibbles
--- a/R/geom.R
+++ b/R/geom.R
@@ -743,9 +743,9 @@ geos_op2_df = function(x, y, geoms) {
 		warning("attribute variables are assumed to be spatially constant throughout all geometries",
 			call. = FALSE)
 	if (inherits(x, "tbl_df")) {
-		if (!requireNamespace("dplyr", quietly = TRUE))
-			stop("package dplyr required: install first?")
-		df = dplyr::tbl_df(df)
+		if (!requireNamespace("tibble", quietly = TRUE))
+			stop("package tibble required: install first?")
+		df = tibble::as_tibble(df)
 	}
 	df[[ attr(x, "sf_column") ]] = geoms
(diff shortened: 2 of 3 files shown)


edzer:
Thanks, Andy!

Issue #653, opened 2018-02-27
WKT axis order and CRS:84


dblodgett-usgs:
I hesitate to bring this up, but there's been some discussion of what the "right" thing to do is regarding axis order and use of EPSG:4326 with WKT geometry in a project I am taking part in.

_Technically_, when writing out WKT, if 4326 is included as the SRID, the axis order should reflect lat - lon. _Practically_, (and this is where I might get some 🔥) we should always use lon - lat axis order to avoid the confusion and misinterpretation introduced by the few clients that us the lat - lon order. I'm not going to suggest changing to y/x axis order for particular CRSs by default, but it would be nice to:

1. have optional "strict" treatment of axis order through a T/F flag on `st_as_text()` and 
2. allow specification of a lon/lat CRS in the WKT string. 

I'm not sure how this would work with the `SRID=####` approach used with the WKT in `sf` as there isn't an EPSG code for the [`urn:ogc:def:crs:OGC:1.3:CRS84 `](http://www.opengis.net/def/crs/OGC/1.3/CRS84) but it would be really nice if there were a way to emit the WKT string with an explicitely lon/lat axis order CRS. Maybe the approach taken in [GeoSPARQL described here](https://www.w3.org/2015/spatial/wiki/Coordinate_Reference_Systems) would be an option?

so... some code to illustrate

This all works now:
```R
lat_lon <- sf::st_sf(id = 1, geom = sf::st_sfc(sf::st_point(c(-90,40))))
sf::st_crs(lat_lon) <- sf::st_crs(4326)
sf::st_as_text(lat_lon$geom)
# [1] "POINT (-90 40)"
sf::st_as_text(lat_lon$geom, EWKT = TRUE)
# [1] "SRID=4326;POINT (-90 40)"
```
This is not the current behavior, but illustrates what might be done:
```R
sf::st_as_text(lat_lon$geom, EWKT = TRUE, STRICT_AXIS_ORDER = TRUE)
# [1] "SRID=4326;POINT (40,-90)"
sf::st_as_text(lat_lon$geom, EWKT = TRUE, STRICT_AXIS_ORDER = FALSE)
# [1] "<http://www.opengis.net/def/crs/OGC/1.3/CRS84> POINT (-90, 40)"
```

That all said, I would be just as happy with the statement that the axis order of lat/lon CRSs is always lon - lat regardless of the SRID of the data. But a statement to that affect should probably be included in the doc for `st_as_text()`?

edzer:
[...]

PR #654, opened 2018-02-28
Add documentation of WKT and EWKT conventions used in st_as_text.

dblodgett-usgs:
Per discussion over in #653.

Note that I removed details line that `EWKT=FALSE` would suppress SRID. `EWKT=FALSE` is the default for `sfc` and `TRUE` for `sfg`. Since `sfg` doesn't carry a CRS, the `EWKT=TRUE` there doesn't cause the SRID to be included.


edzer:
Thanks!

PR #679, opened 2018-03-19
help file geos_binary_pred: link to sgbp

MatthieuStigler:
Documentation on the st_predicates mentions sgbp, but does not link to its help file? Just added links to that help file. 

Note I did not recompile the help file, seems like the workflow is slightly more complex than what I am used to...


edzer:
Makes sense - thanks!

PR #684, opened 2018-03-22
Allow read_sf to read non-spatial tables

ateucher:
As of commit 77a31f2989a1d217438fb629ac34d7ef31baa9c2, `st_read()` can read non-spatial tables (and throws a warning). This extends that ability to `read_sf()`, and adds tests for both.

The file `inst/gpkg/nospatial.gpkg` is added for testing - it is 98kb and contains a 2-column, 1-row table with no geometry column.


edzer:
Thanks, Andy!

ateucher:
My pleasure!

PR #686, opened 2018-03-23
Update required gdal version & fix error

rundel:
The inclusion of gdal_utils.h requires gdal > 2.1, this was not reflected in the DESCRIPTION file.

Additionally, the ordering of version check and #include "gdal_utils.h" resulted in the compiler failing at the include rather than the more useful error message. I've just swapped those lines and included gdal_version.h to ensure the macro values are defined.

It seems like this check should be moved into configure.ac but I'm not that ambitious at the moment.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -54,7 +54,7 @@ Suggests:
 	rmarkdown
 LinkingTo: Rcpp
 VignetteBuilder: knitr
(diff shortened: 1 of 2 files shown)


edzer:
Thanks for notifying! An alternative solution would be to make the whole gdal_utils stuff conditional to 2.1 availability, and keep 2.0 as requirement.

rundel:
That would also be a good solution, I just got bit by this when updating on some machines running Fedora 23. Having an error message only about gdal_utils.h was non-obvious about the underlying cause.

PR #708, opened 2018-04-17
Optimize st_make_grid

dbaston:
I was doing some work with `st_make_grid` and noticed a couple of bottlenecks in the profiler. The changes included in this PR reduce `st_make_grid` runtime by 50% in my testing.

### DIFF
--- a/R/sfg.R
+++ b/R/sfg.R
@@ -1,6 +1,5 @@
 # dim: what does the third dimension, if present, refer to? (XYZ or XYM)
 getClassDim = function(x, d, dim = "XYZ", type) {
-	stopifnot(d > 1 && d < 5)
 	type = toupper(type)
 	if (d == 2)
 		c("XY", type, "sfg")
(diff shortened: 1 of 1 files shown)


R/sfg.R
```diff
@@ -26,14 +25,17 @@ MtrxSet = function(x, dim = "XYZ", type, needClosed = FALSE) {
...
 		lapply(x, is_numeric_matrix)
-		NotClosed = function(y) any(head(y, 1) != tail(y, 1))
+		NotClosed = function(y) any(y[1, ] != y[nrow(y), ])
 		if (needClosed && any(vapply(x, NotClosed, TRUE)))
 			stop("polygons not (all) closed")
-		structure(x, class = getClassDim(x, nc, dim, type))
-	} else
-		structure(x, class = getClassDim(x, nchar(dim), dim, type))
+		class(x) = getClassDim(x, nc, dim, type)
+		return(x)
```
edzer: I can't believe that this brings improvement; I can understand you find this more readable, but since the rest of `sf` also uses `structure` and avoids returns()s, please leave this the way it was.

R/sfg.R
```diff
@@ -1,6 +1,5 @@
 # dim: what does the third dimension, if present, refer to? (XYZ or XYM)
 getClassDim = function(x, d, dim = "XYZ", type) {
-	stopifnot(d > 1 && d < 5)
```
edzer: if not here, where is this sanity check done? What if this is not true?

R/sfg.R
dbaston: I made the change for performance. Using `structure` instead of `class` increases `st_make_grid` runtime by 15-30%, which isn't that surprising if you look at the implementation of `structure`. Obviously you make the call on the performance/style tradeoff.

R/sfg.R
dbaston: The same check is done on the last line of `getClassDim`.

edzer: Then it doesn't make sense, indeed.

R/sfg.R
edzer: OK, that's a good reason.

edzer:
Thanks!

Issue #700, opened 2018-04-11
Performance: import of many points is much faster in sp than in sf


pbaylis:
I'd like to import many points as quickly as possible. It was taking longer than expected in `sf`, so I tried `sp` and was surprised to find that it was much quicker. I'm not sure if this is issue-appropriate, but perhaps this issue is unique to me? I'm running OS X, sf 0.6.0.

```
library(sf)
library(sp)

N <- 1000000

df <- data.frame(lng = runif(N, -120, -100), 
                 lat = runif(N, 30, 48))

tic <- Sys.time()
df_sf <- st_as_sf(df, coords = c("lng", "lat"), crs = "+proj=longlat +datum=WGS84")
print(Sys.time() - tic)
# Time difference of 22.33289 secs

tic <- Sys.time()
coordinates(df) <- ~lng+lat
proj4string(df) <- "+proj=longlat +datum=WGS84"
print(Sys.time() - tic)
# Time difference of 0.8412709 secs
```

rsbivand:
In **sp**, SpatialPoints objects hold the coordinates in a matrix, so:

```
mat <- cbind(lng = runif(N, -120, -100), lat = runif(N, 30, 48))
system.time(sp_mat <- SpatialPoints(mat, proj4string=CRS("+proj=longlat +datum=WGS84")))
#   user  system elapsed 
#  0.071   0.000   0.071 
```

An **sf** `MULTIPOINT` geometry does the same:

```
 system.time(df_sf <- st_sfc(st_multipoint(mat), crs = "+proj=longlat +datum=WGS84"))
#   user  system elapsed 
#  0.029   0.000   0.029 
```

but casting to single POINT objects is costly:

```
system.time(df_sf1 <- st_cast(df_sf, "POINT"))
#   user  system elapsed 
# 22.808   0.000  22.881 
```
(Unchecked) **sf** objects will take less time in topological predicates than **sp** objects because they are already simple features.

Writing and reading are equivalent:

```
td <- tempdir()
system.time(st_write(df_sf1, file.path(td, "sf_write.gpkg"), driver="GPKG"))
#Writing layer `sf_write' to data source `/tmp/RtmploLAJz/sf_write.gpkg' using driver `GPKG'
#features:       1000000
#fields:         0
#geometry type:  Point
#   user  system elapsed 
# 15.345   3.132  18.512 
system.time(df_sf2 <- st_read(file.path(td, "sf_write.gpkg")))
[...]

PR #716, opened 2018-04-20
Rcpp for st_as_sf POINTs by coords names #700

mdsumner:
Provides a modest speed-up for a points data frame with `st_as_sf(, coords = )`


mdsumner:
Also, needed some test fixes from my previous PR for this to pass!

edzer:
Thanks!

PR #725, opened 2018-04-26
Update `st_ll_sample` for regular sample on LINES

statnmap:
This update for regular sample is only for lines.

- Same code as in {sp} for `type="regular"`
- Set `offset` as random for a random starting point.
- Drop units of `l` necessary for `type="regular"`, otherwise units problem in `grp`.
     + It may be necessary to add `@importFrom units drop_units`
- If only a unique "size" is fixed, then sample will be regular over all. This means one point every x meter, whatever the length of the line. I guess this is similar to `type = random` where not the same number of points are sampled on each line. 
    + Can maybe add some information about this in the doc and propose to set multiple "size" so that the same number of point is set on each line.


R/sample.R
```diff
@@ -84,15 +84,23 @@ st_multipoints_sample = function(x, size, ..., type = "random") {
...
-	grp = split(d, cut(d, lcs, include.lowest = TRUE))
-	grp = lapply(seq_along(x), function(i) grp[[i]] - lcs[i])
-	st_sfc(CPL_gdal_linestring_sample(x, grp), crs = st_crs(x))
+st_ll_sample = function (x, size, ..., type = "random") 
+{
+  if (isTRUE(st_is_longlat(x))) {
+    message_longlat("st_sample")
+    st_crs(x) = NA_crs_
+  }
+  l = drop_units(st_length(x))
```
edzer: see travis output: drop_units is not found.

R/sample.R
```diff
@@ -84,15 +84,23 @@ st_multipoints_sample = function(x, size, ..., type = "random") {
...
+  } else if (type == "regular") {
+    offset = runif(1)
+    d = ((1:size) - (1-offset))/size * sum(l)
+  } else {
+    stop(paste("type", type, "not available for LINESTRING"))
+  }
+  lcs = c(0, cumsum(l))
+  grp = split(d, cut(d, lcs, include.lowest = TRUE))
+  grp = lapply(seq_along(x), function(i) grp[[i]] - lcs[i])
+  st_sfc(CPL_gdal_linestring_sample(x, grp), crs = st_crs(x))
```
edzer: For PR's in general, try to keep the source formatting, so that we can see what you actually changed.

statnmap:
It was not really easy to find where you call the `@importFrom` as I usually call them for each function, even if there are repetitions...  
- Whatever, I modified it and run `devtools::document` to update the Namespace. I do not really know why this changed the order of lines in " R/RcppExports.R" and "src/RcppExports.cpp" but I pushed the changes.
- I also modified the indentation to be in accordance with the rest of the code

edzer:
Thanks for the PR! Things should now have been merged; please test.

statnmap:
Tested. It works !
Next step: implement it for polygons as well...

edzer:
That one is more tricky! Looking forward to a PR,

edzer:
Now available in branch `hex`.

Issue #628, opened 2018-01-26
writing gpkg and sqlite on samba shares fails


rnuske:
I'm lost. I can't figure how to write GPKG or sqlite files with `st_write()`. Examples from #470 fail as well (see below). 
Files are written but are empty. st_write() does not tell that it failed to write GPKG. GPKG takes way longer to crash than sqlite.

```
library(sf)

n = 1000
d = data.frame(a = 1:n, X = rnorm(n,1,1), Y = rnorm(n,1,1))
mp1 = st_as_sf(d, coords = c("X","Y"))

st_write(mp1, dsn="dat1.gpkg")
```

```
Writing layer `dat1' to data source `dat1.gpkg' using driver `GPKG'
features:       1000
fields:         1
geometry type:  Point
Warning messages:
1: In CPL_write_ogr(obj, dsn, layer, driver, as.character(dataset_options),  :
  GDAL Error 1: sqlite3_exec(COMMIT) failed: database is locked
2: In CPL_write_ogr(obj, dsn, layer, driver, as.character(dataset_options),  :
  GDAL Error 1: sqlite3_exec(PRAGMA synchronous = OFF) failed: Safety level may not be changed inside a transaction
3: In CPL_write_ogr(obj, dsn, layer, driver, as.character(dataset_options),  :
  GDAL Error 1: sqlite3_exec(BEGIN) failed: cannot start a transaction within a transaction
```

In a new R Session with the above data set
```
st_write(mp1, dsn="dat2.sqlite")
```

```
Creating dataset dat2.sqlite failed.
Error in CPL_write_ogr(obj, dsn, layer, driver, as.character(dataset_options),  : 
  Creation failed.
In addition: Warning message:
In CPL_write_ogr(obj, dsn, layer, driver, as.character(dataset_options),  :
  GDAL Error 1: sqlite3_exec(CREATE TABLE geometry_columns (     f_table_name VARCHAR,      f_geometry_column VARCHAR,      geometry_type INTEGER,      coord_dimension INTEGER,      srid INTEGER,     geometry_format VARCHAR );CREATE TABLE spatial_ref_sys        (     srid INTEGER UNIQUE,     auth_name TEXT,     auth_srid TEXT,     srtext TEXT)) failed: database is locked
```

**Versions:**
geo-packages are from ubuntugis-unstable
```
> sf_extSoftVersion()
          GEOS           GDAL         proj.4         lwgeom GDAL_with_GEOS 
[...]

PR #736, opened 2018-05-07
add extension to tempfile to calm GDAL

rnuske:
add an extension to the tempfile to get rid of the following GDAL message (cf #628)
```GDAL Message 1: The '' extension is not allowed by the GPKG specification, which may cause compatibility problems```


edzer:
Thanks!

PR #762, opened 2018-05-30
Change autoconf tests to use C++11

rundel:
I just had a failure when attempting to install sf on my macbook. Homebrew recently updated gdal to 2.3.0 which apparently requires C++11 within some of its headers. This was causing a failure with the gdal linking checks.

```
checking GDAL: linking with --libs and --dep-libs... no
In file included from gdal_test.cpp:1:
In file included from /usr/local/Cellar/gdal/2.3.0/include/gdal.h:45:
/usr/local/Cellar/gdal/2.3.0/include/cpl_port.h:187:6: error: Must have C++11 or newer.
#    error Must have C++11 or newer.
     ^
1 error generated.
In file included from gdal_test.cpp:1:
In file included from /usr/local/Cellar/gdal/2.3.0/include/gdal.h:45:
/usr/local/Cellar/gdal/2.3.0/include/cpl_port.h:187:6: error: Must have C++11 or newer.
#    error Must have C++11 or newer.
     ^
1 error generated.
```

Since a C++ standard isn't specified my compiler was defaulting to C++98 and hence the failure.

I've updated configure.ac to construct CXX from R CMD config's CXX11 and CXX11STD and also added C++11 to SystemRequirements in the DESCRIPTION file (Writing R Extensions states that either this **or** CXX_STD=CXX11 in Makevars should be included  but that seems a  bit silly).

It may be worth including an explicit test for C++11 compatibility within configure.ac, I have not tested on a system with an older compiler -- it seems like everything will fail at some point  but the error messages may be confusing.

This is also likely going to affect rgdal at somepoint (@rsbivand).

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -58,7 +58,7 @@ Suggests:
 	rmarkdown
 LinkingTo: Rcpp
 VignetteBuilder: knitr
-SystemRequirements: GDAL (>= 2.0.1), GEOS (>= 3.4.0), PROJ.4 (>= 4.8.0)
+SystemRequirements: C++11, GDAL (>= 2.0.1), GEOS (>= 3.4.0), PROJ.4 (>= 4.8.0)
 License: GPL-2 | MIT + file LICENSE
 URL: https://github.com/r-spatial/sf/
 BugReports: https://github.com/r-spatial/sf/issues/
--- a/configure
+++ b/configure
@@ -2045,10 +2045,25 @@ fi
 
 RBIN="${R_HOME}/bin/R"
 
+RVER=`"${RBIN}" --version | head -1 | cut -f3 -d" "`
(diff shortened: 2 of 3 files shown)


rsbivand:
Thanks, @rundel  - a CentOS 7 user reported this yesterday, on [this thread](https://stat.ethz.ch/pipermail/r-sig-geo/2018-May/026591.html). So a similar step for **rgdal** is needed, conditional on GDAL being >= 2.3.0.

rundel:
rgdal seemed like a slightly more complicated case since (I believe) you don't require C++11

edzer:
Thanks, Colin!

rsbivand:
As of r780, CXX11 conditional testing added to **rgdal** (have to ship the m4 macro).

PR #767, opened 2018-06-05
Complete code coverage for database interface

etiennebr:
I've had to refactor a few things and also remove the geom_column argument, but I'd say that the coverage is pretty good now. 

Related #558


edzer:
Great, thanks @etiennebr !

PR #687, opened 2018-03-23
Update .travis.yml, use tic and automatic deployment of pkgdown site

pat-s:
#### Changes

* move ppa installations to `addons`
* build on oldrelease, release & devel (failures are allowed for r-devel)
* cache packages a
* automatic deployment of `pkgdown` site on every commit using `tic`. Preview [here](https://pat-s.github.io/sf/docs/).
* example build: https://travis-ci.org/pat-s/sf/builds/357620507

#### Notes

* `lwgeom` is installed with force on every build using its shipped binary as linking to the `postgis` installation sometimes causes issues
* pkgdown is deployed to `docs/` subdirectory of the `master` branch for every commit to `master`. 
* To use the `pkgdown` deployment, you need to call `travis::use_travis:deploy()` once to set up a ssh key pair with travis and change your Github pages settings in the repo to serve from `gh-pages`.
* Why `tic`? -> https://ropenscilabs.github.io/tic/; CI-agnostic workflow definitions and simplified pkgdown deployment

(You may want to [squash all commits into a single one](https://help.github.com/articles/configuring-commit-squashing-for-pull-requests/))

PR #780, opened 2018-06-27
Use older tic version to fix deployment of docs

pat-s:
Works: https://travis-ci.org/pat-s/sf/jobs/397207994

The most recent `tic` version seems to be broken; https://github.com/ropenscilabs/tic/issues/56

cross-ref https://github.com/r-spatial/sf/pull/687#issuecomment-400220114


edzer:
That worked - thanks a lot for sorting this out!

PR #795, opened 2018-07-13
Travis: Fix false positives in R cmd check and deployment issues

pat-s:
Both forks solves the issues above. Tested in `mlr`.
Hotfix until PRs are merged in both repos.


pat-s:
Hm, apparently it does still not catch the test failures in `sf`. Investigating. Please don't merge yet.

pat-s:
Ok, works now as expected. The error is catched!

edzer:
Travis still fails on all three platforms.

pat-s:
Yes, but that's how it should be because it catches errors in the tests. These errors were not catched before - > false positives green travis build. 

The error which is related to an outdated GEOS version should be solved in another PR.

edzer:
Excellent - thanks a lot!

PR #797, opened 2018-07-14
Add support for non-sf workflows

etiennebr:
Essentially enables to `st_write` data frames. This will enable workflows with non-sf objects (but still probably a geometry column). It is with the ultimate objective of making the `dbplyr` interface smoother.


edzer:
Thanks!

PR #821, opened 2018-08-14
Avoid constructing prepared geometries that are not used

dbaston:
This PR modifies predicate functions (intersects, contains, etc.) to avoid creating a prepared geometry that will never be used.


edzer:
Thanks -- makes sense!

PR #823, opened 2018-08-20
geos 3.7.0rc1 workaround; GEOS_VERSION_PATCH is 0rc1

rsbivand:
When GEOS_VERSION_MINOR is 6, the test for PATCH >= 1 fails in 3.7.0rc1 because PATCH is 0rc1. Testing for MINOR first avoids looking at PATCH.

### DIFF
--- a/src/geos.cpp
+++ b/src/geos.cpp
@@ -5,8 +5,10 @@
 # if GEOS_VERSION_MINOR >= 5
 #  define HAVE350
 # endif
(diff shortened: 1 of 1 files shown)


edzer:
Thanks!

rsbivand:
The details are a [malformed sed RE](https://lists.osgeo.org/pipermail/geos-devel/2018-August/008446.html), and subsequent postings, and [trac ticket](https://trac.osgeo.org/geos/ticket/887).

PR #841, opened 2018-09-09
Use latest version of tic

pat-s:
@krlmlr did a lot of work lately and fixed most of the issues. I cannot fully guarantee that this version is fully stable now but we are close to a CRAN release.

@edzer You can use this branch to solve the test errors. The `tic` implementation looks fine. 

I also tidied the DESCRIPTION using `usethis::use_tidy_description()`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,7 +53,7 @@ before_install:
   - psql -d postgis -c "GRANT ALL PRIVILEGES ON ALL TABLES IN SCHEMA public TO travis"
   - createdb empty
   - psql -d empty -c "CREATE EXTENSION postgis;"
-  - R -q -e 'install.packages("devtools"); devtools::install_github("ropenscilabs/tic"); tic::prepare_all_stages()'
+  - R -q -e 'install.packages("remotes"); remotes::install_github("ropenscilabs/tic"); tic::prepare_all_stages()'
 
 after_success:
   - dropdb postgis
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,67 +1,138 @@
 Package: sf
-Version: 0.6-4
 Title: Simple Features for R
-Description: Support for simple features, a standardized way to encode spatial vector data. Binds
-	to 'GDAL' for reading and writing data, to 'GEOS' for geometrical operations, and to 'PROJ' for projection
-	conversions and datum transformations.
-Authors@R: c(person("Edzer", "Pebesma", role = c("aut", "cre"), email = "edzer.pebesma@uni-muenster.de", comment = c(ORCID = "0000-0001-8049-7069")),
-	person("Roger", "Bivand", role = "ctb", comment = c(ORCID = "0000-0003-2392-6140")),
-	person("Etienne", "Racine", role = "ctb"),
-	person("Michael", "Sumner", role = "ctb"),
-	person("Ian", "Cook", role = "ctb"),
-	person("Tim", "Keitt", role = "ctb"),
-	person("Robin", "Lovelace", role = "ctb"),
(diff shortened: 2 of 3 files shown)


edzer:
Still some issues, looks like tic related to me: https://travis-ci.org/r-spatial/sf/jobs/428628046

pat-s:
Yes, saw it.

Works locally, hm. Do you know what the `nativeRoutines` refers to? https://travis-ci.org/r-spatial/sf/jobs/428628046#L5445

Some c++ stuff that is missing here? The error is thrown from `pkgdown::build_site()`. Atm I don't think its an tic issue but rather travis related.

krlmlr:
Maybe it would help to avoid stashing in pre-deploy: see link at https://travis-ci.org/r-spatial/sf/jobs/428628046#L5232 .

This might be a good default for `.travis.yml`, too.

edzer:
Any news here, @pat-s ? Would like to get this working.

pat-s:
Still trying, unsure why the `sf` installation in the deploy step fails.

It could be caused because its installed in a new R process and this process is missing some c++ deps, i.e. the "native routines" error. Currently trying to force installation in the same R process.

PS: The stage of the PR you just merged included a duplicate "version" field in the DESCRIPTION by accident.

edzer:
@etiennebr I now see errors coming from tests you wrote: https://travis-ci.org/r-spatial/sf/builds/432212868 could you pls take a look?

pat-s:
`new_process = F` didn't help.
Atm I have no clue where the error comes from and why it appears.
I don't think its related to *tic* (maybe *pkgdown* or travis itself).
Deployment and *pkgdown* currently works for all repos besides *sf*.

Do you know where this "nativeRoutines" error might come from? It seems to look for a dll file in `src` that can't be found.

pat-s:
@edzer Trying to run the `pkgdown` stuff manually later (without the custom tic step). Not expecting that anything changes but lets see.

If that doesn't help we can try to run this step on appveyor (to see if its a travis thing).

edzer:
The `nativeRoutines` stuff had apparently to do with r-devel;l now deploying under release, which works. Thanks!

pat-s:
ahh, all that checking and in the end its sth that trivial - well, glad that it works now (hopefully for a longer time ;)!

edzer:
I now see in various repos a break at random points with:
```
Error: HTTP error 403.
  API rate limit exceeded for 104.154.120.187. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
```
does this have anything to do with tic?

krlmlr:
Thanks, I'm seeing it too in _tibble_, but thought this is some random error.

@gaborcsardi: Do you see a reason why `GITHUB_PAT` would be ignored now when using `remotes::install_github()` ? I've seen that the message that the env var is used is no longer printed, but it should still work?

gaborcsardi:
It _should_. There was a mishap with it in the dev version, but AFAIK this was fixed  for the CRAN version. @jimhester ?

pat-s:
I assume it is related to https://github.com/klutometis/roxygen/issues/822.

We have a workaround for this already, let me do a quick test.

pat-s:
I was right. Opened PRs in the respective repos.

edzer:
Hi @pat-s , any idea what is going wrong here? https://travis-ci.org/r-spatial/sf/jobs/537456557

pat-s:
The error message looks a bit strange but I hope that #1058 solves it.  
The old (manual) approach should actually still work but the error message from the build above is new to me.

edzer:
Thanks a lot!!

Issue #833, opened 2018-08-29
Feature request: st_buffer customisation


stPhena:
First of all, thanks for developing this excellent package!
I would love to have more arguments to customise the behaviour of `st_buffer()`, especially about the shape of the joints and caps. It would be useful to not be forced to use `rgeos::gBuffer()`. I have made some research on the internet and it seems to be a frequently sought for feature.
Thanks in advance.

mdsumner:
@stPhena for discussing stuff like this I'd recommend #rstats on Twitter, and there are a few other friendly places. Feel free to ping me anytime here or elsewhere. Personally I find it very helpful to air possible changes and whether they are feasible or welcome before crafting PRs.

edzer:
Discussing this here is also fine. I have for instance no clue why anyone would want this kind of customisation for spatial analysis, or other purposes. I'd also be interested in pointers to others searching for this feature in sf, which would further help me understand the need for it.

mdsumner:
I wrote it up here, unfortunately on top of a different PR branch so I'll have to unpick it or wait until execute_sql is merged in:

https://github.com/dis-organization/sf/tree/buffer_styles

edzer:
... means we can close here, right?

PR #842, opened 2018-09-11
buffer styles

mdsumner:
* `st_buffer` now includes line end and join stylings using optional arguments `endCapStyle`, `joinStyle` and `mitreLimit`, https://github.com/r-spatial/sf/issues/833. 

I believe this is faithful to the support in `rgeos`, which I used as a guide. 

One issue is that these style examples are at the front of 'examples', which seems funny - I'm not sure how to fix apart from moving all the examples into one?


edzer:
Great work, thanks!!!

PR #846, opened 2018-09-14
st_ll_sample: Allow MULTILINESTRING of length zéro

statnmap:
In my last dataset, I had some lines of 2 points with starting point = end point. So basically lines of length zéro. This conducted to error with `st_ll_sample` used inside `st_sample`.  
Here I propose a solution to sample only one point whatever is `size` and show a `message` (that could be a `warning` too).


edzer:
Thanks!

Issue #849, opened 2018-09-21
st_read with PostGIS treats character column with all NA's as a geometry column


ateucher:
If there is a text column in a PostGIS table where every value is `NULL`, `st_read()` interprets it as a `geometry` column. This is especially surprising when the `CRS` and/or active geometry for the resulting `sf` object are set from this column. Thanks to @boshek for sleuthing this out with me!

``` r
library(RPostgreSQL)
#> Loading required package: DBI
library(sf)
#> Linking to GEOS 3.6.3, GDAL 2.3.1, PROJ 5.2.0
library(sp)
data("meuse")

## Add a character column to meuse with all `NA`s
meuse$dummy <- rep(NA_character_, nrow(meuse))
meuse_sf <- st_as_sf(meuse, coords = c("x", "y"), crs = 28992)

con <- dbConnect(PostgreSQL(),
                 host = "localhost",
                 dbname = "postgis",
                 user = "postgres")

dbSendQuery(con, "CREATE SCHEMA temp_scratch;")
#> <PostgreSQLResult>

dbWriteTable(con, c("temp_scratch", "meuse"), value = meuse_sf)
#> [1] TRUE

out <- st_read(con, query = "SELECT * FROM temp_scratch.meuse;")

# the 'dummy' column has been set to the active geometry
attr(out, "sf_column")
#> [1] "dummy"
st_crs(out)
#> Coordinate Reference System: NA

# We can get the proper one back
st_geometry(out) <- "geometry"
attr(out, "sf_column")
#> [1] "geometry"
st_crs(out)
#> Coordinate Reference System:
#>   EPSG: 28992 
#>   proj4string: "+proj=sterea +lat_0=52.15616055555555 +lon_0=5.38763888888889 +k=0.9999079 +x_0=155000 +y_0=463000 +ellps=bessel +towgs84=565.2369,50.0087,465.658,-0.406857,0.350733,-1.87035,4.0812 +units=m +no_defs"

# Run this to clean up the temporary schema
# dbSendQuery(con, "DROP SCHEMA temp_scratch CASCADE;")
```

<sup>Created on 2018-09-21 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

[...]

PR #850, opened 2018-09-21
Check for all NAs in character columns when checking for geometry columns

ateucher:
Fix for #849


edzer:
Thanks a lot, @ateucher !

PR #852, opened 2018-09-25
Fix tests

etiennebr:
See context in https://github.com/r-spatial/sf/pull/841#issuecomment-423847877


edzer:
Thanks!

edzer:
For some reason we're still failing on travis: https://travis-ci.org/r-spatial/sf/jobs/433030839#L4369 although I cannot reproduce this locally. Is this maybe because it is being run twice @pat-s ?

etiennebr:
Ok, I can reproduce the issue locally. I'll try to come up with a fix in a few hours.

Issue #834, opened 2018-08-30
Run SQL query on OGR layers 


barryrowlingson:
An SO questioner: https://stackoverflow.com/questions/52086882/issue-with-query-in-st-read/52102296#52102296 wanted to filter shapefiles with an SQL query, as you can do in `ogr2ogr` for example, but the `query` parameter is only valid for DBIObject reads in `st_read`.

I answered by showing how to build a VRT file with the SQL query, which works fine but is a bit of a pain.

Could gdal_read.cpp use the `ExecuteSQL` method on an optional passed SQL query string? I experimented and had success with putting:

	poDS->ExecuteSQL("select * from codes_postaux_region where POP2010 > 20000", NULL, NULL);

 on line 225, which successfully reduced my read data from 6000 features to 707. I'm not sure I completely understand what the ExecuteSQL method does, or what the returned OGRLayer is (see https://www.gdal.org/classGDALDataset.html#a5b65948b1e15fa63e96c0640eb6c5d7c) or what the spatial filter or dialect parameters can do (I left them NULL) but its a proof-of-concept. Real code would have to pass the query string through, changing the parameters in the C++ and the R and so on, and my Rcpp is inadequate to get this all right, so I just hardcoded one SQL test string.

I can't see any discussion here about implementing this already, apologies if this has already been excluded as a possible enhancement.

B

edzer:
Nice idea! PR welcome.

jsta:
This issue is very similar to https://github.com/hypertidy/vapour/issues/34. Maybe it would be of use here?

barryrowlingson:
There's a more complete working example in a branch of my fork here: https://github.com/barryrowlingson/sf/tree/execute_sql - I've no idea of the quality so I'm not sure I should hit the PR button. I can do if it helps, but I don't really know what I'm doing with C++...

edzer:
Looks OK; I wonder whether the `select * from nc` cannot be constructed (probably on the R side), since we know that the layer is `nc`. The real query is the where clause, right? @mdsumner 

Also, would it be possible to do selections based on geometries? Maybe not, or at least not portably, since GDAL on MacOS X does not compile against GEOS (where dynamic linking not allowed by CRAN, static not made possible by GDAL devs); @rsbivand ? Nevertheless, it might be nice-to-have on platforms that do support it?

barryrowlingson:
[...]

PR #854, opened 2018-09-26
ExecuteSQL

mdsumner:
As discussed in https://github.com/r-spatial/sf/issues/834 and mostly written by @barryrowlingson this allows `query` argument to be used by the `ExecuteSQL` function of a data set. 

An example, resulting in three fields and two features: 

```R
nc_sql = st_read(system.file("shape/nc.shp", package="sf"),
                      query = "SELECT NAME, SID74, FIPS FROM nc WHERE BIR74 > 20000")
```

It's unsuccessful on GPKG however, because the geometry column detection doesn't work unless the native geometry column is included in the select: 

```R
# it's also a bit inconvenient because layer name must be found and used explicitly 
# (multiple layers can be used with joins etc, so it overrides 'layer')
gpkg_layer <- st_layers(system.file("gpkg/nc.gpkg", package = "sf"))$name[1]
## this fails to find geometry unless we include 'geom' in the select
nc_gpkg_sql = st_read(system.file("gpkg/nc.gpkg", package = "sf"),
        query = sprintf("SELECT NAME, SID74, FIPS, geom  FROM '%s' WHERE BIR74 > 20000", gpkg_layer))

```

I can't yet see if this can be overcome easily or if it requires a new approach (?).


mdsumner:
Ugh, I've made a mess of this by not merging master first ... probably just ignore

edzer:
https://github.com/r-spatial/sf/pull/855 merges master into this. I can run your examples. LGTM!

mdsumner:
Oh great!

PR #864, opened 2018-10-16
tmap v2.* does not convert to `Spatial*` anymore, uses `sf` internally

dpprdan:
Since v2.0, tmap does not convert to `Spatial*` anymore, but uses `sf` internally (see tmap's [NEWS](https://github.com/mtennekes/tmap/blob/3e78bd8e0406f779374d155799247010cebb732d/NEWS#L15), so I deleted the relevant paragraph about the former incompatibility from the sf5 vignette.

I opted for suggesting `tmap (>= 2.0)` in DESCRIPTION. A milder approach would be to state that v2.* is preferred or else there might be limitations (i.e. edit not delete the relevant paragraph in sf5.Rmd). 

I have not yet build the vignette.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -86,7 +86,7 @@ Suggests:
     tibble (>= 1.4.1),
     tidyr (>= 0.7-2),
     tidyselect,
-    tmap
+    tmap (>= 2.0)
 LinkingTo: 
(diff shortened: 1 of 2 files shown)


edzer:
Thanks!

PR #888, opened 2018-11-07
Further optimize bounding-box calculations

dbaston:
So this is interesting.

I've been working with a shapefile of ~5500 multipolygons comprising 38000 polygons and 58000 rings. I'm using callgrind to inspect a single `st_read` call:

`R -d "valgrind --tool=callgrind" -e "library(sf); st_read('/tmp/basins_lev05.shp')"
`

The callgrind output shows that, even after the changes recently committed, `CPL_get_bbox` remains at 25% of total program runtime (compared to `CPL_read_ogr` at 43%). `CPL_get_bbox` is still calling `Rf_xlength` an astonishing 69 million times!

Some more digging pins this on the use of `operator()` for vector access instead of `operator[]`. Apparently `NumericVector::operator()` does a bounds-check (and hence, a call to `Rf_xlength`), while `NumericVector::operator[]` does not. So a simple statement like `bb(0) = std::min(bbi(0),bb(0))` results in at least 3 bounds checks.

This PR simply switches access to `operator[]`.

After this PR, `Rf_xlength` is called 341k times, and CPL_get_bbox is reduced to 3.2% of program runtime. Total time for `st_read` is reduced by about 40%.

```
# Master
                            expr      min       lq     mean   median       uq      max neval
st_read("/tmp/basins_lev05.shp") 1.768442 2.005508 2.487873 2.670088 2.725343 3.159911    10

# This PR
                            expr      min       lq     mean   median       uq      max neval
st_read("/tmp/basins_lev05.shp") 1.173727 1.390973 1.456033 1.411426 1.449768 2.001311    10


```

### DIFF
--- a/src/bbox.cpp
+++ b/src/bbox.cpp
@@ -5,7 +5,7 @@
 // [[Rcpp::export]]
 Rcpp::NumericVector CPL_get_bbox(Rcpp::List sf, int depth = 0) {
 	Rcpp::NumericVector bb(4);
-	bb(0) = bb(1) = bb(2) = bb(3) = NA_REAL;
+	bb[0] = bb[1] = bb[2] = bb[3] = NA_REAL;
 	auto n = sf.size();
 
 	switch(depth) {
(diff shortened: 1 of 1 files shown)


edzer:
Good to know!

Issue #813, opened 2018-08-03
st_sample size


WWolf:
Hello,

I am using `st_sample` on a polygon, and I realized in the documentation that the sampling size might differ from the requested size because it is sampling from the bounding box of the polygon. What I am not entirely clear is how it is possible to get *more* samples than the requested size, which I often encounter.

The following is a simple test:
```
x = st_sfc(st_polygon(list(rbind(c(0,0),c(90,0),c(90,90),c(0,0)))))
plot(x); plot(st_sample(x, 5, type = "random"), add = TRUE)
```

As for a slightly different question, what would be the best way to ensure that I get the final sample number to the requested size?

Thank you!

WWolf:
Thank you! This explains a lot!

However, the simple solution of subsetting did not work many times because the number of samples is less than the requested size for roughly half of the time. A histogram of the difference between requested size and resulting size is attached: 

![image](https://user-images.githubusercontent.com/531000/43647460-8c1be7da-96ec-11e8-83ee-789dc56c2efa.png)

My dirty way of dealing with this was to increase the number of samples such that it is very unlikely to yield less than the requested number (in my case, arbitrarily adding +30 to the `st_sample` parameter) but if there could be more elegant solution, I would be grateful!

edzer:
I haven't made a study out of this, but would be happy to hear of pointers to efficient algorithms sampling _exactly_ n points from a (set of) polygon(s). What we now have is a random process, which performs badly when the ratio of polygon area / bounding box area is small (e.g. linear elements represented as narrow polygons).

damianooldoni:
I developed this workaround by using a `while` in an ad-hoc function:
```r
get_random_pt <- function(x) {
  random_pt <- st_sample(x , size = 1, type = "random")
  while (length(random_pt) == 0) {
    random_pt <- st_sample(x , size = 1, type = "random")
  }
  return(random_pt)
}
```
Then, if you want do it for a data.frame of polygons (`polygons`) as I needed, I just wrapped this function in a `map()` and then assigning the CRS of the original polygons df (`crs_polygons <- st_crs(polygons)`) back via `st_sfc()`:
```r
library(purrr)
random_pts <- map(st_geometry(polygons), get_random_pt)
random_pts <- st_sfc(unlist(random_pts, recursive = FALSE), 
                                   crs = crs_polygons
)
```

Robinlovelace:
@damianooldoni you could probably get away with:


```r
get_random_pt <- function(x, size) {
[...]

PR #896, opened 2018-11-13
Add exact argument to st_sample

Robinlovelace:


Robinlovelace:
Note: I'm still getting some non-exact results with the latest version, e.g. from `st_sample()` `@examples`:

```r
length(p_exact); length(p_not_exact)
[1] 1998
[1] 1008
```

Robinlovelace:
OK that issue was fixed in the latest commit - it was because of this (which I accidentally changed): https://github.com/r-spatial/sf/pull/896/commits/f6628ed4d4fe85764af3e749c21796edceb48ac7#diff-ebf3a432ef2b2f7d562f69cb20ea38dfL77

Robinlovelace:
Nudge @edzer - feedback welcome on this implementation of your suggestion in https://github.com/r-spatial/sf/issues/813#issuecomment-438219926. I think it works as expected and is future-proof for additional sampling types. Travis is passing but appveyor failed due to time-out I think.

edzer:
Please add some tests, so that we retain 100% coverage.

Robinlovelace:
Will do.

Robinlovelace:
Heads-up @edzer I've added some tests. They are based on previous tests of `st_sample()`. Let me know of any way to improve these you can think of. Not sure how it performs when `length(size) > 1`.

Robinlovelace:
Hi Edzer I've written another test in a file called `test_sample.R`:

```r
context("sf: st_crs tests")

test_that("st_sample works", {
	nc = read_sf(system.file("shape/nc.shp", package="sf"))
	n = 100
	sample_default = st_sample(x = nc, size = n)
	expect_s3_class(sample_default, "sfc")
	sample_exact = st_sample(x = nc, size = n, exact = TRUE)
	expect_equal(length(sample_exact), n)
})
```

Not committed this yet. Want to get your feedback: seem like a good test?

edzer:
LGTM!

Robinlovelace:
Test added. Heads-up @edzer there are a number of other tests that could be added here, a few of which are listed below. However I think that for the purposes of the PR the test is OK (assuming it passes on Travis which is never guaranteed!). Anything else you'd like doing on this let me know.

- Test that only points inside the objects are returned
- Test different types of subsetting
- Test what happens when size is zero and other edge cases, perhaps

Robinlovelace:
Thought of more tests: with different geometry types for `x` - not tried `exact = TRUE` with lines yet.

edzer:
Thanks - I'd like to have `exact=TRUE` as default, but that means it has to work everywhere too.

Robinlovelace:
For sure. I'll plan to add some informal tests with results in comments below and think about including those as formal tests.

Robinlovelace:
Not working as expected with the multilinestring example in the documentation:

``` r
devtools::install_github("robinlovelace/sf", ref = "sample-exact-argument")
#> Skipping install of 'sf' from a github remote, the SHA1 (b7f7646f) has not changed since last install.
#>   Use `force = TRUE` to force installation
library(sf)
#> Linking to GEOS 3.6.2, GDAL 2.2.3, PROJ 4.9.3

ls = st_sfc(st_linestring(rbind(c(0,0),c(0,1))),
            st_linestring(rbind(c(0,0),c(.1,0))),
            st_linestring(rbind(c(0,1),c(.1,1))),
            st_linestring(rbind(c(2,2),c(2,2.00001))))

s_line = st_sample(ls, 80)
plot(ls)
plot(s_line, add = TRUE)
```



``` r
length(s_line)
#> [1] 4
s_line %>% st_cast("POINT")
#> Geometry set for 80 features  (with 1 geometry empty)
#> geometry type:  POINT
#> dimension:      XY
#> bbox:           xmin: 0 ymin: 0 xmax: 0.08801631 ymax: 1
#> epsg (SRID):    NA
#> proj4string:    NA
#> First 5 geometries:
#> POINT (0 0.2467622)
#> POINT (0 0.524114)
#> POINT (0 0.782803)
#> POINT (0 0.09910305)
#> POINT (0 0.3666392)
sapply(s_line, function(x) nrow(st_coordinates(x)))
#> [1] 68  5  7  0

s_line = st_sample(ls, 80, exact = TRUE)
plot(ls)
plot(s_line, add = TRUE)
```



``` r
length(s_line)
#> [1] 80
s_line %>% st_cast("POINT")
#> Geometry set for 1524 features  (with 20 geometries empty)
#> geometry type:  POINT
#> dimension:      XY
#> bbox:           xmin: 0 ymin: 0 xmax: 0.09905899 ymax: 1
#> epsg (SRID):    NA
#> proj4string:    NA
#> First 5 geometries:
#> POINT (0 0.252475)
#> POINT (0 0.4454628)
#> POINT (0 0.9109935)
#> POINT (0 0.404977)
#> POINT (0 0.6040693)
sapply(s_line, function(x) nrow(st_coordinates(x))) # currently fails
#>  [1] 67  6  7  0 67  5  4  0 61  5 10  0 65  8  3  0 68  3  5  0 66  4  6
#> [24]  0 65  8  3  0 67  3  6  0 63  7  6  0 61  8  7  0 62  8  6  0 61  7
#> [47]  8  0 69  2  5  0 62  6  8  0 62 11  3  0 68  3  5  0 61  6  9  0 73
#> [70]  2  1  0 68  2  6  0 68  2  6  0
```

<sup>Created on 2018-12-13 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

Robinlovelace:
Another test I did raised a different issue with the current implementation of `type = hexagonal` - see below. Worth raising an issue on this?

``` r
devtools::install_github("r-spatial/sf")
#> Skipping install of 'sf' from a github remote, the SHA1 (5a483935) has not changed since last install.
#>   Use `force = TRUE` to force installation
library(sf)
#> Linking to GEOS 3.6.2, GDAL 2.2.3, PROJ 4.9.3

# hexagonal example from docs:
sfc = st_sfc(st_polygon(list(rbind(c(0,0), c(1,0), c(1,1), c(0,0)))))
plot(sfc)
h = st_sample(sfc, 100, type = "hexagonal")
h1 = st_sample(sfc, 100, type = "hexagonal")
plot(h, add = TRUE)
plot(h1, col = 'red', add = TRUE)
```



``` r
c(length(h), length(h1)) # approximate!
#> [1] 101 104

# hexagonal example on real data
nc = read_sf(system.file("shape/nc.shp", package="sf"))
n = 100
s_hex = st_sample(nc, size = n, type = "hexagonal") # fails
#> although coordinates are longitude/latitude, st_sample assumes that they are planar
#> although coordinates are longitude/latitude, st_intersects assumes that they are planar
#> although coordinates are longitude/latitude, st_intersects assumes that they are planar
#> Warning in seq_len(nrow(xy)): first element used of 'length.out' argument
#> Error in seq_len(nrow(xy)): argument must be coercible to non-negative integer
s_hex_nc = st_sample(nc$geometry, size = n, type = "hexagonal") # fails
#> although coordinates are longitude/latitude, st_sample assumes that they are planar
#> although coordinates are longitude/latitude, st_intersects assumes that they are planar
#> although coordinates are longitude/latitude, st_intersects assumes that they are planar
#> Warning in seq_len(nrow(xy)): first element used of 'length.out' argument
#> Error in seq_len(nrow(xy)): argument must be coercible to non-negative integer
s_hex_exact = st_sample(nc, size = n, type = "hexagonal", exact = TRUE) # fails in same way
#> although coordinates are longitude/latitude, st_sample assumes that they are planar
#> although coordinates are longitude/latitude, st_intersects assumes that they are planar
#> although coordinates are longitude/latitude, st_intersects assumes that they are planar
#> Warning in seq_len(nrow(xy)): first element used of 'length.out' argument
#> Error in seq_len(nrow(xy)): argument must be coercible to non-negative integer
```

<sup>Created on 2018-12-13 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

edzer:
Robin, looking good -- to get this merged, please add to the documentation of param `exact` for which geometries and which sampling methods it does have an effect, so users don't get confused when they specify exact=TRUE and they still get another size out. Either do this at the parameter description or in a details section.

For regular sampling methods (including hexagonal), I doubt you can guarantee an exact output size: how would you arrange a regular sample of size 10 from the unit square?

Robinlovelace:
Hi @edzer thanks for the feedback. I think it makes sense for `exact` to only affect relevant `type` samples (random currently) and also geometry types of `x` (polygons and multi-polygons) - lines always return the correct sample size is my understanding.

I will document this behaviour and fix `exact = TRUE` so that it has no impact on hexagonal sampling types and sampling on lines - good plan. Note: it may have to wait until after :santa: comes to town though, unless I find time during the winter break.

Robinlovelace:
Heads-up @edzer I finally had a chance to look at this again. Now `exact` is true by default (as seemed sensible from discussions in #813 but, to mitigate issues with sampling types that are not random, it only affects sampling when `type == "random"` and the result has `POINT` geometries:

https://github.com/r-spatial/sf/pull/896/commits/5893fafd5fda680012c108879678b9497640f618#diff-ebf3a432ef2b2f7d562f69cb20ea38dfR71

robitalec:
Thanks @Robinlovelace and @edzer for adding this!

Issue #114, opened 2016-12-12
How to go from MULTIPOINT to just point?


Robinlovelace:
Reproducible example:

```r
x = st_line_sample(st_transform(ls, 3857), density = c(1/1000, 1/10000)) # one per km, one per 10 km
class(x)
## [1] "sfc_MULTIPOINT" "sfc"   
```

Is there a function, e.g. `as.st_points` that could  convert this into just points? If not any ideas of the best way to go about creating such a function?

Robinlovelace:
Update - works (with warning) for **sp** objects:

```r
y = as(x, "Spatial")
plot(y)
SpatialPoints(y)
```

```
class       : SpatialPoints 
features    : 112 
extent      : 0, 5565.975, -7.081155e-10, 110823.7  (xmin, xmax, ymin, ymax)
coord. ref. : NA 
Warning message:
In validityMethod(object) :
  duplicate rownames are interpreted by rgeos as MultiPoints; use SpatialMultiPoints to define these; in future sp versions this warning will become an error
```

edzer:
could `sfc_MULTIPOINT` -> `sfc_POINT` be part of [st_cast](https://github.com/edzer/sfr/issues/110)? We now see
```
> st_cast(st_sfc(st_multipoint(matrix(1:6,,2))), "POINT")
Geometry set for 1 feature 
geometry type:  POINT
dimension:      XY
bbox:           xmin: 1 ymin: 4 xmax: 3 ymax: 6
epsg (SRID):    NA
proj4string:    NA
POINT(1 4) 
Warning message:
In chk(x, "MULTIPOINT") :
  object does not have length one: casting causes information loss
```
but maybe one could return all points? Or an `sf` object with an integer attribute denoting the original `ID` (nr) in the `sfc_MULTIPOINT`?

Robinlovelace:
I think returning all points would be good - I imagine the saved `ID` could be useful in some instances. Not sure if that reply is useful...

I found a way to stop the warning in **sp** btw - just remove the repeated matrix row names:

https://github.com/ropensci/stplanr/pull/164/commits/52734f77395960c0e67a75f8919ee996b861e887#diff-c451334402a5623377b6bfc78472163dR73

Robinlovelace:
Note also: I think I found a bug in **sp**: `spTransform()` seems to fail on multipoint objects...

mdsumner:
[...]

PR #917, opened 2018-12-12
WIP: Fix corner case for st_cast()

krlmlr:
Otherwise:

``` r
library(sf)
#> Linking to GEOS 3.6.2, GDAL 2.2.3, PROJ 4.9.3

st_cast(st_multipoint(), "POINT")
#> Warning in st_cast.MULTIPOINT(st_multipoint(), "POINT"): point from first
#> coordinate only
#> Error in is.numeric(x): subscript out of bounds
```

<sup>Created on 2018-12-12 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

From a quick glance at the code, there are many more such corner cases lurking in the various `st_cast()` implementations. Am I missing a better alternative?

### DIFF
--- a/R/cast_sfg.R
+++ b/R/cast_sfg.R
@@ -118,7 +118,15 @@ st_cast.MULTIPOINT <- function(x, to, ...) {
          POLYGON = st_polygon(list(unclass(ClosePol(x)))), 
          LINESTRING = st_linestring(unclass(x)),
          ## loss, drop to first coordinate
-         POINT = {warning("point from first coordinate only"); st_point(unclass(x)[1L, , drop = TRUE])},
+         POINT = {
+             if (st_is_empty(x)) {
+                 row <- NA_integer_
+             } else {
+	             warning("point from first coordinate only")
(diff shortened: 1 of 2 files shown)


edzer:
An alternative to `st_cast`? GEOS or GDAL may have much better facilities for that, never gave it a serious thought, we just started doing this in R, but not in a way I enjoy going back to.

krlmlr:
#114 seems slightly related.

edzer:
Thanks!

krlmlr:
The whole casting code can be made much more systematic if we base on _vctrs_ and implement `vec_cast()` . Is this something you'd be interested in in the long run?

edzer:
Maybe; I'll first need to understand `vctrs` before I can give an informed answer.

krlmlr:
https://vctrs.r-lib.org/articles/s3-vector.html, up to and including list-of classes, is a good start.

edzer:
Thanks!

PR #926, opened 2018-12-19
retain tbl_df for x in st_join

kendonB:


edzer:
I see the point of this, but would try a different approach:
* instead of pkg %in% .packages use requireNamespace(, quite = TRUE), like everywhere else
* check one command earlier on whether we have a tbl_df, and then use bind_cols instead of cbind(as.data.frame(...

kendonB:
Made those changes. Cheers

PR #927, opened 2018-12-19
add as_tibble

kendonB:


edzer:
Isn't `as_tibble` supposed to return a `tbl_df`, rather than an `sf` object? Wouldn't you expect it to de-sf an object? `as.data.frame.sf` now de-sf-s an `sf`,`data.frame` object.

kendonB:
`as_tibble` tibbleizes `data.table`s (as in, creates a `tbl_dt`), so I would also expect it to tibbleize an `sf` `data.frame` also.

kendonB:
@krlmlr, what should the behavior be?

krlmlr:
I'd expect it to de-sf indeed, just like `as.data.frame.tibble` un-tibbleizes. You can implement using something like

```r
new_tibble(x, nrow = nrow(x))
```

. This keeps the geometry column.

PR #938, opened 2019-01-06
Add a short description about nc dataset

yutannihilation:
`nc` dataset is so popular, yet I still feel it's difficult to find the definition of the columns. It's really helpful if there's just a link to the spdep's vignette.

In addition, ~~I deleted `bgMap` because it doesn't actually exist in sf (right?).~~ I moved `bgmap` to tests, as it seems not a dataset for users.


yutannihilation:
Oh, nice. Thanks for the notice, I'll wait for the vignette to be on the pkgdown website!

> In addition, I deleted `bgMap` because it doesn't actually exist in sf (right?).

Sorry, it seems I was wrong on this part... Will revert later (though I haven't seen `data("bgmap", package = "sf")` is used in actual).

https://travis-ci.org/r-spatial/sf/jobs/475884038#L2963-L2964

yutannihilation:
Instead of reverting, I moved `data/bgmap.rda` to `tests`, as it seems to be used only in `tests/plot.R`. If this change is too much for this PR, I'll create a different PR for it.

rsbivand:
The nc shapefile does not have many of the fields used in the spdep sids vignette:

```
> names(nc)
 [1] "AREA"      "PERIMETER" "CNTY_"     "CNTY_ID"   "NAME"      "FIPS"     
 [7] "FIPSNO"    "CRESS_ID"  "BIR74"     "SID74"     "NWBIR74"   "BIR79"    
[13] "SID79"     "NWBIR79"   "geometry" 
> names(nc.sids)
 [1] "CNTY_ID"   "AREA"      "PERIMETER" "CNTY_"     "NAME"      "FIPS"     
 [7] "FIPSNO"    "CRESS_ID"  "BIR74"     "SID74"     "NWBIR74"   "BIR79"    
[13] "SID79"     "NWBIR79"   "east"      "north"     "x"         "y"        
[19] "lon"       "lat"       "L_id"      "M_id"     
```
The median smoothing and model fitting IIRC need `L_id` and `M_id`.

edzer:
Somebody must have thrown them out. That person should have thrown out `AREA` too. 

Are they worth keeping?

rsbivand:
I'll cross-check, but to run the **spdep** `sids` vignette from the file(s) shipped with **sf**, some are needed. The neighbour objects from the articles are also needed and are in **spData**.

angela-li:
By the way, @yutannihilation, the SIDS data and a description of the columns can currently be found on the GeoDa Center data site [here](https://geodacenter.github.io/data-and-lab/sids/). (It used to be on a UIUC site, but Luc's center moved!)

yutannihilation:
Oh, thanks @angela-li! Should I use this URL in my PR?

> Somebody must have thrown them out. That person should have thrown out AREA too.

If possible, I think it's nice to include the script to produce the datasets in the package, so that anyone can reproduce the result. (e.g. https://github.com/tidyverse/ggplot2/blob/master/data-raw/economics.R)

rsbivand:
No, the original source has been offline for years, and the tabular data are hand-punched from the cited articles and book. Read the introduction to the vignette and you'll see. Anyway, I would never impose tidyverse on anyone, it is so 2016 and in ~20 years will not be understood (this vignette and the data set are about 20 years old now in this form).

yutannihilation:
I meant, if sf's `nc` is the modified version of spdep's `nc.sids`, and if the transformation was done with some R scripts, I thought it's a good idea to include the script (it's not related to tidyverse at all). But, I misunderstood the process, sorry.

rsbivand:
The github pkgdown version of sids.html is OK (even though pkgdown is really badly written, building *.o and *.so in tree, not elsewhere). **sf** should use the **spData** version (from **spdep**), really, including the omitted variables. If they are omitted, you can't fully replicate the Cressie et al. articles.

yutannihilation:
Thanks, but ... sorry, I'm confused. What changes do you suggest that I should make in this pull request?

I now roughly understand there are several problems to solve before describing `nc` correctly, but I'm afraid I can't handle them due to my lack of the knowledge of this domain; I just wanted to add a reference so that users can follow this kind of discussion relatively easier.

yutannihilation:
Thanks for clarification! I now see what I should do.

edzer:
The footnote in the vignette points to a dead link; @angela-li pointed me to https://geodacenter.github.io/data-and-lab/ which has a sids.zip, this might be identical to the one of the dead link (didn't check), and might be worth mentioning. Angela mentioned she is working on a package wrapping all the sources on this site.

rsbivand:
Thanks, will add this to the vignette footnote.

rsbivand:
Pushed bb9db8cee156c08063a763602f1ccccfd3e2a71e to add documentation for the **sf** and **spData** versions, and the two GeoDa versions.

edzer:
Looking good, thanks!

yutannihilation:
Thanks for all your help!

PR #952, opened 2019-01-17
harden R version check in configure for use in R devel

rnuske:
The R version check of configure fails in R-devel because `"${RBIN}" --version | head -1 | cut -f3 -d" "` returns "development" instead of a version string. The following tests expect numbers and fail consequently if confronted with "development".


edzer:
Looking good -- thanks!

Issue #953, opened 2019-01-19
Clarify difference between st_intersection() and st_intersects()


angela-li:
It's easy to get confused between `st_intersection` and `st_intersects` because their names are so similar (People often type "`st_intersects`" when "`st_intersection`" is what they intend). I propose adding a note in the function documentation for each regarding the other function. 

Additionally, we could also add a note in the vignettes clarifying the differences between the two.

angela-li:
This is a great addition, thanks all for helping out!

PR #955, opened 2019-01-19
Added notes in documentation and vignette 3 differentiating between s…

lambdamoses:
…t_intersects and st_intersection. Addressing issue #953

### DIFF
--- a/R/geom.R
+++ b/R/geom.R
@@ -246,7 +246,7 @@ st_relate	= function(x, y, pattern = NA_character_, sparse = !is.na(pattern)) {
 #'
 #' Sparse geometry binary predicate (\code{\link{sgbp}}) lists have the following attributes: \code{region.id} with the \code{row.names} of \code{x} (if any, else \code{1:n}), \code{ncol} with the number of features in \code{y}, and \code{predicate} with the name of the predicate used.
 #'
-#' @note For intersection on pairs of simple feature geometry sets, use
+#' @note For intersection on pairs of simple feature geometries, use
 #' the function \code{\link{st_intersection}} instead of \code{st_intersects}.
 #'
 #' @examples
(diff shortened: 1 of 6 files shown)


R/geom.R
```diff
@@ -245,6 +245,10 @@ st_relate	= function(x, y, pattern = NA_character_, sparse = !is.na(pattern)) {
 #' If \code{y} is missing, `st_predicate(x, x)` is effectively called, and a square matrix is returned with diagonal elements `st_predicate(x[i], x[i])`.
 #'
 #' Sparse geometry binary predicate (\code{\link{sgbp}}) lists have the following attributes: \code{region.id} with the \code{row.names} of \code{x} (if any, else \code{1:n}), \code{ncol} with the number of features in \code{y}, and \code{predicate} with the name of the predicate used.
+#'
+#' @note For intersection on pairs of simple feature geometry sets, use
```
edzer: I would say "geometries" rather than "geometry sets", since we don't do the intersection of the _set_s, but of the pairs of geometries.

R/geom.R
```diff
@@ -899,6 +903,8 @@ get_first_sfg = function(x) {
 #' @details A spatial index is built on argument \code{x}; see \url{http://r-spatial.org/r/2017/06/22/spatial-index.html}. The reference for the STR tree algorithm is: Leutenegger, Scott T., Mario A. Lopez, and Jeffrey Edgington. "STR: A simple and efficient algorithm for R-tree packing." Data Engineering, 1997. Proceedings. 13th international conference on. IEEE, 1997. For the pdf, search Google Scholar.
 #' @seealso \link{st_union} for the union of simple features collections; \link{intersect} and \link{setdiff} for the base R set operations.
 #' @export
+#' @note To find whether pairs of simple feature geometry sets intersect, use
```
edzer: same here.

R/geom.R
lambdamoses: OK, I'm changing it

edzer:
Great, thanks!

PR #962, opened 2019-01-23
Fixed build for macos in conda environment

vfonov:
Hello,
I had trouble building sf package for MacOS in conda environment, related to C++-11 and linking tests programs to gdal library  - it was caused by wrong flags in configure.ac. I fixed the problem and regenerated configure with autoreconf.

### DIFF
--- a/configure
+++ b/configure
@@ -623,7 +623,6 @@ ac_subst_vars='LTLIBOBJS
 LIBOBJS
 GEOS_CONFIG
 PROJ_LIBS
-PROJ_CPPFLAGS
 EGREP
 GREP
 CPP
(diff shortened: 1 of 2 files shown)


edzer:
I'm not merging this because this is a pretty massive set of changes to a messy code base I hate to touch, and that needs testing on all platforms available. I hope conda will at some stage simplify such that a smaller set of changes make sf install there on the fly. Thanks for your efforts!

vfonov:
Most of the changes are in autogenerated file configure (produced by autoconf/automake ). What's important is to use `${RBIN}" CMD config CXX` to determine which compiler to use and also to pass `$LDFLAGS` to the build commands for test binaries.

PR #971, opened 2019-02-11
New grob constructor for sfc objects

thomasp85:
This PR adds `st_as_grob` methods for `sfc` object thus removing the need to loop over the individual features and make separate grobs for each. This change also adds support for the upcoming changes in grid that allows multiple paths to be specified in a single grob. This change makes sf plotting with ggplot2 much faster, depending of course on the number of features in the data to plot. Plotting countries110 from rnaturalearth is around 3x faster, and completely removes grob creation as a bottleneck


edzer:
Cool! There are a couple of warnings in the travis check that seem easy to address. Did you also provide tests for the new code?

thomasp85:
I'll go through these and add tests as well

edzer:
This still seems to break on r-release.

thomasp85:
Yeah - I’m looking into it (it’s because of the parameters in pathGrob not matching in release despite them no being called in release)

thomasp85:
@edzer I’m getting a flurry of unrelated errors on travis now for the unit tests. Does this make sense to you?

edzer:
Amazing! CI has given me a lot of headaches lately; restarted both failing jobs - lets see.

edzer:
Clueless once more, and same error on travis release and appveyor! Maybe try switch of the cache in the .travis.yml of your PR?

thomasp85:
I’ll poke some more at it tomorrow

edzer:
Ready to go?

thomasp85:
I have yet to add unit tests to the new grab constructors as I don't really understand how your current setup works... If you can point me in the right direction I'll add some formal tests

edzer:
You can add a test in the testthat directory, somewhere where it fits or in a new file.

thomasp85:
But you already have some grid tests outside of testthat — shouldn't they be put there? and if so, how would you like it, because I don't really understand non-testthat tests all that well

edzer:
They are there so that you can see output if there are differences on CI or CRAN platforms, which is often more helpful then testthat output. Testthat is fine.

edzer:
Did you btw figure out _why_ https://github.com/r-spatial/sf/pull/971/commits/8065a0871bc479ffe5109c7f1bf69aabd6f9d716 was needed, or was it pure guesswork? I'm asking because the code was a result of decades of trial-and-error, and I'd like to understand its need. Cc: @rsbivand

edzer:
Timing note: I will submit sf to CRAN Today EOB. If we want to get this PR in, we need to sort this out before. I'm also OK with leaving it for now.

thomasp85:
I'd really like it in - let me focus on the tests, and I'll have it done today

thomasp85:
Regarding [8065a08](https://github.com/r-spatial/sf/commit/8065a0871bc479ffe5109c7f1bf69aabd6f9d716) - it was needed purely because you called `.onLoad` inside the tests. After loading the package environment gets locked meaning that `pathGrob` couldn't be modified when it was called as part of the tests... So I had to refactor to make gdal load/unload testable without calling `.onLoad`

edzer:
Thanks, that is a good reason.

thomasp85:
I've added tests for the grob construction. Assuming they pass (they do locally) you are free to merge at your discretion

edzer:
Great work, thanks!

PR #973, opened 2019-02-13
Optimised Ops.sfc for simple cases

thomasp85:
This PR adds a fast approach to `+`, `-`, `*`, `/` when the rhs is a simple numeric that should get applied to each element in the list...

for `+`/`-` the new methods kicks in when rhs is of length 1 or 2, for `*`/`/` it kicks in if rhs is of length one or a strictly diagonal matrix

Further this PR adds `st_normalize` that leverage the C++ code for even faster normalisation (avoids some copying by doing all the manipulation in one pass)


edzer:
Great work, thanks!

PR #981, opened 2019-02-20
Update example + docx in st_sample

Robinlovelace:
Heads-up @edzer this shows the vectorised form of st_sample (I was not aware of this!)


Robinlovelace:
Update: I think the second commit may be more important/useful for users. I only appreciated that this function is vectorised recently. The example now shows the most common use case (sampling n points in a polygon) early, and a comparison with a vectorised alternative.

PR #982, opened 2019-02-21
Fix issue with st_sample exact

Robinlovelace:
I suspect the second one will be faster and yield the same result.

```r
remotes::install_github("r-spatial/sf")
nc = sf::read_sf(system.file("shape/nc.shp", package="sf"))
system.time({p = sf::st_sample(x = nc, size = 1:nrow(nc))})
remotes::install_github("robinlovelace/sf", "update-st_sample_exact")
system.time({p = sf::st_sample(x = nc, size = 1:nrow(nc))})
```

### DIFF
--- a/R/sample.R
+++ b/R/sample.R
@@ -217,7 +217,3 @@ st_sample_exact = function(x, size, ..., type) {
 	}
 	random_pt
 }
-
-# library(sf)
-# nc = read_sf(system.file("shape/nc.shp", package="sf"))
-# system.time({p = st_sample(x = nc, size = 1:nrow(nc))})
--- a/R/sample.R
+++ b/R/sample.R
@@ -209,7 +209,7 @@ st_sample_exact = function(x, size, ..., type) {
 	random_pt = st_sample(x = x, size = size, ..., type = type, exact = FALSE)
 	while (length(random_pt) < size) {
 		diff = size - length(random_pt)
-		random_pt_new = st_sample(x, size, ..., type, exact = FALSE)
+		random_pt_new = st_sample(x, size = diff, ..., type, exact = FALSE)
 		random_pt = c(random_pt, random_pt_new)
 	}
 	if(length(random_pt) > size) {
@@ -217,3 +217,7 @@ st_sample_exact = function(x, size, ..., type) {
 	}
 	random_pt
 }
+
+# library(sf)
+# nc = read_sf(system.file("shape/nc.shp", package="sf"))
+# system.time({p = st_sample(x = nc, size = 1:nrow(nc))})


R/sample.R
```diff
@@ -209,11 +209,15 @@ st_sample_exact = function(x, size, ..., type) {
 	random_pt = st_sample(x = x, size = size, ..., type = type, exact = FALSE)
 	while (length(random_pt) < size) {
 		diff = size - length(random_pt)
-		random_pt_new = st_sample(x, size, ..., type, exact = FALSE)
+		random_pt_new = st_sample(x, size = diff, ..., type, exact = FALSE)
```
Robinlovelace: I missed that somehow 1st time @edzer. Apologies. The result will be the same, just slightly slower I guess.

Robinlovelace:
Thanks for quick merge @edzer . Proof that this speeds things up in reprex below:

``` r

remotes::install_github("r-spatial/sf", quiet = TRUE)
nc = sf::read_sf(system.file("shape/nc.shp", package="sf"))
system.time({p = sf::st_sample(x = nc, size = 1:nrow(nc))})
#>    user  system elapsed 
#>   2.773   0.024   2.798
remotes::install_github("robinlovelace/sf", "update-st_sample_exact", quiet = TRUE)
system.time({p = sf::st_sample(x = nc, size = 1:nrow(nc))})
#>    user  system elapsed 
#>   1.856   0.028   1.884
```

<sup>Created on 2019-02-21 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

Robinlovelace:
(That ran before you merged this PR.)

edzer:
I've put `exact = FALSE` back as default for now; if you run a reverse dependency check with pkg cartography you'll find breaks caused by `exact = TRUE` being the new default, and the implementation sometimes returning a GEOMETRYCOLLECTION, which is what cartography doesn't expect (and neither do I).

Robinlovelace:
Thanks for the update. Will take a look. Any hints on how to reproduce that finding appreciated.

edzer:
Run R CMD check on cartography, and find out why `st_sample` returns GEOMETRYCOLLECTIONS.

Robinlovelace:
@edzer I'm trying to reproduce the error. I've tried running the following lines on both versions of **sf** and cannot find a difference:

```r
remotes::install_github("riatelab/cartography", dependencies = TRUE)
download.file(url = "https://cran.r-project.org/src/contrib/cartography_2.2.0.tar.gz", "cartography.tar.gz")
system("R CMD check 'cartography.tar.gz'")
untar("cartography.tar.gz")
devtools::check("cartography")
```

The only place I can see `st_sample()` used in the package is here, but I don't think that `makelines()`, called from `getPencilLayer()` was responsible. Will continue tomorrow.

Robinlovelace:
Update, I think I've found the issue:

```r
mtq <- st_read(system.file("gpkg/mtq.gpkg", package="cartography"))
Reading layer `mtq' from data source `/home/robin/cartography.Rcheck/cartography/gpkg/mtq.gpkg' using driver `GPKG'
Simple feature collection with 34 features and 7 fields
geometry type:  MULTIPOLYGON
dimension:      XY
bbox:           xmin: 690574 ymin: 1592536 xmax: 735940.2 ymax: 1645660
epsg (SRID):    32620
proj4string:    +proj=utm +zone=20 +datum=WGS84 +units=m +no_defs
> mtq_pencil <- getPencilLayer(x = mtq)
Error in st_coordinates.sfc(pt) : 
  not implemented for objects of class sfc_GEOMETRY
```

Robinlovelace:
And yes, it passes with the latest version of sf:

```
* checking running R code from vignettes ...
   ‘cartography.Rmd’ using ‘UTF-8’ ... OK
   ‘cheatsheet.Rmd’ using ‘UTF-8’ ... OK
 NONE
* checking re-building of vignette outputs ... OK
* checking PDF version of manual ... OK
* DONE

Status: OK
```

PR #984, opened 2019-02-22
Sample man - type arg

robitalec:
Just a quick PR for the sampling type options in the man. 
It said "only random is implemented right now" despite examples and underlying code for types hexagonal, regular.


robitalec:
Also fixes the exact argument since it was still described in the manual as TRUE by default:

Lines 17-19
```r
#' @param exact logical; should the length of output be exactly
#' the same as specified by \code{size}? \code{TRUE} by default. Only applies to polygons, and
#' when \code{type = "random"}.
```

edzer:
Thanks!

Issue #900, opened 2018-11-19
error: ‘isLongjumpSentinel’ is not a member of ‘Rcpp::internal’


samfux84:
Hi,

I am trying to install the sf package, using GDAL 2.2.3, PROJ 4.9.3 and GEOS 3.6.2. When installing the package, I get some errors regarding the RCpp package:

```
In file included from ../inst/include/sf.h:7:0,
                 from RcppExports.cpp:4:
../inst/include/sf_RcppExports.h: In function ‘Rcpp::List sf::CPL_read_wkb(Rcpp::List, bool, bool)’:
../inst/include/sf_RcppExports.h:41:13: error: ‘isLongjumpSentinel’ is not a member of ‘Rcpp::internal’
         if (Rcpp::internal::isLongjumpSentinel(rcpp_result_gen))
             ^
../inst/include/sf_RcppExports.h:42:19: error: ‘LongjumpException’ is not a member of ‘Rcpp’
             throw Rcpp::LongjumpException(rcpp_result_gen);
                   ^
../inst/include/sf_RcppExports.h: In function ‘Rcpp::List sf::CPL_write_wkb(Rcpp::List, bool)’:
../inst/include/sf_RcppExports.h:62:13: error: ‘isLongjumpSentinel’ is not a member of ‘Rcpp::internal’
         if (Rcpp::internal::isLongjumpSentinel(rcpp_result_gen))
             ^
../inst/include/sf_RcppExports.h:63:19: error: ‘LongjumpException’ is not a member of ‘Rcpp’
             throw Rcpp::LongjumpException(rcpp_result_gen);
                   ^
RcppExports.cpp: In function ‘SEXPREC* _sf_CPL_read_wkb(SEXP, SEXP, SEXP)’:
RcppExports.cpp:922:32: error: ‘isLongjumpSentinel’ is not a member of ‘Rcpp::internal’
     bool rcpp_isLongjump_gen = Rcpp::internal::isLongjumpSentinel(rcpp_result_gen);
                                ^
RcppExports.cpp:924:9: error: ‘resumeJump’ is not a member of ‘Rcpp::internal’
         Rcpp::internal::resumeJump(rcpp_result_gen);
         ^
RcppExports.cpp: In function ‘SEXPREC* _sf_CPL_write_wkb(SEXP, SEXP)’:
RcppExports.cpp:957:32: error: ‘isLongjumpSentinel’ is not a member of ‘Rcpp::internal’
[...]

PR #986, opened 2019-02-24
Specify min Rcpp version

nikolai-b:
Ref #900


edzer:
Makes sense - thanks!

PR #998, opened 2019-03-05
Tweak Makevars.win for possible future toolchain

jeroen:
Small tweak that will make it easier to update to the next toolchain.

### DIFF
--- a/src/Makevars.win
+++ b/src/Makevars.win
@@ -10,6 +10,7 @@ PKG_CPPFLAGS =\
 
 PKG_LIBS = \
 	-L$(RWINLIB)/$(TARGET) \
+	-L$(RWINLIB)/lib$(R_ARCH) \
 	-lgdal -lsqlite3 -lspatialite -lproj -lgeos_c -lgeos  \
 	-ljson-c -lnetcdf -lmariadbclient -lpq -lintl -lwebp -lcurl -lssh2 -lssl -lcrypto \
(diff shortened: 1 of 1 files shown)


jeroen:
@rsbivand could you make this same change in rgdal? I'm not sure what the appropriate way is to propose a patch in rforge.

rsbivand:
Thanks, change made for **rgdal**, will commit shortly. Does **rgeos** need attention: https://r-forge.r-project.org/scm/viewvc.php/pkg/src/Makevars.win?view=markup&root=rgeos ?

jeroen:
@rsbivand yes it would be nice if you could also update __rgeos__ to the gdal-2.2.3 build so that I no longer have to look after the gdal-2.2.0 build.

The easiest way to do this is copy 2 files [`src/Makevars.win`](https://github.com/r-spatial/sf/blob/6ca86b88be7615005358c024ffb2607d6569f718/src/Makevars.win) and [`tools/winlibs.R`](https://github.com/r-spatial/sf/blob/6ca86b88be7615005358c024ffb2607d6569f718/tools/winlibs.R) from rgdal into rgeos. Then perhaps you can remove the `-I` and `-L` and `-l` flags from `Makevars.win` which are unneeded for rgeos (though it's harmless having them in there, the linker will just ignore them).

rsbivand:
Done for rgeos, committed to R-Forge.

jeroen:
This PR can be merged ✅

edzer:
Thanks a lot!

PR #1003, opened 2019-03-13
Check argument type in st_simplify

dbaston:
This type check avoids a malformed call of `st_simplify(nc, 1e-4)` being interpreted as `st_simplify(nc, dTolerance=0, preserveTopology=TRUE)`.


edzer:
Thanks

PR #1005, opened 2019-03-14
Avoid union of identical geometries in summarize()

dbaston:
This PR updates `CPL_geos_union` to avoid performing an expensive union of identical geometries, optimizing a usage pattern in which we:

a) perform a non-spatial one-to-many join between `sf` object and a data frame
b) group by an ID that originated in the `sf` object
c) call summarize with the default `do_union = TRUE`

Short code example:

```
# take spatial data
nc <- st_read(system.file("shape/nc.shp", package='sf'))

# combine with tabular data; 1-to-many
data <- data.frame(FIPS=rep(nc$FIPS, each=3),
                   year=rep(c(1990, 2000, 2010), times=nrow(nc)),
                   obs=runif(n=3*nrow(nc), min=10, max=20))

a <- nc %>%
     inner_join(data) %>%
     group_by(FIPS) %>%
     summarize(max_obs=max(obs))
```

In this case the `summarize()` call takes, on average:

| |  `do_union=TRUE` |  `do_union=FALSE` |
| -- | -- | --- |
this branch | 80 ms | 78 ms |
master branch | 148 ms | 76 ms |

Of course, the performance gain would be much larger as the number and complexity of duplicated geometries increases.

### DIFF
--- a/src/geos.cpp
+++ b/src/geos.cpp
@@ -560,8 +560,22 @@ Rcpp::List CPL_geos_union(Rcpp::List sfc, bool by_feature = false) {
 			gmv_out[i] = geos_ptr(GEOSUnaryUnion_r(hGEOSCtxt, gmv[i].get()), hGEOSCtxt);
 		}
 	} else {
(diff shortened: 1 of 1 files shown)


dbaston:
@edzer any thoughts on this?

edzer:
Looks good, thanks!

PR #1041, opened 2019-05-04
Use new_tibble

etiennebr:
* Silence a quoting warning (03223f3)
* Create tibbles using `new_tibble`
* Test returned class for database
* Add logic to respect `stringsAsFactors` defaults (for tibbles it should be FALSE, for data.frames it should use `default.stringsAsFactors()`, I believe)

As always, happy to make modifications if needed.


edzer:
Thanks! I didn't know about `new_tibble` - will check it out.

PR #1043, opened 2019-05-09
Issue1015 - Z and M range attributes

SymbolixAU:
`z_range` and `m_range` attributes added to `sfc` objects when the Z and M dimensions are present.


edzer:
Sorry this took so long!

dcooley:
Thanks - this feature is already extremely useful.

PR #1080, opened 2019-06-15
Read wkt containing new lines

etiennebr:
I ran into the issue than long `WKT` couldn't be parsed in `sf`. Apparently line breaks are rejected by OGR.

I wonder if this line should be in `CPL_sfc_from_wkt` or if it's fine to keep it `st_as_sfc.character` since it's the only function calling it?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,14 @@
 # version 0.7-5
 
+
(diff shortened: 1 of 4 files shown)


edzer:
Thanks!

PR #1083, opened 2019-06-17
Quote numbers in graticule labels

paleolimbot:
This PR quotes the numbers generated in `st_graticule()`, so that these values can be safely passed to `parse()` in any locale (as reported in tidyverse/ggplot2#3365). This doesn't cause any changes in the ggplot2 doppleganger tests (tested using the re-enabled geom_sf() tests in tidyverse/ggplot2#3361), and thus I think it's unlikely that this will cause problems elsewhere (except in ggplot2's own tests, where we'll have to find a different way to test manual breaks in the graticule).


edzer:
Thanks Dewey!

PR #1086, opened 2019-06-18
Add pkgdown website to description

Bisaloo:
- Users can now directly visit the pkgdown website from the CRAN landing page for sf
- Other pkgdown websites that reference sf functions will now link to the https://r-spatial.github.io/sf/ instead of https://rdrr.io/


edzer:
Thanks!

Issue #1087, opened 2019-06-18
Error in CPL_read_ogr


NoushinN:
Any ideas why this error happens?
I am using the sf library to read a .shp file..appreciate any help!

`Error in CPL_read_ogr(dsn, layer, query, as.character(options), quiet,  : 
  Open failed.`

edzer:
Please provide a reproducible example.

NoushinN:
thank you, did you want more info? 
![image](https://user-images.githubusercontent.com/37668936/59780573-13a33900-926f-11e9-97ea-44eaf6c08341.png)

edzer:
Maybe load library `sf` first, so that `%>%` also works?

adrfantini:
The error you see in those screenshots (sigh) is related to the fact that the function `%>%` cannot be found. Load `sf`, `dplyr` or `magrittr` to get it.

Also, when providing a reproducible example you should also provide your input data. Otherwise it is hardly _reproducible_, is it?

NoushinN:
thank you both, it doesn't seem to be a matter of loading libraries as they are loaded, I include them every time with the two columns (e.g. here::here()) because there seems to be conflicts between packages (e.g. here with lubridate)
here's a repress of the larger workflow:
![image](https://user-images.githubusercontent.com/37668936/59805864-58e35d00-92a7-11e9-9261-4363a076194a.png)

rsbivand:
Please provide a real example. Also try not using `here::here()`, I guess that the fact that a shapefile is actually at least three files is confusing for **here**. I installed **here**, and tried to set it to:

```
library(sf)
here::set_here(system.file("shapes", package="spData")[1])
file.exists(file.path(system.file("shapes", package="spData")[1], ".here")) # TRUE
col <- st_read(here::here("columbus.shp")) # failed
# looks in /home/rsb
col <- read_sf(here::here("columbus.shp")) #failed
# looks in /home/rsb
col <- read_sf(system.file("shapes/columbus.shp", package="spData")[1]) # succeeded
col <- st_read(system.file("shapes/columbus.shp", package="spData")[1]) # succeeded
wrld <- read_sf(here::here("world.gpkg")) # failed
# looks in /home/rsb
wrld <- st_read(system.file("shapes/world.gpkg", package="spData")[1]) # succeeds
```
So, don't use **here**, not fit for purpose. I have removed **here**, a classic case of breaking the "as simple as possible, but not more so" rule.

etiennebr:
[...]

Issue #873, opened 2018-10-20
Can not read shapefile with absolute filepath


waholulu:
I can run the sample code but when I want to read the data from the absolute path, it fails. So funny.

```

file.exists("C://Users//Administrator//Documents//R//win-library//3.5//sf//shape//nc.shp")
[1] TRUE

nc <- st_read(system.file("C://Users//Administrator//Documents//R//win-library//3.5//sf//shape//nc.shp", package="sf"))

Cannot open data source 
Error in CPL_read_ogr(dsn, layer, as.character(options), quiet, type,  : 
  Open failed.

nc <- st_read(system.file("/shape//nc.shp", package="sf"))**

Reading layer `nc' from data source `C:\Users\Administrator\Documents\R\win-library\3.5\sf\shape\nc.shp' using driver `ESRI Shapefile'
Simple feature collection with 100 features and 14 fields
geometry type:  MULTIPOLYGON
dimension:      XY
bbox:           xmin: -84.32385 ymin: 33.88199 xmax: -75.45698 ymax: 36.58965
epsg (SRID):    4267
proj4string:    +proj=longlat +datum=NAD27 +no_defs
```

yutannihilation:
`system.file()` is a function that converts a path relative to the system directory to the absolute path. So, If you want to provide an absolute path, you don't need to use `system.file()`; you can just do this:

``` r
nc <- st_read("C:/Users/Administrator/Documents/R/win-library/3.5/sf/shape/nc.shp")
```

Issue #838, opened 2018-09-08
Error in Reading Data into R Notebook


cubiloju:
When I try the following line of code in an R script file on my computer, it works fine:

```{r}
mvc <- st_read('ga_mvc.gpkg')
```
However, when I try to do the same in my notebook, I get the following error message:
```{r}
Cannot open data source ga_mvc.gpkg
 
Error in CPL_read_ogr(dsn, layer, as.character(options), quiet, type, : Open failed.
4. stop(structure(list(message = "Open failed.\n", call = CPL_read_ogr(dsn, 
layer, as.character(options), quiet, type, promote_to_multi, int64_as_string), 
cppstack = structure(list(file = "", line = -1L, stack = "C++ stack not available 
on this system"), class = "Rcpp_stack_trace")), class = c("Rcpp::exception", ...

3. CPL_read_ogr(dsn, layer, as.character(options), quiet, type, promote_to_multi, int64_as_string)

2. st_read.default("ga_mvc.gpkg")

1. st_read("ga_mvc.gpkg")
```

Any clues on how to fix this would be useful. I actually can get the code to run on my laptop, but my desktop simply refuses to acknowledge that the data file exists. I've checked the working directory and spelling and I've tried reinstalling calling the sf package as well.

Nowosad:
Is the notebook in the same folder as your file?

cubiloju:
That was the problem. I checked everything except that it seems. Thank you!

PR #1095, opened 2019-06-25
improve error messages from st_read

### LINKED ISSUE #967: Add check if file.exists() for st_read( )
Would it be possible to do an early check/stop for if a file doesn't exist for st_read? The error message:

> Error in CPL_read_ogr(dsn, layer, query, as.character(options), quiet, : Open failed.

doesn't tell you if you've done something silly like given the wrong directory path to a file (as shown in #608). Adding an early check with `file.exists()` for `st_read` could provide a more informative error message like "File does not exist. Perhaps you have the incorrect file path?"

I'm speaking as someone who spent far too long trying to figure out what was wrong with my installation of `sf`, when it just turned out I'd made a typo in the file path 😆

etiennebr:
Did a survey of a few issues, it should help resolve #967, #1087, #873, #838

explictly identify:
- Empty string
- Missing file
- Corrupt files
- Missing connexion information

### DIFF
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -101,8 +101,8 @@ CPL_get_layers <- function(datasource, options, do_count = FALSE) {
     .Call('_sf_CPL_get_layers', PACKAGE = 'sf', datasource, options, do_count)
 }
 
-CPL_read_ogr <- function(datasource, layer, query, options, quiet, toTypeUser, fid_column_name, promote_to_multi = TRUE, int64_as_string = FALSE) {
-    .Call('_sf_CPL_read_ogr', PACKAGE = 'sf', datasource, layer, query, options, quiet, toTypeUser, fid_column_name, promote_to_multi, int64_as_string)
+CPL_read_ogr <- function(datasource, layer, query, options, quiet, toTypeUser, fid_column_name, promote_to_multi = TRUE, int64_as_string = FALSE, dsn_exists = TRUE, dsn_isdb = FALSE) {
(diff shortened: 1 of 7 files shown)


etiennebr:
Here's a version without the call. (I personally prefer that, but it affects all error messages.)

``` r
library(sf) 
#> Linking to GEOS 3.7.0, GDAL 2.4.0, PROJ 5.2.0

# missing file
st_read(tempfile(fileext = ".csv"))
#> Error: Cannot open "/tmp/Rtmp6FWqdN/file67e74dda6a79.csv"; The file doesn't seem to exist.

# wrong connection
st_read("PG:host=wrong")
#> Warning in CPL_read_ogr(dsn, layer, query, as.character(options), quiet, : GDAL Error 1: PQconnectdb failed.
#> could not translate host name "wrong" to address: Name or service not known
#> Error: Cannot open "PG:host=wrong"; Check connection parameters.

# corrupt file
corrupt <- tempfile(fileext = ".kml")
cat("empty", file = corrupt)
st_read(corrupt)
#> Warning in CPL_read_ogr(dsn, layer, query, as.character(options),
#> quiet, : GDAL Error 4: ERROR parsing kml /tmp/Rtmp6FWqdN/
#> file67e766aea47e.kml :syntax error on line 1 at offset 0
#> Error: Cannot open "/tmp/Rtmp6FWqdN/file67e766aea47e.kml"; The source could be corrupt or not supported. See `st_drivers()` for a list of supported formats.

# empty string, error comes from user translation of examples
st_read(system.file("file.system/is/for/examples"))
#> Error: `dsn` must point to a source, not an empty string.
```

edzer:
Looking good from my side! @angela-li any comments?

etiennebr:
After a few weeks I'm still comfortable with the error messages, so I think we could merge. I'm sure they'll evolve over time, but it looks like an improvement over what we have in place.

PR #1098, opened 2019-06-27
preserve column names in data.frames

etiennebr:
I found an hacky way, but it seems to work.
``` r
library(sf)
#> Linking to GEOS 3.7.0, GDAL 2.4.0, PROJ 5.2.0

# works
st_as_sf(
    data.frame(
        a = st_sfc(st_point(c(0, 0))),
        b = st_sfc(st_point(c(1, 0)))
    ), 
    sf_column_name = "b"
)
#> Simple feature collection with 1 feature and 0 fields
#> Active geometry column: b
#> geometry type:  POINT
#> dimension:      XY
#> bbox:           xmin: 1 ymin: 0 xmax: 1 ymax: 0
#> epsg (SRID):    NA
#> proj4string:    NA
#>             a           b
#> 1 POINT (0 0) POINT (1 0)
```
But unnamed columns are now ugly (default data.frame behavior)
``` r
st_as_sf(data.frame(st_sfc(st_point(c(0, 0)))))
#> Simple feature collection with 1 feature and 0 fields
#> geometry type:  POINT
#> dimension:      XY
#> bbox:           xmin: 0 ymin: 0 xmax: 0 ymax: 0
#> epsg (SRID):    NA
#> proj4string:    NA
#>   st_sfc.st_point.c.0..0...
#> 1               POINT (0 0)
```

The other option would be to build the data.frame from scratch with something like
``` r
as.data.frame.sfc = function(x, ...) {
	structure(list(x), row.names = seq_along(x), class = "data.frame")
}
```
But it might be more brittle since it doesn't use default `data.frame` constructor?

### DIFF
--- a/R/sfc.R
+++ b/R/sfc.R
@@ -222,8 +222,8 @@ summary.sfc = function(object, ..., maxsum = 7L, maxp4s = 10L) {
 #' @export
 as.data.frame.sfc = function(x, ...) {
 	ret = data.frame(row.names = seq_along(x))
-	ret$geometry = x
-	ret
+	ret[["geometry"]] = x
+	setNames(ret, NULL)
 }
 
 
--- a/tests/testthat/test_postgis_RPostgreSQL.R
+++ b/tests/testthat/test_postgis_RPostgreSQL.R
@@ -44,9 +44,9 @@ test_that("can write to db", {
 
 test_that("can handle multiple geom columns", {
 	skip_if_not(can_con(pg), "could not connect to postgis database")
-	multi <- cbind(pts[["geometry"]], st_transform(pts, 4326))
+	multi <- cbind(geometry_2 = pts[["geometry"]], st_transform(pts, 4326))
 	expect_silent(st_write(multi, pg, "meuse_multi", overwrite = TRUE))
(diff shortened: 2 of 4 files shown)


edzer:
I like it because it solves our problem, but I don't like it as `as.data.frame.sfc` now no longer returns a valid data frame, which according to `?data.frame` should have non-empty column names. Maybe we can in addition stop exporting it? Or would that cause havoc downstream?

etiennebr:
I agree that it doesn't make sense to call `as.data.frame` on an `sfc` object directly (it would make sense on an `sf` object, though). However if we don't export `as.data.frame.sfc`, then `data.frame` can't find it. 
``` r
> x <- data.frame(
+ 	a = st_sfc(st_point(c(0, 0)))
+ )
Error in as.data.frame.default(x[[i]], optional = TRUE, stringsAsFactors = stringsAsFactors) : 
  cannot coerce class ‘c("sfc_POINT", "sfc")’ to a data.frame
```

etiennebr:
Ok, after a few experiments, it is annoyingly more complex than I thought.

PR #1103, opened 2019-07-08
Parse GEOS dev version numbers

dbaston:
Avoids the following error message during configure:
`checking GEOS version >= 3.4.0... ./configure: line 4170: test: 380dev: integer expression expected`


edzer:
Thanks, Dan!

Issue #990, opened 2019-02-28
st_nearest binary predicate?


kendonB:
Would be a good addition for use in `st_join` and looks like it can rely on `st_nearest_feature`.

kendonB:
Would enable this https://github.com/r-spatial/sf/issues/200

See also https://github.com/r-spatial/sf/issues/200#issuecomment-354680274

tim-salabim:
You can already use `st_nearest_feature` in `st_join`. 
```r
library(sf)
library(mapview)

st_join(breweries, franconia, join = st_nearest_feature)

Simple feature collection with 225 features and 14 fields
geometry type:  POINT
dimension:      XY
bbox:           xmin: 9.462785 ymin: 48.90074 xmax: 11.93539 ymax: 50.44162
epsg (SRID):    4326
proj4string:    +proj=longlat +datum=WGS84 +no_defs
First 10 features:
                                                 brewery                    address zipcode       village  state founded number.of.types number.seasonal.beers NUTS_ID
1                                     Brauerei Rittmayer     Aischer Hauptstrasse 5   91325     Adelsdorf Bayern    1422               2                     1   DE257
2                                       Brauhaus Leikeim            Gewerbegebiet 4   96264 Altenkunstadt Bayern    1887              11                     1   DE24C
3  Ammerndorfer Bier Dorn-Braeu H. Murmann GmbH & Co. KG             Marktplatz 1-2   90614    Ammerndorf Bayern    1730              10                     0   DE258
4                          Wittelsbacher Turm Braeu GmbH       Wittelsbacher Turm 1   97688 Bad Kissingen Bayern      NA               6                     0   DE265
5                                    Arnsteiner Brauerei    Schweinfurter Strasse 9   97450      Arnstein Bayern    1885               5                     0   DE26A
6                                     Aufsesser Brauerei                 Im Tal 70b   91347       Aufsess Bayern    1886               7                     0   DE246
7                                       Brauhaus Doebler                Kornmarkt 6   91438 Bad Windsheim Bayern    1867               8                     0   DE25A
8                                     Ambraeusianum GmbH      Dominikanerstrasse 10   96049       Bamberg Bayern    2004               3                     3   DE241
9                         Brauerei Faessla GmbH & Co. KG Obere Koenigsstrasse 19-21   96052       Bamberg Bayern    1694               5                     0   DE241
[...]

PR #1113, opened 2019-07-24
Add a couple of predicates to st_join doc, format as list

dbaston:
Motivated by https://github.com/r-spatial/sf/issues/990


edzer:
Good idea, thanks!

PR #1120, opened 2019-08-01
Optimize reading sf geometries from WKB

dbaston:
- Make wkb_read function inline
- Use Rcpp::no_init

Reduces overall time of CPL_read_wkb by about 50%, most of which is due
to the inlining. Another roughly 15% is possible by avoiding checking the
buffer length before every single read, and by making specialized
versions of the read_numeric_matrix for the supported cases of 2, 3, and
4 dimensions. These changes are a bit more invasive and are not included
here.

### DIFF
--- a/src/wkb.cpp
+++ b/src/wkb.cpp
@@ -26,20 +26,28 @@ typedef struct {
 	size_t size;
 } wkb_buf;
 
-void wkb_read(wkb_buf *wkb, void *dst, size_t n);
-
 Rcpp::List read_data(wkb_buf *wkb, bool EWKB, bool spatialite, int endian, 
 	bool addclass, int *type, uint32_t *srid);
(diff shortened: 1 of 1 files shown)


src/wkb.cpp
```diff
@@ -42,6 +40,16 @@ void wkb_read(wkb_buf *wkb, void *dst, size_t n) {
 	wkb->size -= n;
 }
 
+template<typename T>
+inline T wkb_read(wkb_buf *wkb) {
```
dbaston: This new signature of `wkb_read` doesn't provide any performance benefit on its own; I just found it simpler to use.

edzer:
Thanks, makes a lot of sense! Would you mind if I add you as one of the contributors to sf?

dbaston:
Happy to be added, thanks!

PR #1128, opened 2019-08-16
Fix typo in st_sample documentation

Pakillo:
Three types of spatial sampling: random, hexagonal, and regular. 'regular' was repeated twice - changed to 'random'


edzer:
Thanks!

PR #1145, opened 2019-09-11
Fix compatible with tidyr 1.0.0

yutannihilation:
**DO NOT MERGE THIS**

Fixes https://github.com/r-spatial/sf/issues/1068

This PR will

* define `vec_proxy.sfc()` to make `unnest()` and `separate_rows()` work
* remove tests and examples that no longer work
* comment out `nest()` related codes until https://github.com/tidyverse/tidyr/pull/729 is merged

PR #1150, opened 2019-09-14
Remove Remotes field from DESCRIPTION

yutannihilation:
I added this in #1145 to test with tidyr 1.0.0. Now that it's released on CRAN, these lines are no more needed (sorry that I didn't do this soon...).


edzer:
Thanks!

PR #1151, opened 2019-09-15
First attempt to adapt pivot_longer to sf

lepennec:
As pivot_longer is not yet a s3_methods, the new function is only available under the name pivot_longer.sf...


R/tidyverse.R
```diff
@@ -249,6 +249,34 @@ spread.sf <- function(data, key, value, fill = NA, convert = FALSE, drop = TRUE,
 		drop = drop, sep = sep), sf_column_name = attr(data, "sf_column"))
 }
 
+#' @name tidyverse
+#' @export
+#' @examples
+#' library(tidyr)
+#' nc %>% select(SID74, SID79) %>% gather("VAR", "SID", -geometry) %>% summary()
+pivot_longer.sf <- function(data, cols, names_to = "name", names_prefix = NULL,
```
edzer: I suggest to replace all the parameters that you want to pass on untouched as ... , unless you want to change their default. This would spare you (and me) the work of documenting them, and keeping track of changes in upcoming tidyr versions.

R/tidyverse.R
```diff
@@ -405,6 +433,7 @@ register_all_s3_methods = function() {
 	register_s3_method("dplyr", "ungroup", "sf")
 	register_s3_method("tidyr", "gather", "sf")
 	register_s3_method("tidyr", "spread", "sf")
+	register_s3_method("tidyr", "pivot_longer", "sf")
```
edzer: if `pivot_longer` is not an s3 method, this shouldn't work.

lepennec:
I agree with your two comments. I will wait until the situation around
pivot_longer is clearer. In between, I will prepare everything for all the
pivot_* functions.

edzer:
please re-PR when there is some progress in this area.

Issue #1160, opened 2019-09-30
clean_columns drops list columns silently


rsbivand:
`clean_columns()` says nothing but drops not-permitted columns silently. `rgdal::writeOGR()` throws an error; should it at least warn?

```
df <- data.frame(a=letters[1:3], b=1:3)
df$c <- list(d=1, e="1", f=TRUE)
library(sf)
pt1 <- st_point(c(1,3))
pt2 <- pt1 + 1
pt3 <- pt2 + 1
pt_sfc <- st_as_sfc(list(pt1, pt2, pt3))
st_geometry(df) <- pt_sfc
df
sf:::clean_columns(as.data.frame(df), factorsAsCharacter=TRUE)
tf <- tempfile(fileext=".gpkg")
st_write(df, dsn=tf)
(st_read(dsn=tf, quiet=TRUE))
tf1 <- tempfile(fileext=".gpkg")
rgdal::writeOGR(df, dsn=tf, driver="GPKG")
```

edzer:
Sounds like a good idea!

rsbivand:
PR https://github.com/r-spatial/sf/pull/1162

PR #1162, opened 2019-10-01
added warning in clean_columns for dropped columns

rsbivand:
Re: https://github.com/r-spatial/sf/issues/1160. Not tested against arbitrary column classes, just `list`


edzer:
Thanks!

Issue #1172, opened 2019-10-10
Merging layers


dicorynia:
I used list-columns and `map` + `unnest` to merge multiple shapefiles into one `sf` object. But it recently doesn't work any longer (it worked in March 2019). I'm not sure if it's related to `sf` or `tidyr`.

Reprex : 

1. get 3 shapefiles :

```r
    library(tidyverse)
    library(sf)
    library(fs)
    library(httr)
    
    # some shapefiles from https://fr.actualitix.com/blog/shapefiles-des-departements-de-france.html
    url <-  c("https://fr.actualitix.com/blog/actgeoshap/01-Ain.zip",
              "https://fr.actualitix.com/blog/actgeoshap/73-savoie.zip",
              "https://fr.actualitix.com/blog/actgeoshap/74-haute-savoie.zip")
    
    dep <- str_extract(url, "\\d{2}.*$")
    
    list(url, dep) %>% 
      pwalk(~ GET(.x, write_disk(.y)))
    
    walk(dep, unzip, junkpaths = TRUE, exdir = "shp")
```
2. merge

```r
    res <- dir_ls("shp", glob = "*.shp") %>% 
      tibble(fname = .) %>%
      mutate(data = map(fname, read_sf)) %>%
      unnest(data) %>%
      st_as_sf()
    # Error: No common type for `..1$data$geometry` <sfc_POLYGON> and `..2$data$geometry` <sfc_MULTIPOLYGON>.
    # Call `rlang::last_error()` to see a backtrace  
 
    rlang::last_error()
    # <error>
    #   message: No common type for `..1$data$geometry` <sfc_    # Error in rbind.data.frame(...) : 
    #   numbers of columns of arguments do not match> and `..2$data$geometry` <sfc_MULTIPOLYGON>.
    # class:   `vctrs_error_incompatible_type`
    # backtrace:
    #   1. fs::dir_ls("shp", glob = "*.shp")
    # 16. tibble::tibble(fname = .)
    # 17. dplyr::mutate(., data = map(fname, read_sf))
    # 18. tidyr::unnest(., data)
    # 19. vctrs::stop_incompatible_type(x, y, x_arg = x_arg, y_arg = y_arg)
    # 1. vctrs:::stop_incompatible(...)
    # 16. vctrs:::stop_vctrs(...)
```

I don't know the inner workings of `vctrs`... Should `sf` "advertise" in some way that POLYGON and MULTIPOLYGON are compatible (are they ?) ?

[...]

PR #1196, opened 2019-11-19
Finish implementing vctrs methods for `sfc` vectors

lionel-:
* Restore `n_empty`, `crs`, and `precision` attributes.

* Add common type implementation for `sfc` objects. This finds the common class based on heuristics found in `c.sfc()`.

* Added a new `vctrs` help topic, so that argument names can be documented separately from the tidyverse ones.


lionel-:
Should we move the vctrs methods into their own file `vctrs.R` or perhaps `tidyverse-vctrs.R`?

I'm now looking into other data types, it might make the source more easily browsable to keep the vctrs methods separate as we might end up with a non-trivial amount of those.

edzer:
Yes, good idea!

lionel-:
hmm, last Travis build failed when it shouldn't have:

```
  Error: processing vignette 'sf5.Rmd' failed with diagnostics:

   package or namespace load failed for 'tidyr':

    object 'cnd_bullets' not found whilst loading namespace 'vctrs'
```

This is because dev rlang and dev vctrs are temporarily in incompatible state.

Is it intended that Travis checks against dev rlang and dev vctrs? I was assuming we'd be testing against CRAN versions.

yutannihilation:
It seems because some `r_github_packages` is used by default. I guess they were necessary for development at the time when they were added. Are they still needed now?

https://github.com/r-spatial/sf/blob/fe70feab8375147bfa4c8299496785051c7f08e7/.travis.yml#L19-L24

lionel-:
Following up on https://github.com/r-spatial/sf/issues/1172#issuecomment-555911831, the CRS and precision attributes are now combined. I took the conservative approach of throwing an error if they are inconsistent. This can be changed to something more appropriate by modifying `common_crs()` and `common_prec()`.

edzer:
Looks good -- thanks so much!

PR #1208, opened 2019-12-01
Avoid PPA for newer Ubuntu versions?

wibeasley:
It isn't necessary for 19.10 eoan, is it?  Please triple-check me, because I'm making some guesses.

It appears to install fine for me without the PPA.  

```sh
E: Failed to fetch http://ppa.launchpad.net/ubuntugis/ubuntugis/ubuntu/dists/eoan/InRelease  403  Forbidden
```

* https://packages.ubuntu.com/eoan/libudunits2-dev
* https://packages.ubuntu.com/eoan/libgdal-dev
* https://packages.ubuntu.com/eoan/libgeos-dev
* https://packages.ubuntu.com/eoan/libproj-dev

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,12 @@ Alternatively [these instructions](https://stat.ethz.ch/pipermail/r-sig-mac/2017
 For Unix-alikes, GDAL (>= 2.0.1), GEOS (>= 3.4.0) and Proj.4 (>= 4.8.0) are required.
 
 #### Ubuntu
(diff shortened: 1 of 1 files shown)


edzer:
Yes, this would be good. Which versions do you get? (what is printed after loading `sf`?)

wibeasley:
Yay -it's the same versions on the repository webpages.  New session:

```r
R version 3.6.1 (2019-07-05) -- "Action of the Toes"
...

> library(sf)
Linking to GEOS 3.7.2, GDAL 2.4.2, PROJ 5.2.0
```

edzer:
Thanks; I think I'll add a note that the PPA has versions for ubuntu LTS. It doesn't say so, but that seems the case.

wibeasley:
Oh, that's a good clarification.

Issue #1213, opened 2019-12-09
How can I set INTERLEAVED_READING options with PBF files


agila5:
Hi! I'm working on a project on spatial networks in Greater London and I'm downloading the network data from [Geofabrik](https://download.geofabrik.de/europe/great-britain/england/greater-london.html). 

When I read the data I get a warning 
``` r
london_highways <- sf::st_read(
    dsn = "https://download.geofabrik.de/europe/great-britain/england/greater-london-latest.osm.pbf", 
    layer = "lines", 
    stringsAsFactors = FALSE
)
#> Reading layer `lines' from data source `https://download.geofabrik.de/europe/great-britain/england/greater-london-latest.osm.pbf' using driver `OSM'
#> Warning in CPL_read_ogr(dsn, layer, query, as.character(options), quiet, :
#> GDAL Error 1: Too many features have accumulated in points layer. Use
#> OGR_INTERLEAVED_READING=YES mode
#> Simple feature collection with 62788 features and 9 fields
#> geometry type:  LINESTRING
#> dimension:      XY
#> bbox:           xmin: -0.6124681 ymin: 51.2822 xmax: 0.3760979 ymax: 51.78736
#> epsg (SRID):    4326
#> proj4string:    +proj=longlat +datum=WGS84 +no_defs
```

<sup>Created on 2019-12-09 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

so I checked on [gdal website](https://gdal.org/drivers/vector/osm.html#open-options) on the meaning of that error and, after reading your answers in #1157, I tried to that set parameter. The problem is that none of these options work: 
``` r
london_highways <- sf::st_read(
    dsn = "https://download.geofabrik.de/europe/great-britain/england/greater-london-latest.osm.pbf", 
    layer = "lines", 
    options = c("OGR_INTERLEAVED_READING=YES"), 
    stringsAsFactors = FALSE
)
#> options:        OGR_INTERLEAVED_READING=YES 
#> Reading layer `lines' from data source `https://download.geofabrik.de/europe/great-britain/england/greater-london-latest.osm.pbf' using driver `OSM'
#> Warning in CPL_read_ogr(dsn, layer, query, as.character(options), quiet, :
#> GDAL Error 1: Too many features have accumulated in points layer. Use
#> OGR_INTERLEAVED_READING=YES mode
#> Simple feature collection with 62788 features and 9 fields
#> geometry type:  LINESTRING
#> dimension:      XY
[...]

PR #1217, opened 2019-12-12
Add one example to gdal_utils

agila5:
Following the discussion in #1213, I added one example at gdal_utils help page. I added the \dontrun tag so that it doesn't create a new object everytime it tests the examples but maybe it would be fine.

### DIFF
--- a/R/gdal_utils.R
+++ b/R/gdal_utils.R
@@ -20,7 +20,6 @@ resampling_method = function(option = "near") {
 }
 # nocov end
 
-
 #' Native interface to gdal utils
 #' @name gdal_utils
 #' @param util character; one of \code{info}, \code{warp}, \code{rasterize}, \code{translate}, \code{vectortranslate}, \code{buildvrt}, \code{demprocessing}, \code{nearblack}, \code{grid}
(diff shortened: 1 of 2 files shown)


edzer:
Thanks; for the same reason I did that too, this afternoon, see here: https://github.com/r-spatial/sf/blob/wkt2/R/gdal_utils.R#L35-L39 (where I was, mistakenly, still working in branch wkt2)

Please let me know how my example can be further improved.

agila5:
I would add both examples since they show two different options of vectortranslate utils (and I can change my example using the same tempfile structure as your example if you want) and I would add an example of another utils, maybe something like:

``` r
geomatrix_path <- system.file("tif/geomatrix.tif", package = "sf")

gdal_utils(
	util = "info",
	source = geomatrix_path, 
	options = c("-json")
)
```
I can create another commit with this example if you agree.

edzer:
Good idea!

agila5:
I've just update the examples following your suggestions.

R/gdal_utils.R
```diff
@@ -33,22 +33,43 @@ resampling_method = function(option = "near") {
 #' @export
 #' @examples
 #'
+#' # info utils can be used to list information about about a raster
+#' # dataset. More info: https://gdal.org/programs/gdalinfo.html
+#' in_file <- system.file("tif/geomatrix.tif", package = "sf")
+#'
 #' \dontrun{
```
edzer: why do we need a \dontrun{} here?

agila5: I forgot it from previous example 😅

edzer:
Great, thanks!

PR #1218, opened 2019-12-12
add st_recenter

tim-salabim:
This PR adds `st_recenter` as a generic for *.sf and *.sfc. It fails if `st_is_longlat()` is either `FALSE` or `NA`. The description text was basically copied from `sp::recenter` with a link included to this help page.

I've included two tests which are simply copied from the examples. 

See https://twitter.com/TimSalabim3/status/1205111889979727873 for context.


etiennebr:
It looks similar to [`ST_ShiftLongitude` in postgis](https://postgis.net/docs/ST_Shift_Longitude.html). If so, maybe we could reuse the name (or `st_shift_longitude`)?

tim-salabim:
I'll take @edzer 's thumbs up as a confirmation and rename the methods. Not before tomorrow though

tim-salabim:
renamed `st_recenter` to `st_shift_longitude`, added some more tests and added a NEWS.md entry

PR #1221, opened 2019-12-19
Change output of st_as_text() for MULTIPOINT to include inner parens. #1219

ateucher:
If merged, closes #1219.

As stated in the linked issue, the output of `lwgeom::st_astext()` will continue to omit the inner parens, which will cause an inconsistency unless PostGIS decides to go the same way. 

I know you're working on other higher priority issues (i.e., the `proj` stuff), so no big rush on this.

I have run `revdepcheck::revdep_check()` on all 214 reverse dependencies and found no issues:

``` r
library(revdepcheck)
revdep_report_summary()
#> # Platform
#> 
#> field      value                        
#> ---------  -----------------------------
#> version    R version 3.6.2 (2019-12-12) 
#> os         macOS Mojave 10.14.6         
#> system     x86_64, darwin15.6.0         
#> ui         X11                          
#> language   (EN)                         
#> collate    en_CA.UTF-8                  
#> ctype      en_CA.UTF-8                  
#> tz         America/Vancouver            
#> date       2019-12-19                   
#> 
#> # Dependencies
#> 
#> package    old     new       Δ  
#> ---------  ------  --------  ---
#> sf         0.8-0   0.8-999   *  
#> classInt   0.4-2   0.4-2        
#> DBI        1.1.0   1.1.0        
#> e1071      1.7-3   1.7-3        
#> magrittr   1.5     1.5          
#> Rcpp       1.0.3   1.0.3        
#> units      0.6-5   0.6-5        
#> 
#> # Revdeps
revdep_summary()
#> ✓ adklakedata 0.6.1                      ── E: 0     | W: 0     | N: 0    
#> ✓ amt 0.0.7                              ── E: 0     | W: 0     | N: 0    
#> ✓ arcos 0.8.2                            ── E: 0     | W: 0     | N: 0    
#> ✓ areal 0.1.5                            ── E: 0     | W: 0     | N: 0    
#> ✓ bcdata 0.1.2                           ── E: 0     | W: 0     | N: 0    
#> ✓ bcmaps 0.18.0                          ── E: 0     | W: 0     | N: 0    
#> ✓ bcmaps.rdata 0.1.5                     ── E: 0     | W: 0     | N: 0    
#> ✓ bcmapsdata 0.3.1                       ── E: 0     | W: 0     | N: 0    
#> ✓ BIOMASS 2.1.1                          ── E: 0     | W: 0     | N: 0    
#> ✓ biscale 0.1.2                          ── E: 0     | W: 0     | N: 0    
[...]

PR #1224, opened 2019-12-20
Update NEWS for #1221

ateucher:


edzer:
Thanks a lot, Andy!

ateucher:
Thank you!

PR #1226, opened 2019-12-23
Avoid side effects on PROJ_LIB when loading sf

lbusett:
This attempts to solve https://github.com/r-spatial/discuss/issues/31 for `sf`, by avoiding changing environment variables on load. Required changes are done on-the-fly upon calls to `st_crs` or `st_transform`, and automatically reverted on.exit. (See comments on last commit)

Hope this helps and makes sense.

PS2: `testthat` test are all passing, but many differences are found on tests comparing with Rout.save files. However, the differences seem similar to those also observed on travis build on `wkt2` branch.


lbusett:
...still failing on Appveyor - will dig further in a few days....

lbusett:
Closing this - much more difficult than expected, so it is much better to wait for GDAL3/PROJ6 support on Windows binary on R 4.0... 

Sorry for the clutter.

Issue #1227, opened 2019-12-26
sf_project() fails at S pole with proj=lcc


dankelley:
I noticed that `sf_project()` fails at the south pole, with the LCC projection.  A test code is provided below, along with results on my (osx catalina, R-devel) machine.

This came up for me in coding the oce package to use `sf::sf_project()` (reference https://github.com/dankelley/oce/issues/1629#issuecomment-569106096), but it is not a blocking issue for me, because I can make oce move such points to be a tiny bit north of the pole. Still, I thought I ought to point this out, in case doing so would be of use to the `sf` team.

```R
## Show that sf_project() fails at south pole, with Lambert Conic Conformal projection
library(sf)
from <- "+proj=longlat"
to <- "+proj=lcc +lat_0=30 +lat_1=60"
tol <- 1e-4
try(sf::sf_project(from, to, cbind(0, 90))) # OK
try(sf::sf_project(from, to, cbind(0, -90))) # fails
for (power in seq(-8, -11)) {
    eps <- 10^power
    cat(sprintf("eps=%15.12e\n", eps))
    try(sf::sf_project(from, to, cbind(0, -90 + eps)))
}
```

[sf_bug_report.txt](https://github.com/r-spatial/sf/files/4003245/sf_bug_report.txt)

edzer:
Thanks; this is a property of the underlying library, PROJ; using command line `proj`:
```bash
$ echo "0 -89.99999999" | proj +proj=lcc +lat_0=30 +lat_1=60
0.00	-5879242108888943.00
$ echo "0 -89.999999999" | proj +proj=lcc +lat_0=30 +lat_1=60
*	*
```
as this is not an issue of `sf`, I'm closing here.

dankelley:
Thanks for the very quick response.  I apologize for commenting on a closed issue, and would understand completely if you just ignore this, but I wonder whether I could make a respectful suggestion (after a prelude)...

*1. Prelude*

I tried something closer to my real application, with multiple points
```
sf echo "0 0\n0 -90\n0 1" | proj +proj=lcc +lat_0=30 +lat_1=60
0.00    -4329324.29
*       *
0.00    -4158078.42
```
i.e. it does not fail, but rather reveals problematic points

But I see why `st_transform()` is acting as it does, because its c++ function `CPL_proj_direct()` (at `src/proj.cpp` line 93) calls `proj_trans_array()`, which  https://proj.org/development/reference/functions.html reveals will return an error code if any single point is problematic.

[...]

PR #1228, opened 2019-12-28
let sf_project() handle unprojectable points by setting to Inf

dankelley:
**1. Preamble**

This commit follows up on the discussion at
https://github.com/r-spatial/sf/issues/1227 relating to a desire to offer the
user a way to make `sf::sf_project()` handle datasets that contain some
unprojectable points. This is done by adding a new argument named `keep` to
`sf::sf_project()`, which causes it to issue a warning and map problematic
values to `c(Inf,Inf)`, instead of issuing an error and not return anything.

But, before getting into the changes, it's important to address some points of
concern, which are worth checking into before considering merging this commit
into `sf`.

**2. Three preliminary notes of caution**

*2.1 Note 1*

I have access to the new proj API, so the code at and after `src/proj.cpp`
line 95 is what is exercised in the examples given below.  However, with no
access to the old proj API, I am not *at all* certain whether I have done the
right thing at and after `src/proj.cpp` line 248.  I would appreciate it
greatly if someone could check on my changes in the second spot.

*2.2 Note 2*

The commandline proj in the example I give below yields a y value of -4.3e6
metres, whereas the R code yields -3.0e6 metres. I don't think this is an
ellipse or offset problem since I get the same results with

```
echo "0 0\n0 -90\n0 0" | proj +proj=lcc +lat_0=30 +lat_1=60 +x_0=0 +y_0=0 +ellps=GRS80 +units=m
```

This difference is a concern to me, since it is so large.  For reference, `proj
--version` at the commandline yields "Rel. 6.2.1, November 1st, 2019", and

```R
library(sf)

```

yields

```R
Linking to GEOS 3.8.0, GDAL 2.4.2, PROJ 6.2.1

```

My R version is `Under development (unstable) (2019-12-17 r77592)`, although I
get the same concerning results with GEO 3.7.3 and PROJ 6.2.0.

I should note that my computer configuration is a bit odd, being partly set up
by apple, partly set up by `homebrew`, and partly by me. I think it is possible
that when I build `sf`, I am using a different version of the proj library or
some other library than what my commandline `proj` is using. Therefore, I am
pointing out this discrepancy not as a bug report, but rather to let others
know that this might be worth checking on a more conventional (e.g. linux)
system that is unlikely to rest on a frayed patchwork of system software.

*2.3 Note 3*

I cannot build vignettes because they involve running code that is not on my
machine, e.g. `edzer_or_travis`.  Also, the build tests fail, for various
reasons, e.g. dependences on `library(dplyr)`, `library(spatstat)`, etc., and
possibly other reasons.  (I have not checked into this carefully, but can send
the output from my checks, if desired.)

**3. Explanation of changes to sf::sf_project() and associated C++ code**

*3.1 Overview*

The main change is the addition of an argument, `keep`, to `sf_project()`.
That argument is set to `FALSE` by default, so that existing scripts using
`sf_project()` will continue to work as before. However, setting `keep=TRUE`
makes the unprojectable points map to the `Inf` values that `proj_trans()`
yields in such cases. (I assume `-Inf` values can also occur.)

*3.2 commandline proj test*

As an example, the south pole is unprojectable in the Lambert Conformal Conic
projection according to proj (Rel. 6.2.1, November 1st, 2019), as illustrated
in a (macos catalina 19D49f) terminal:

```
$ echo "0 0\n0 -90\n0 0" | proj +proj=lcc +lat_0=30 +lat_1=60
0.00    -4329324.29
*       *
0.00    -4329324.29
```
(See question 2 above.)

*3.3 sf_project old behaviour (and new default behaviour)*

The default call,

```R
pts <- cbind(c(0, 0, 0), c(0, -90, 0))
sf::sf_project("+proj=longlat", "+proj=lcc +lat_0=30 +lat_1=60", pts)
```

produces

```R
 Error in CPL_proj_direct(as.character(c(from[1], to[1])), as.matrix(pts),  :
  tolerance condition error > xy
```

and no output from the function call.

*3.4 sf_project old behaviour (new non-default behaviour)*

This behaviour is recovered (with `Inf` for the unprojectable point) with
`keep=TRUE`, viz.

```R
pts <- cbind(c(0, 0, 0), c(0, -90, 0))
sf::sf_project("+proj=longlat", "+proj=lcc +lat_0=30 +lat_1=60", pts, TRUE)
```

yields output

```R
     [,1]     [,2]
[1,]    0 -3026228
[2,]  Inf      Inf
[3,]    0 -3026228
Warning message:
In CPL_proj_direct(as.character(c(from[1], to[1])), as.matrix(pts),  :
  one or more projected point(s) not finite
```

For reference, `rgdal` reproduces commandline proj:

```R
> rgdal::project(pts, "+proj=lcc +lat_0=30 +lat_1=60")
     [,1]     [,2]
[1,]    0 -4329324
[2,]  Inf      Inf
[3,]    0 -4329324
```

I realize that the `sf` community will desire the default behaviour, which is
why it remains the default in this commit. However, the option to retain good
data, with some `Inf` values mixed in, would be of great benefit to users of
the `oce` package, which presently does map projection with rgdal.  Further to
that matter of switching from rgdal to sf I note that `rgdal::project()`
produces results identical to the commandline version of proj (this relates to
note 2 above).


src/proj.cpp
```diff
@@ -196,20 +211,22 @@ bool CPL_have_datum_files(SEXP foo) {
...
-Rcpp::NumericMatrix CPL_proj_direct(Rcpp::CharacterVector from_to, Rcpp::NumericMatrix pts) {
+Rcpp::NumericMatrix CPL_proj_direct(Rcpp::CharacterVector from_to, Rcpp::NumericMatrix pts, Rcpp::IntegerVector keep) {
 
 	using namespace Rcpp;
 
 	if (from_to.size() != 2)
 		stop("from_to should be size 2 character vector"); // #nocov
 	if (pts.ncol() != 2)
 		stop("pts should be 2-column numeric vector"); // #nocov
+        if (na.size() != 1)
```
edzer: na is not known: do you mean keep?

edzer:
You can look at the travis-ci check that failed, to get an idea what went wrong on a pre-PROJ6 platform.

src/proj.cpp
```diff
@@ -228,12 +245,30 @@ Rcpp::NumericMatrix CPL_proj_direct(Rcpp::CharacterVector from_to, Rcpp::Numeric
...
-
-	if (pj_transform(fromPJ, toPJ, pts.nrow(), 0, xx.data(), yy.data(), NULL) != 0) {
-		pj_free(fromPJ); pj_free(toPJ); // #nocov start
-		Rcout << "error in pj_transform: " << pj_strerrno(*pj_get_errno_ref()) << std::endl;
-		stop("error"); // #nocov end
-	}
+        if (keep[0] == 1) {
+            // use proj_trans() on individual points, making unprojectable points be NA
+            // FIXME: not tested, since author has no access to the old proj API.
+            PJ_COORD row;
```
edzer: remove this line?

src/proj.cpp
dankelley: Sorry, yes that should be 'keep' where I have 'na'.  This is a remnant; at first I named the argument differently, but forgot to change it.  I am not sure of the best argument name, actually.  So, line 222 needs `s/na/keep` and so does line 223.

I'm not too sure how to make a PR upon a PR, and since this is just two edit commands, I hope stating this is sufficient.

I see that you've pushed recently and travisCI is checking now.

Also, and sorry for this, but I polluted your files with one ending in `~`.  (That's because I use the vim editor, and I forgot to clean the file before doing my `git add`.  I normally have that listed in my `.gitignore` file.)  It was rude of me not to check what I was adding to the git repo.

src/proj.cpp
dankelley: Oh, I'm embarrassed now. Yes, that is a remnant.

src/proj.cpp
```diff
@@ -228,12 +245,29 @@ Rcpp::NumericMatrix CPL_proj_direct(Rcpp::CharacterVector from_to, Rcpp::Numeric
...
-	}
+        if (keep[0] == 1) {
+            // use proj_trans() on individual points, making unprojectable points be NA
+            // FIXME: not tested, since author has no access to the old proj API.
+            double thisx, thisy;
+            for (int i = 0; i < pts.nrow(); i++) {
+                thisx = xx[i];
+                thisy = yy[i];
+                if (pj_transform(fromPJ, toPJ, 1, 0, &thisx, &thisy, NULL) != 0) {
+                    xx[i] = NA_REAL;
```
edzer: According to the documentation, +Inf or -Inf values should be returned, and I guess this is what the PROJ6 code does.

dankelley: Yes, you are correct. I pushed an update to my fork, changing the assigment from `NA_REAL` to `R_PosInf`.  (I do not know whether to be doing new pull requests for these changes, or posting patches in these comments.  I apologize of my answering in this way is making more work for you, in addition to that occasioned by my sloppy coding.)

edzer: Did/does `rgdal` also return `Inf` values for unprojectable points?

dankelley: Yes, e.g. as below. I don't know about whether it ever gives `-Inf`, and so in my update to `src/proj.cpp` I made unprojectable points turn into `Inf`.
```R
> rgdal::project(pts, "+proj=lcc +lat_0=30 +lat_1=60")
     [,1]     [,2]
[1,]    0 -4329324
[2,]  Inf      Inf
[3,]    0 -4329324
```

edzer:
Thanks; I now manually merged this.

PR #1232, opened 2020-01-02
Remove redundant attributes(x) argument

krlmlr:
to `new_tibble()`.

This breaks in tibble 3.0.0, and never had an effect to begin with. The `...` expects named arguments, an unnamed list is ignored. On the up side, all arguments in `x` are retained.


edzer:
Thanks, @krlmlr !

Issue #1220, opened 2019-12-19
Error in `st_sample()` with `type = "hexagonal"` for unprojected polygons


dieghernan:
It seems that there is a bug on `st_sample` when `type="hexagonal"` is performed over an unprojected polygon:

```r
library(sf)
nc <- st_read(system.file("shape/nc.shp", package = "sf"))[1, ]
nc = st_transform(nc, 4326)
hexsample_4326 = st_sample(nc, size = 50, type = "hexagonal")
```

This returns this error:

```r 
Error in seq_len(nrow(xy)) : 
  argument must be coercible to non-negative integer
In addition: Warning message:
In seq_len(nrow(xy)) : first element used of 'length.out' argument
```

Whereas this works fine:

```r
hexsample_other = st_sample(st_transform(nc, 3857), size = 50, type = "hexagonal")
```

I think the issue comes from this line

https://github.com/r-spatial/sf/blob/b96ffe8061307cbd302d65d868a4d45c9b26f94f/R/sample.R#L101 

as it returns the area on m2, and then
https://github.com/r-spatial/sf/blob/b96ffe8061307cbd302d65d868a4d45c9b26f94f/R/sample.R#L113
comes in meters as well, and finally

https://github.com/r-spatial/sf/blob/b96ffe8061307cbd302d65d868a4d45c9b26f94f/R/sample.R#L180-L186

performs the operations mixing longitude and latitude with meters.

edzer:
Yes, I can see this, but wonder whether this is better than giving an error message telling that users should first convert data to an equidistant rectangular (`+proj=eqc`) projection to get meaningful regular and hexagonal samples.

PR #1237, opened 2020-01-09
st_sample for unprojected polygons 

dieghernan:
Small changes applied in order to make `st_sample` to work with unprojected shapes. `a0` parameter (area) is now computing as width*length of the bounding box when `st_is_longlat(x)==TRUE`.

Following #1220


dieghernan:
Cancelling request

PR #1252, opened 2020-01-28
add a blurb to ?st_sf about agr argument

### LINKED ISSUE #1251: Examples for agr in documentation
I'm trying to learn how to use `st_as_sf`; it points me to `?st_sf` for understanding its `agr` argument, which is described in `agr` with:

> `agr`, attribute-geometry-relationship, specifies for each non-geometry attribute column how it relates to the geometry, and can have one of following values: `"constant"`, `"aggregate"`, `"identity"`. `"constant"` is used for attributes that are constant throughout the geometry (e.g. land use), `"aggregate"` where the attribute is an aggregate value over the geometry (e.g. population density or population count), `"identity"` when the attributes uniquely identifies the geometry of particular "thing", such as a building ID or a city name. The default value, `NA_agr_`, implies we don't know.

This sounds basically OK, but I still don't know how to _use it_. Do we supply `agr` as a vector? Named list? I've only come across a few examples in a quick search and they all supply only one `agr` value, so the "for each non-geometry... column" part is unexplored.

MichaelChirico:
Closes #1251 

Thanks @edzer, I added a quick blurb with more detail, feel free to adjust as you see fit.

Again my editor removed some trailing whitespace it seems; `-w` version:

https://github.com/r-spatial/sf/pull/1252/files?w=1

### DIFF
--- a/R/sf.R
+++ b/R/sf.R
@@ -199,7 +199,9 @@ list_column_to_sfc = function(x) {
 #' there is more than one and \code{sf_column_name} is \code{NULL}, the first one is taken.
 #' @param sfc_last logical; if \code{TRUE}, \code{sfc} columns are always put last, otherwise column order is left unmodified.
 #' @param check_ring_dir see \link{st_read}
-#' @details \code{agr}, attribute-geometry-relationship, specifies for each non-geometry attribute column how it relates to the geometry, and can have one of following values: "constant", "aggregate", "identity". "constant" is used for attributes that are constant throughout the geometry (e.g. land use), "aggregate" where the attribute is an aggregate value over the geometry (e.g. population density or population count), "identity" when the attributes uniquely identifies the geometry of particular "thing", such as a building ID or a city name. The default value, \code{NA_agr_}, implies we don't know.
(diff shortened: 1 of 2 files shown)


R/sf.R
```diff
@@ -201,7 +201,9 @@ list_column_to_sfc = function(x) {
...
 #'
-#' When confronted with a data.frame-like object, `st_sf` will try to find a geometry column of class `sfc`, and otherwise try to convert list-columns when available into a geometry column, using \link{st_as_sfc}.
+#' When a single value is provided to \code{agr}, it is cascaded across all input columns; otherwise, a named vector like \code{c(feature1='constant', ...)} will set \code{agr} value to \code{'constant'} for the input column named \code{feature1}. See \code{demo(nc)} for a worked example of this.
+#'
+#' When confronted with a data.frame-like object, \code{st_sf} will try to find a geometry column of class \code{sfc}, and otherwise try to convert list-columns when available into a geometry column, using \link{st_as_sfc}.
```
MichaelChirico: @edzer  heads up I pushed a second commit, it seems only a few seconds before you hit "merge".

I thought `devtools::document('.')` would interpret backticks in `roxygen` sections and map them to `\code{}` but it seemed not from the output `.Rd` file, so I reverted to using explicity `\code{}`

edzer: Thanks, I'll be more patient in the future.

MichaelChirico: no worries, I should have marked it as WIP once I saw a 2nd commit was needed. Thanks for the speed in merging! I caught you at the right time.

PR #1257, opened 2020-02-03
Wkt2 PR just to let it build

rsbivand:
Only d8815d2 really, maybe the previous small fix.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: sf
-Version: 0.8-1
+Version: 0.8-3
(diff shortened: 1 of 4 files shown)


edzer:
Thanks! Please note that branch `SetFromUserInput` will be the next release, it's a fork of `wkt2` branch but I dropped work on `wkt2`.

rsbivand:
So should I rather follow `Set FromUserInput`?

edzer:
Yes: wkt2 will be dropped.

PR #1258, opened 2020-02-03
SetFromUserInput update version and configure

rsbivand:
Same PR on other branch


edzer:
FYI - in bash "or" is `||`, not `|`

PR #1263, opened 2020-02-11
Set from user input

rsbivand:
I've added in R/sp.R a repeated assignment because the `handle_bbox()` abstraction does not take NG WKT2 into account.


edzer:
I see why this is needed, but would like to to this at the point where the crs is set in the first place, in the call to `st_as_sfc.SpatialXxx`. (Why) must the warning be suppressed?

rsbivand:
The warning comes because df$geometry's `"crs"` is set in the odd `handle_bbox()` use of the naked and deprecated p4s string accessed through two `@`'s, so has to be re-set, overwriting the wrong value. Would it be better if `handle_bbox()` just assigned character `NA`?

edzer:
In this alternative to your PR (https://github.com/r-spatial/sf/commit/65f84f15b74429be73fa6cc93a2a45c9cf7c3482), the crs is directly set using `st_crs.CRS`. It doesn't suppress a warning, but that is not needed if we set it only once, right?

rsbivand:
https://github.com/r-spatial/sf/commit/65f84f15b74429be73fa6cc93a2a45c9cf7c3482 looks good - changed to branch sp_crs.

edzer:
I'll merge `sp_crs` into `SetFromUserInput`.

rsbivand:
OK, I'll go back to SetFromUserInput at next pull.

PR #1271, opened 2020-02-18
`Euclidian` and `Haussdorff` distance args misspelled in docs

jwolfson:
Giving the arguments `which = "Euclidian"` and `which = "Haussdorff"` (as implied by docs) to `st_distance` throws an error.

Using the corrected spellings gives the anticipated result.


edzer:
Great, thanks!

PR #1274, opened 2020-02-20
Test build GDAL3 stack on Windows

jeroen:
This is a test run with the [rwinlib/gdal3](https://github.com/rwinlib/gdal3) stack to be able to see if there are unexpected issues in either `sf` or the windows builds. The stack may change, is not ready to be merged.

I think we should wait for R 4.0 before shipping these changes to CRAN because it may cause temporary disruption with the gdal/proj data files and environment variables as some R packages are still using gdal2 and others gdal3.

PR #1275, opened 2020-02-20
Jeroen gdal3 rwinlib

edzer:


jeroen:
FYI, I'll be on vacation until Mar 13, we can pick it up after that. In the mean time, perhaps it would be wise to run a revdep check on Windows to test for any unexpected breakage in dependent packages. That gives us a bit more testing coverage to ensure that the new rwinlib stack is stable.

mdsumner:
I'm finally in a place where I can check this, so far so good ...

edzer:
Great work -- I think this is ready to go. Any objections?

jeroen:
I think this is good 👍 Will rgdal switch switch as well ?

rsbivand:
Thanks for the heavy lifting @jeroen, yes. I have a win 10 R 4.0.0 machine
to try things out. Is it mostly changing the download link and the linking
flags?

Roger Bivand
Falsensvei 32
5063 Bergen

lør. 25. apr. 2020, 11:59 skrev Jeroen Ooms <notifications@github.com>:

jeroen:
Yes, the required changes are in my original PR here (which edzer further tweaked in this PR): https://github.com/r-spatial/sf/pull/1274/files

mdsumner:
@jeroen are you planning to separate PROJ for 6.3.1 so it's independent? I found issues with 6.1.0 so it would be good to have 6.3.1 on its own independent of this gdal bundle

jeroen:
@mdsumner what do you mean? Would you like me to host a copy of just proj 6.3.1 on https://github.com/rwinlib/proj ?

mdsumner:
Yes, thank you

jeroen:
@mdsumner OK I've added a copy and made a PR for you: https://github.com/hypertidy/PROJ/pull/22

rsbivand:
@jeroen Is it possible to detect and condition on the RTools version that the user is running? I'm afraid that people with an RTools will accept install from source on Windows, and we'll get errors reported in because they are mixing versions? Can we use that to pick GDAL3 or GDAL2, and whch set of `PKG_LIBS` to go with?

jeroen:
@rsbivand don't worry, the binaries in rwinlib are compatible with both the old and new toolchain.

rsbivand:
Thanks! Current devel **rgdal** Win binaries built using GDAL 3.0.4 and PROJ 6.3.1 linked from https://stat.ethz.ch/pipermail/r-sig-geo/2020-April/028063.html

rsbivand:
@jeroen On running revdeps for **sp**, **rgdal** and **raster**, I'm seeing three fatal errors: `fatal error: gsl/gsl_rng.h: No such file or directory` in **hSDM**, **jSDM** and **landsepi**. In the latter case, https://gitlab.paca.inrae.fr/CSIRO-INRA/landsepi/-/blob/master/tools/winlibs.R seems to want to extract GSL from the GDAL2 archive.**jSDM** depends on RcppGSM, and **hSDM** expects GSL to be installed; both declare GSL a system requirement. I guess I haven't installed GSL properly, will check. Otherwise, **imager** says `fatal error: tiffio.h: No such file or directory`, but again, I guess I haven't followed up a system requirement. On balance, nothing related to the GDAL2/3 switch. I'm revdeping about 900 packages, so only 4 anomalies isn't bad. So no red lights, I think.

jeroen:
@rsbivand those are packages that have system dependencies that are not automatically installed. If you are using rtools40 you can install those sysdeps as follows:

```sh
pacman -S mingw-w64-{i686,x86_64}-gsl
pacman -S mingw-w64-{i686,x86_64}-libtiff
```

Afterwards you should be able to check the above packages in R 4.0.

PR #1279, opened 2020-02-24
Update sf1.Rmd

renejuan:
I think st refers to spatial type according to postgis page https://postgis.net/docs/reference.html


edzer:
Thanks!

Issue #1266, opened 2020-02-13
Warn user when appending to existing GPKG (or modify default)


lbusett:
I recently found out that, when saving to GPKG using `sf::st_write()`, if the target file already exists the current default setting is to append features to the already existing ones, such as in : 

``` r
nc <- sf::st_read(system.file("shape/nc.shp", package="sf"), quiet = TRUE)
dim(nc)
[1] 100  15

sf::st_write(nc, "/home/lb/tmp/nc.gpkg")
Updating layer `nc' to data source `/home/lb/tmp/nc.gpkg' using driver `GPKG'
Writing 100 features with 14 fields and geometry type Multi Polygon.

sf::st_write(nc, "/home/lb/tmp/nc.gpkg")
Updating layer `nc' to data source `/home/lb/tmp/nc.gpkg' using driver `GPKG'
Updating existing layer nc
Writing 100 features with 14 fields and geometry type Multi Polygon.

dim(sf::st_read("/home/lb/tmp/nc.gpkg", quiet = TRUE))
[1] 200  15
``` 
while usually, when saving to "vector files" in other formats (e.g., json, shp, kml), the default is to fail on attempts to write to an existing file, unless delete_layer/delete_dsn is specified . 

Although this is clearly documented, this took me by surprise and led to problems in my current analysis which (and I guess it could surprise also other users). 

I was thus wondering if it could make sense to: 

- change this default behaviour for GPKG datasets (which, although being a SQLITE DB, are in my experience usually "treated" like normal files, ) - thus conforming with `write_sf()` defaults; 

OR/AND, 

-  provide a "stronger" warning when appending happens. The current message about "updating" (__Updating existing layer nc__) is in my opinion quite easy to go unnoticed. Would you consider issuing a Warning, in addition, when trying to write to gpkg on an existing file if delete_layer is FALSE ?

edzer:
I am against considering GPKG as a non-database drivers. I can imagine to issue a warning if updating will take place _without_ `update = TRUE` being specified, although this will hurt users counting on the current behaviour. Opinions?

rsbivand:
I would suggest issuing a message rather than a warning if an existing layer is being updated. I've been bitten by this, and, like Lorenzo, see GPKG as a file format first and foremost, like SQLite files, because the files are portable and need no server. My preference would be for the split to go between database formats needing a server, and file formats that don't need a user-visible server.

edzer:
But in terms of user experience, a message is no different from the current output, is it?

[...]

PR #1294, opened 2020-03-06
Make writing behavior of database files consistent with other files

etiennebr:


etiennebr:
@edzer I opened a PR to make it easier to track the changes.

cc #1266

PR #1306, opened 2020-03-17
default geotransform

mdsumner:
Currently, GDAL raster read invalidates the default geotransform setting it to NA, so we cannot plot non-geo rasters read with stars: 

https://github.com/r-spatial/sf/blob/master/src/stars.cpp#L312-L318

Is there any reason not to return the default transform (0,1,0,0,0,1)?  https://gdal.org/doxygen/classGDALDataset.html#a5101119705f5fa2bc1344ab26f66fd1d

Currently: 
```R
f <- file.path(R.home(), "Tcl/lib/tk8.6/images/logo100.gif")
sf::gdal_read(f)$geotransform
#[1] NA NA NA NA NA NA
```

I'm inclined to detect the situation and set it to (0,1,0,0,0, -1) so that stars plots in the likely correct orientation, with a warning as is common in geo-sw that reads images. But, that also means we should consider shifting the raster up to put it in [xmin = 0, ymin = 0, xmax = ncol, ymax = nrow] (a question for stars). 

With this PR: 

- [x] set default geotransform (0,1,0,0,0,1)
- [x] negate pixel height (y-spacing)  for convenience (0,1,0,0,0,1)
- [x] return flag "default_geotransform" = 0 (normal geotransform), = 1 (default geotransform with y-negate)
- [ ] issue warning about default geotransform (in sf, stars ...)
- [ ] perhaps the extent should also change, to push the raster north so it's in positive ymin = 0?
 
``` r
    f <- file.path(R.home(), "Tcl/lib/tk8.6/images/logo100.gif")
sf::gdal_read(f)$geotransform
#> [1]  0  1  0  0  0 -1

## this means we can plot straight away
## and the negative y puts the image the "right way up"
plot(stars::read_stars(f))
```

``` r

## we also have a new field to indicate the behaviour
sf::gdal_read(f)$default_geotransform
#> [1] 1
```

<sup>Created on 2020-03-18 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### DIFF
--- a/src/stars.cpp
+++ b/src/stars.cpp
@@ -37,7 +37,7 @@ CharacterVector get_meta_data(GDALDatasetH ds, CharacterVector domain_item) {
 }
 
 List get_band_meta_data(GDALDataset *poDataset) {
-	int n_bands = poDataset->GetRasterCount(); 
+	int n_bands = poDataset->GetRasterCount();
 	List ret(n_bands);
 	for (int band = 1; band <= n_bands; band++) { // unlike x & y, band is 1-based
 		GDALRasterBand *poBand = poDataset->GetRasterBand( band );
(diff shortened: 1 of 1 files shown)


mdsumner:
Of course, all of this could go into R-code in stars itself - detecting the NA geotransform there and acting acoordingly. Happy to change tack if that's preferred.

edzer:
Makes sense - thanks!

PR #1307, opened 2020-03-18
don't negate Y-res in default geotransform

mdsumner:
I think this should return as per the GDAL documentation - with Y-res = 1. 

Cf https://github.com/r-spatial/sf/pull/1306

This way, the `gdal_read()` function does as GDAL says, and other tools can apply the Y-flip and shift to put a non-geo raster into natural orientation, and with [0, nrow, 0, ncol]. 

I had more to follow up with stars but that's stalled for now


edzer:
I'm going to revert this back, because otherwise
```r
r = read_stars("~/svn/rgdal/pkg/inst/pictures/Rlogo.jpg")
```
gives

x

mdsumner:
Well, it should be changed in stars which is what I said above ...

PR #1322, opened 2020-03-27
Early skip for solaris

### LINKED ISSUE #1321: sf failing on Solaris
See https://www.r-project.org/nosvn/R.check/r-patched-solaris-x86/sf-00check.html

@etiennebr could you please modify tests that require RPostgres so that they are not being run on solaris?

etiennebr:
RPostgres is not available for solaris

close #1321


edzer:
Thanks!

PR #1337, opened 2020-04-03
Issue1332

dcooley:
I've added [these](https://github.com/r-spatial/sf/compare/master...dcooley:issue1332?expand=1#diff-b14e8733b035cbed978d319881866251R5) `get_m_position()` and `get_z_postiion()` functions to check the objects are the correct size before they are subset.

This should solve the valgrind warnings, and I believe the reported crashes.

However, I have an outstanding question where the `m_range` is being printed as (0,0), where it logically should be `NA_m_range_` as it doesn't exist (the dimension is `XYZ`)

```r
# example taken from issue 1332
read_sf("~/Desktop/Fisheries/FisheriesRecreation.gdb/", "bathymetry")
# Simple feature collection with 4887 features and 5 fields
# geometry type:  MULTILINESTRING
# dimension:      XYZ
# bbox:           xmin: 218674 ymin: 4494114 xmax: 713612.8 ymax: 4818581
# z_range:        zmin: 0 zmax: 0
# m_range:        mmin: 0 mmax: 0
# projected CRS:  NAD83 / UTM zone 15N
```

I would expect it to behave the same as these two examples, where only the relevant z or m range is printed, given the dimension of the object.

```r
sf_z <- sf::st_read(system.file("/shape/storms_xyz.shp", package = "sf"), quiet = TRUE)
sf_m <- sf::st_read(system.file("/shape/storms_xyzm.shp", package = "sf"), quiet = TRUE)

sf_z
# Simple feature collection with 71 features and 0 fields
# geometry type:  LINESTRING
# dimension:      XYZ
# bbox:           xmin: -102.2 ymin: 8.3 xmax: 0 ymax: 59.5
# z_range:        zmin: 924 zmax: 1017
# CRS:            NA
# First 10 features:
# 	geometry

sf_m
# Simple feature collection with 71 features and 0 fields
# geometry type:  LINESTRING
# dimension:      XYM
# bbox:           xmin: -102.2 ymin: 8.3 xmax: 0 ymax: 59.5
# m_range:        mmin: 924 mmax: 1017
# CRS:            NA
# First 10 features:
# 	geometry

```

### DIFF
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -325,11 +325,11 @@ CPL_write_wkb <- function(sfc, EWKB = FALSE) {
     .Call('_sf_CPL_write_wkb', PACKAGE = 'sf', sfc, EWKB)
 }
 
-CPL_get_z_range <- function(sf, depth = 0L) {
+CPL_get_z_range <- function(sf, depth) {
     .Call('_sf_CPL_get_z_range', PACKAGE = 'sf', sf, depth)
 }
 
-CPL_get_m_range <- function(sf, depth = 0L) {
+CPL_get_m_range <- function(sf, depth) {
     .Call('_sf_CPL_get_m_range', PACKAGE = 'sf', sf, depth)
 }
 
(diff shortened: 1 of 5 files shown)


edzer:
Looks like that did it. Thanks @dcooley !

PR #1340, opened 2020-04-05
Update MacOS install instructions

ateucher:


edzer:
Thanks!

rsbivand:
Please be aware of big changes coming for 4.0.0, and on-going discussions on R-sig-mac: https://stat.ethz.ch/pipermail/r-sig-mac/2020-April/013289.html and very definite advice not to use homebrew in general for regular OSX users. We really need to stop OSX users installing from source (such as adding a configure argument to **sf** etc. requiring positive assent to consequences of not using CRAN binaries - you're on your own, do not raise issues on this site). The underlying problem is that the RStudio install functions suggest installing from source if source is newer, and users go for it without understanding what is needed in terms of abilities to manage a software installation with extensive system requirements.

edzer:
With the CRAN MacOSX binary build train both having outdated GDAL/GEOS/PROJ and being unreliable (no sf 0.9-0 built after 13 days now, I don't believe this is going to happen without us reminding Simon to once more give it a kick), I'm not sure I would agree entirely if I were using a Mac.

IIRC, `install.packages()` asks you to choose if source is newer than available binary. It's docs say: "     For a binary install from a repository, the function checks for the availability of a source package on the same repository, and  reports if the source package has a later version, or is available but no binary version is."

ateucher:
As far as I understand it, the strong recommendations are to not use **R** itself from homebrew, nor the compiler toolchains etc, and I agree with this. Libraries like the geo suite are usually fine. I agree that users are almost always better off using binaries, but sometimes it is necessary to use source installs, especially package developers.

As for the current need to use `configure.args`, I understood this to be because of an incorrect `proj.pc` in `PROJ 7.0.0`, not necessarily because of homebrew. Before this version of `proj` I've always been able to install `sf` from source on my mac with the standard `gdal`, `proj`, and `geos` from homebrew.

Robinlovelace:
Just an idea, seeing all this great work on MacOS testing, would it be worth writing a short blog post for the r-spatial website on installing spatial packages on Mac? By the looks of it @ateucher could be well-placed to write such an article.

That reminds me I said I'd put in a PR to get the 'installing on Ubuntu' post on r-spatial, just a matter of time...

PR #1348, opened 2020-04-08
Add skip_if_not_installed

daissi:
Hi,

Several tests fail in Debian due to some missing R packages that are not officially packaged for Debian. This patch prevents some errors.

https://salsa.debian.org/r-pkg-team/r-cran-sf/-/blob/master/debian/patches/Skip_tests.patch

Best,
Dylan


edzer:
Thanks!

PR #1355, opened 2020-04-09
Update URL field to put pkgdown URL first

paleolimbot:
For whatever reason, pkgdown uses the first URL in the DESCRIPTION to do its automagic linking. This tiny edit should fix that for sf!

(what current links to sf look like: https://r-spatial.github.io/stars/articles/stars1.html#cropping-a-rasters-extent ...the dplyr functions automagically link to the dplyr pkgdown site)


edzer:
Good catch - thanks!!

edzer:
But it looks like they still do, right?

edzer:
pkgdown uses CI, and reflects dev version.

PR #1357, opened 2020-04-10
Clarify installation instructions

Robinlovelace:


edzer:
Am I right this would imply we recommend using PROJ 4.9.3 on systems that still carry that? I don't agree.

Robinlovelace:
Before it didn't provide any recommendation but I see what you mean. Do you agree with this?:

> To get more up-to-date versions of dependencies such as GDAL (recommended)

edzer:
I would just go for saying "we recommend using the PPA". Indeed, it doesn't say so right now, the current text was written in 2017 when this was less of an issue (before the PROJ storm).

Robinlovelace:
> I would just go for saying "we recommend using the PPA"

Done. Also tried to 'un-recommend' installing dependencies by source by adding the bit about 'advanced users'. Installing the OSGeo stack from source seems to cause problems.

edzer:
OK, thanks!

PR #1360, opened 2020-04-11
Move shell command onto 1 line

Robinlovelace:
Realise after looking at http://r-spatial.github.io/sf/ that this will be easier to read and use, and is consistent with other code chunks in the readme.


Robinlovelace:
:tada:

PR #1368, opened 2020-04-15
Add -fPIC flag using R config PIC variant

lepennec:
The latest sf version does not compile under fedora: configure yields
```
checking GDAL: linking with --libs only... yes
checking GDAL: /usr/share/gdal/pcs.csv readable... yes
checking GDAL: checking whether PROJ is available for linking:... no
/usr/bin/ld: /tmp/ccJ892zj.o: relocation R_X86_64_32 against symbol `__gxx_personality_v0@@CXXABI_1.3' can not be used when making a PIE object; recompile with -fPIC
/usr/bin/ld: final link failed: nonrepresentable section on output
collect2: error: ld returned 1 exit status
configure: Install failure: compilation and/or linkage problems.
configure: error: cannot link projection code
```
Using the PIC version of the FLAGS with R Config solves the issue.

### DIFF
--- a/configure
+++ b/configure
@@ -656,7 +656,6 @@ infodir
 docdir
 oldincludedir
 includedir
-runstatedir
 localstatedir
 sharedstatedir
 sysconfdir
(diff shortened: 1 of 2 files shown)


edzer:
Did you check [here](https://cran.r-project.org/web/checks/check_results_sf.html)?

rsbivand:
No problem with fully updated Fedora 31, R 4.0.0 beta, and **sf** from CRAN.

edzer:
I think this was resolved by adding `CXXFLAGS` to `configure.ac` in #1369; let me know if I am wrong.

PR #1374, opened 2020-04-17
add subset assignment for vctrs + sfc

paleolimbot:
This is a PR to support vctrs-style subset assignment for sfc objects. In vctrs, subset assignment does a `vec_cast()` to the type of `x`. The motivation is to support functions like [geovctrs::geo_coalesce()](https://github.com/paleolimbot/geovctrs/blob/master/R/geo-coalesce.R#L19-L33) that rely on subset-assignment for compatibility with sf.

sf already has a `[<-` method for sfc, and so I tried to do this in such a way that supported vectors created using `vctrs::new_vctr()`, which are likely to have a `vec_cast()` method implemented for sfc objects (if they are compatible with sf).

Happy to change...not knowing the history of `[<-` method for sfc, I'm not sure if this introduces problems.


edzer:
Did you align this with the author of the vctrs stuff in `sf`?

edzer:
The authors of https://github.com/r-spatial/sf/blob/master/R/tidyverse-vctrs.R - @krlmlr and @lionel- ?

R/sfc.R
```diff
@@ -142,8 +142,12 @@ sfg_is_empty = function(x) {
 #' @export
 #"[<-.sfc" = function (x, i, j, value) {
 "[<-.sfc" = function (x, i, value) {
-	if (is.null(value) || inherits(value, "sfg"))
+	if (is.null(value) || inherits(value, "sfg")) {
 		value = list(value)
+	} else if(inherits(value, "vctrs_vctr")) {
+		value = vctrs::vec_cast(value, vctrs::vec_ptype(x))
```
krlmlr, line 148: Do we need `vec_ptype()` here?

```suggestion
		value = vctrs::vec_cast(value, x)
```

R/sfc.R
```diff
@@ -142,8 +142,12 @@ sfg_is_empty = function(x) {
 #' @export
 #"[<-.sfc" = function (x, i, j, value) {
 "[<-.sfc" = function (x, i, value) {
-	if (is.null(value) || inherits(value, "sfg"))
+	if (is.null(value) || inherits(value, "sfg")) {
 		value = list(value)
+	} else if(inherits(value, "vctrs_vctr")) {
```
lionel-, line 147:
```suggestion
	} else if (inherits(value, "vctrs_vctr")) {
```

edzer:
For that,
```r
st_as_sfc.sfg = function(x,...) st_sfc(x,...)
```
would be enough; would make sense to add.

edzer:
Sorry I lost track -- is this obsolete, now, or still needed?

PR #1382, opened 2020-04-26
For upcoming changes in "raster"

rhijmans:
Foreshadowing upcoming changes in the crs slot of Raster objects


edzer:
Thanks!

rsbivand:
@rhijmans Can you handle the **sp** `""CRS""` interface without code changes in **sp** or **rgdal**? I've added support in `rgdal::rawTransform()` for PROJ strings only so far, as I don't know what you are planning for **raster**.

PR #1405, opened 2020-05-26
Add `dplyr_reconstruct()` method

lionel-:
This method is used to restore the class of the output of `bind_rows()` and `bind_cols()` to the class of the first input.

With CRAN dplyr:

```r
sf1 = st_sf(x = 1, y = st_sfc(st_point(0:1)))
sf2 = st_sf(z = st_sfc(st_point(2:3)), x = 2)

bind_rows(sf1, sf2)
#> Error in .subset2(x, i, exact = exact) :
#>   attempt to select less than one element in get1index
#> In addition: Warning messages:
#> 1: In bind_rows_(x, .id) :
#>   Vectorizing 'sfc_POINT' elements may not preserve their attributes
#> 2: In bind_rows_(x, .id) :
#>   Vectorizing 'sfc_POINT' elements may not preserve their attributes
```

With dplyr 1.0:

```r
bind_rows(sf1, sf2)
#> Simple feature collection with 2 features and 1 field
#> Active geometry column: y (with 1 geometry empty)
#> geometry type:  POINT
#> dimension:      XY
#> bbox:           xmin: 0 ymin: 1 xmax: 0 ymax: 1
#> CRS:            NA
#>   x           y           z
#> 1 1 POINT (0 1) POINT EMPTY
#> 2 2 POINT EMPTY POINT (2 3)
```

PR #1407, opened 2020-05-27
Fix tests

lionel-:
Follow-up to https://github.com/r-spatial/sf/pull/1405#issuecomment-634603739

`test_tidy.R` and `test_vctrs.R` now succeed with combinations of CRAN and dev dplyr, and CRAN and dev vctrs.


lionel-:
@edzer I'm going to send vctrs 0.3.1 very soon to CRAN because it fixes the sfc handling. It's going to cause failures in the CRAN results for sf, but these are only about the failing assumptions in unit tests that were fixed here.

PR #1417, opened 2020-06-05
Fix renaming of geometry columns in `dplyr::select()`

### LINKED ISSUE #1415: allow for renaming while selecting geometry column
we get an error when trying to do this:

``` r
library(sf)
#> Linking to GEOS 3.8.0, GDAL 3.0.4, PROJ 6.3.1
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
read_sf(system.file("shape/nc.shp", package="sf")) %>% 
  select(geom_new_name = geometry)
#> Error in st_sf(x, ..., agr = agr, sf_column_name = sf_column_name): sf_column_name %in% all_sfc_names is not TRUE
```

<sup>Created on 2020-06-05 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

lionel-:
Closes #1415

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -92,6 +92,7 @@ Suggests:
     testthat,
     tibble (>= 1.4.1),
     tidyr (>= 1.0-0),
+    tidyselect (>= 1.0.0),
     tmap (>= 2.0),
     vctrs
 LinkingTo: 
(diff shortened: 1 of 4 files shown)


edzer:
Great, thanks a lot!

Issue #1401, opened 2020-05-25
macOS install instructions


pat-s:
I am able to install {sf} 0.9-3 from source on macOS without the need for custom `configure.args`.

With 

- proj 7.0.1
- gdal 2.4.4_6
- geos 3.8.1_1

on R 4.0.0 on latest macOS 10.15.4 

`proj`, `gdal` and `geos` installed (and linked) via `brew`. No custom Makevars set.
Might be worth updating the install instructions in the README?

<details>
<summary>Details</summary>

```r
configure: CC: ccache clang
configure: CXX: ccache clang++ -std=gnu++11
checking for gdal-config... /usr/local/bin/gdal-config
checking gdal-config usability... yes
configure: GDAL: 2.4.4
checking GDAL version >= 2.0.1... yes
checking for gcc... ccache clang
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables...
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether ccache clang accepts -g... yes
checking for ccache clang option to accept ISO C89... none needed
checking how to run the C preprocessor... ccache clang -E
checking for grep that handles long lines and -e... /usr/bin/grep
checking for egrep... /usr/bin/grep -E
checking for ANSI C header files... yes
checking for sys/types.h... yes
checking for sys/stat.h... yes
checking for stdlib.h... yes
checking for string.h... yes
checking for memory.h... yes
checking for strings.h... yes
checking for inttypes.h... yes
checking for stdint.h... yes
checking for unistd.h... yes
checking gdal.h usability... yes
checking gdal.h presence... yes
checking for gdal.h... yes
checking GDAL: linking with --libs only... yes
checking GDAL: /usr/local/Cellar/gdal/2.4.4_6/share/gdal/pcs.csv readable... yes
checking GDAL: checking whether PROJ is available for linking:... yes
checking GDAL: checking whether PROJ is available fur running:... yes
configure: GDAL: 2.4.4
configure: pkg-config proj exists, will use it
configure: using proj.h.
configure: PROJ: 7.0.1
checking PROJ: checking whether PROJ and sqlite3 are available for linking:... yes
checking for geos-config... /usr/local/bin/geos-config
checking geos-config usability... yes
configure: GEOS: 3.8.1
checking GEOS version >= 3.4.0... yes
checking geos_c.h usability... yes
checking geos_c.h presence... yes
[...]

PR #1435, opened 2020-06-26
Fix macOS GitHub actions

ateucher:
- remove `configure.args` for sf and rgdal
- remove `--no-build-vignettes` build arg
- ensure rgdal installed from source with same libs as sf
- remove ENV vars for sf and xml2
- add cache version env var to enable manual cache invalidation
- Add `GITHUB_PAT` env var to avoid GitHub rate limits (to both actions workflow files)
- use `release` and `oldrel` for R versions rather than specific version numbers

As mentioned in #1401

### DIFF
--- a/.github/workflows/R-CMD-check-with-db.yaml
+++ b/.github/workflows/R-CMD-check-with-db.yaml
@@ -70,7 +70,7 @@ jobs:
         env:
           RHUB_PLATFORM: linux-x86_64-ubuntu-gcc
         run: |
-          Rscript -e "remotes::install_github('r-hub/sysreqs')"
+          Rscript -e "remotes::install_url("https://github.com/r-hub/sysreqs/archive/HEAD.zip")"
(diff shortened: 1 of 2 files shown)


edzer:
This is good to go, right?

ateucher:
It is, though I haven't quite figured out the error on `R-CMD-check-with-db`. `remotes::install_deps` is segfaulting... But that check was failing already so at least it's not a step backwards? I'll look into it more though.

ateucher:
The error on the docker based workflow may be due to a `libcurl` mismatch, so based on advice from `sysreqs` author, I'm trying to install it via a different method 🤞 . If this build doesn't work I'd suggest merging this and I can have a go at this separately...

ateucher:
It worked! I think this is good to go now

edzer:
Great - thanks so much!!

ateucher:
My pleasure!

PR #1436, opened 2020-06-26
Update README.md

manolo20:
Added instructions to the Rmd on how to install sf in Ubuntu 20.04 LTS


edzer:
I find it hard to believe that all that is needed only to install this packages.

manolo20:
Before installing the packages one needs to modify /usr/lib/ssl/openssl.cnf (that's how I got it to work).

edzer:
I'm not convinced these instructions are special for sf.

PR #1439, opened 2020-07-04
add aliases to ?sfc for sfc_MULTIPOLYGON, etc

MichaelChirico:
Was trying to find the man page and tried `?sfc_MULTIPOLYGON`, it would be nice to land on `?sfc` from this (to me) natural attempt


edzer:
Makes sense - thanks!

Issue #1453, opened 2020-07-21
Documentation/examples for st_snap?


MichaelChirico:
I found myself digging and searching for a long time to find a good example of `st_snap`.  Eventually I had to find the `sf` NEWS to see a link to postgis which has some decent examples.

We should add at least one example to `sf` directly.

I also find the `tolerance` argument under-described. How exactly is `tolerance` applied? Is `st_buffer` applied to `x` and then `x` is "snapped" to `y` only if `y` is entirely within that buffered polygon?

PR #1454, opened 2020-07-21
add an example for st_snap

MichaelChirico:
Part of #1453 

As I don't know the details for `tolerance` it would help if one of the maintainers could please add that part to this PR & then close the issue


edzer:
Thanks - I did this now manually.

PR #1458, opened 2020-07-29
correct a small mistake in the document

microly:
correct a small mistake in the document of "geom-transformers.R"


edzer:
Thanks!

PR #1466, opened 2020-08-10
Add as_tibble argument to st_as_af() and st_sf()

charliejhadley:
The `as_tibble` argument of `read_sf()` is a great feature, but this functionality is not currently provided in `st_as_sf()` or `st_sf()`. This is particularly desirable when converting `{sp}` objects to `{sf}` objects.

By default `as_tibble = FALSE` for both functions to prevent unexpected code changes for end-users.

I feel it is useful for the package to provide a nice way to add the "tbl_df" class to an `{sf}` object, similarly to how `st_drop_geometry()` was added for throwing away the geometry column. But this approach might not fit with the package owner's goals.

### DIFF
--- a/R/sf.R
+++ b/R/sf.R
@@ -34,6 +34,9 @@ st_as_sf = function(x, ...) UseMethod("st_as_sf")
 #' meuse_sf = st_as_sf(meuse, coords = c("x", "y"), crs = 28992, agr = "constant")
 #' meuse_sf[1:3,]
 #' summary(meuse_sf)
+#' # Use as_tibble = TRUE to return a tibble instead of a data.frame
+#' muse_tib = st_as_sf(meuse, coords = c("x", "y"), crs = 28992, agr = "constant", as_tibble = TRUE)
+#' muse_tib[1:3,]
 #' @export
 st_as_sf.data.frame = function(x, ..., agr = NA_agr_, coords, wkt,
 		dim = "XYZ", remove = TRUE, na.fail = TRUE, sf_column_name = NULL, as_tibble = FALSE) {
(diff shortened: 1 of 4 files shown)


R/sf.R
```diff
@@ -33,9 +34,12 @@ st_as_sf = function(x, ...) UseMethod("st_as_sf")
 #' meuse_sf = st_as_sf(meuse, coords = c("x", "y"), crs = 28992, agr = "constant")
 #' meuse_sf[1:3,]
 #' summary(meuse_sf)
+#' # Use as_tibble = TRUE to return a tibble instead of a data.frame
+#' muse_tib = st_as_sf(meuse, coords = c("x", "y"), crs = 28992, agr = "constant", as_tibble = TRUE)
```
edzer, line 38: please use `meuse_tib` here, and below

charliejhadley: done - thanks!

edzer:
Thanks, yes I'm looking at this. In cocrete, my question is whether `st_as_sf` should take an `as_tibble` argument: I would like it should make an sf-data.frame when given a data.frame, and an sf-tibble when given a tibble, and I wanted to run some tests how it now behaves but haven't gotten around it (distracted ATM).

edzer:
So, indeed we have
```r
library(sf)
# Linking to GEOS 3.9.0, GDAL 3.2.0, PROJ 7.2.0
library(tibble)
a = data.frame(a=1)
a$geom = st_sfc(st_point(1:2))
a %>% st_as_sf() %>% class()
# [1] "sf"         "data.frame"
as_tibble(a) %>% st_as_sf() %>% class()
# [1] "sf"         "tbl_df"     "tbl"        "data.frame"
```
and for that reason I don't think that `st_as_sf()` should have an `as_tibble` argument. Users should use `as_tibble()` and `as.data.frame()` to move between one and the other, `st_as_sf()` shouldn't change it's input's character.

charliejhadley:
Thanks for looking through this @edzer. This PR failed to actually achieve what I wanted:

- Generate an "sf" "tbl_df" for Spatial objects

This was because I didn't understand S3 classes. To help simplify matters I've created a new PR https://github.com/r-spatial/sf/pull/1582 which does achieve this. I started a new PR from scratch because this one made several faulty assumptions.

I've closed this PR in favour of https://github.com/r-spatial/sf/pull/1582, but if you'd prefer I work in a different way please let me know.

PR #1470, opened 2020-08-18
Update sf7 vignette, more examples, add benchmarks, evaluate

Robinlovelace:


Robinlovelace:
Heads-up @edzer this is a remote and indirect contribution to OpenGeoHub, from the hills in north Wales. Not sure why it's failing on most set-ups but works on my laptop. I think the benchmark conveys the message: think about resolution!

edzer:
Thanks a lot Robin! Yes, the tests are very noisy ATM and need maintenance.

Robinlovelace:
Any feedback welcome

edzer:
Great work, thanks. I'll try to extend it wrt

* s2 buffers being always cell boundaries
* using `sf_within_distance` rather than computing a buffer, then intersecting with the buffer

PR #1475, opened 2020-08-27
Use GitHub Actions via tic

pat-s:
Per request in https://github.com/r-spatial/mapview/issues/314#issuecomment-676263358

{tic} builds a cache daily during the night in a CRON build and you should have instant builds for manual triggers whenever you do the first commit of a day.

The windows build still fails due to some permission issues of `cleanup` and `configure`.
This is unrelated to the general approach and should be solved afterwards.

There are also some NOTE's in the rcmdcheck log, in case you haven't seen them yet.

Build previews: https://github.com/pat-s/sf/actions

- Use GitHub Actions instead of Travis
- The pkgdown site is auto-deployed to the `gh-pages` branch. You need to call `tic::use_ghactions_deploy()` once from the repo root.
- {tic} now supports automatic updating of the YAML files. If you want to opt in:
  1. `tic::use_update_tic()`
  1. Create a GH PAT with workflow scopes
  1. `tic::gha_add_secret(<secret>, "TIC_UPDATE")`
  This will run once a day in a CRON job and open an PR in your repo if there are updates to the tic template being used here.


edzer:
Thanks! Could you please take a look at the breaking windows test? If I reproduce this locally, it doesn't show up - I'm pretty sure the file `configure.ac` contains no CR or CR/LF.

pat-s:
Tried to set LF explicitly but did not have an effect. I am out of ideas on this one.

edzer:
OK; merging this, as it still looks much better than what we have in place.

edzer:
I've now looked at this several hours, but seem to be completely stuck in authorization problems, both for "update tic" and the pkgdown deploy step.

edzer:
... and windows still has the same problem, a problem that win-builder and CRAN do not have.

pat-s:
Sorry to hear you're facing that many issues with {tic}.
FWIW many other repos are working just fine since a long time (including updates).
Maybe mu suggestions below can get you back on track easily.

## tic

[This message](https://github.com/r-spatial/sf/runs/1673790015?check_suite_focus=true#step:19:245) confirms the authentication issues.

I don't know what you did but I suggest the following to solve it

1. Delete the `TIC_DEPLOY_KEY` secret
2. Run `tic::use_ghactions_deploy()`


I can't help with the Windows issues right now: {tic} does not do anything special here, it just runs the windows runner and calls `rcmdcheck::rcmdcheck()`.

You might want to disable the windows runner and test on winbuilder manually for releases.

I also saw that you manually added `libgit2-dev` [here](https://github.com/r-spatial/sf/commit/60118f15a58b05adf657d17347a4fcd64d723eff#diff-79567d0c6f34621b98467ea7b40a018302e361611bef643f7926a3b823e6df88).
This won't survive tics automatic updating but shouldn't be a problem since it would come with the latest template anyway.

## update-tic

Looking at the logs [authentication fails](https://github.com/r-spatial/sf/runs/1673709539?check_suite_focus=true#step:6:33).

I suggest 

1. Calling `tic::use_update_tic()` again (to get the latest workflow file)
2. Adding a GitHub PAT with `workflow` scopes as a secret again (named `TIC_UPDATE`) via `tic::gha_add_secret()`

edzer:
Thanks; I created a new key called `TIC_DEPLOY_KEY` with repo/user/workflow permissions, and then tried
```r
> tic::use_ghactions_deploy()
ℹ Querying GitHub deploy keys from repo.
→ Deploy keys for GitHub Actions already present. No action required.
```
which doesn't seem successful. How to I remove the secret first?

pat-s:
In the repo settings go to `Secrets` and delete the secret(s). 

> Thanks; I created a new key called TIC_DEPLOY_KEY with repo/user/workflow permissions,

This one should be named `TIC_UPDATE` and is only needed for the `update-tic` action -> `tic::gha_add_secret("TIC_UPDATE", "<token>")`

`TIC_DEPLOY_KEY` is created by `tic::use_ghactions_deploy()` automatically and does not need workflow scopes.

pat-s:
From a dev perspective it would be interesting to know how you got into the current state since the keys and everything else worked at some point in the past -> to better understand possible traps for users that one could fall over).

edzer:
Thanks; deploy works again for sf! 

I don't know why this stopped working; what I did miss is a simple minimal set of instructions how to create a key (personal settings) and then add it to a repo (repo setting). Maybe assume that tic users don't (want to) know anything about GH's secrets management?

edzer:
@pat-s tic deploy seems to hang again for this repo: https://github.com/r-spatial/sf/runs/2938453266?check_suite_focus=true#step:19:419 - things seem to work OK e.g. for the stars repo. Any idea?

pat-s:
> @pat-s tic deploy seems to hang again for this repo: r-spatial/sf/runs/2938453266?check_suite_focus=true#step:19:419 - things seem to work OK e.g. for the stars repo. Any idea?

Looks strange and similar to random race-condition patterns.

I would try to merge #1571 first and see if it resolves things.

#1571 is the automatically triggered update mechanism by tic but I see you haven't merged updates since Jan 10th. Is there a specific reason you did not merge the template updates so far/since then? (Just to gain some understanding)

But note that doing so will re-enable windows and macOS runners again, which I see you commented out in `tic.yml` manually.
`tic` takes most info out if its header, which is still configured as `linux-macos-windows-deploy` and would need to be changed if the runners should be ignored permanently.

edzer:
> #1571 is the automatically triggered update mechanism by tic but I see you haven't merged updates since Jan 10th. Is there a specific reason you did not merge the template updates so far/since then? (Just to gain some understanding)

I hadn't noticed them ;-)

PR #1479, opened 2020-09-01
Found two more GDAL/PROJ version vulnerabilities

rsbivand:
In addition to the single problem Joe hit, I've found two more (in `crs.R` and `sp.R`). In `sp.R`, I also push the WKT2 out through `sp::CRS()` to try to tackle the errant axis order issue (**sp** is only ever visualization order, but written files were taking the WKT2 verbatim, letting other software think that the data were stored lat/lon when they were lon/lat). I'll update **rgdal** and **sp** to use PROJ `proj_normalize_for_visualization()` to modify the WKT2 string, controlled by an option in **rgdal**. I'll say when **sp** is ready for release (after **rgdal** this time, fairly soon).

Please choose only the two recent commits, I don't know how to trim those before them.


R/crs.R
```diff
@@ -164,7 +164,9 @@ st_crs.default = function(x, ...) NA_crs_
 make_crs = function(x) {
 
 	if (inherits(x, "CRS")) {
-		x = if (!is.null(comment(x)))
+		x = if (is.null(comment(x)) || (CPL_proj_version() < "6.0.0" ||
```
edzer, line 167: If `comment(x)` is NULL, you assign NULL to x, which breaks on line 177.

R/sp.R
```diff
@@ -330,7 +330,10 @@ CRS_from_crs = function(from) {
 			sp::CRS(NA_character_)
 		else
 			sp::CRS(from$proj4string)
-	if (!is.null(from$wkt) && !is.na(from$wkt))
-		comment(ret) = from$wkt
+        if (CPL_proj_version() < "6.0.0" || 
+                    CPL_gdal_version() < "3.0.0") {
+	    if (!is.null(from$wkt) && !is.na(from$wkt))
+		comment(ret) = sp::CRS(SRS_string=from$wkt)
```
edzer, line 336: It seems you assign an object of class `CRS` to a comment; I always thought comments had to be text.

R/crs.R
rsbivand: OK, I'll reverse them.

R/sp.R
rsbivand: Added comment. There was a logic error in the condition too. I've committed & pushed, as well as expanding the crs/CRS coercion section in tests/crs.R.

tests/crs.R
```diff
@@ -41,7 +41,15 @@ if (sf_extSoftVersion()["USE_PROJ_H"] == "true" || sf_proj_info("have_datum_file
   "datum files not installed"
 }
 
-x = st_crs(sp::CRS("+proj=longlat +ellps=WGS84 +no_defs"))
+x0 = sp::CRS(SRS_string="EPSG:4326")
```
edzer, line 44: This fails when the GDAL/PROJ versions are too old (see CI: ubuntu 16:04 release). The breaking vignettes on old-rel are also worrying, just not informative where things go wrong.

rsbivand: CI now tells me a very high noise/signal ratio. I cannot see the real R CMD check output anywhere. I've reverted to `"+init=epsg:4326"`.

edzer:
So, if I understand correctly, when you specify `EPSG:4326` the wkt in sp's CRS `comment()` specifies axis order lng [1] lat [2]. I can see that is in agreement with how we treat the data, and how we would write it, but isn't it in disagreement with `EPSG:4326`? In
```
GEOGCRS["WGS 84",
    DATUM["World Geodetic System 1984",
        ELLIPSOID["WGS 84",6378137,298.257223563,
            LENGTHUNIT["metre",1]]],
    PRIMEM["Greenwich",0,
        ANGLEUNIT["degree",0.0174532925199433]],
    CS[ellipsoidal,2],
        AXIS["geodetic latitude (Lat)",north,
            ORDER[1],
            ANGLEUNIT["degree",0.0174532925199433]],
        AXIS["geodetic longitude (Lon)",east,
            ORDER[2],
            ANGLEUNIT["degree",0.0174532925199433]],
    USAGE[
        SCOPE["unknown"],
        AREA["World"],
        BBOX[-90,-180,90,180]],
    ID["EPSG",4326]]
```
the `BBOX` element is still lat lng lat lng. Is there some kind of best practice, or recommendations, that you followed to get here? Or some place (outside R) where this practice is being discussed? Just wondering whether `sf` should do this too.

rsbivand:
I'm uncertain where any such discussions are occurring. I haven't spoken to  Markus about GRASS since talking in your garden a year ago. I did look at the pyproj code in November, but probably everyone is waiting for everyone else. I think that for legacy/**sp**, keeping to GIS/vis order makes sense, because that was the **sp** design choice. The problem I first spotted was that WKT2 generated in **sf** were being copied into `"CRS"` comments verbatim, so my current attempt is to make sure they are GIS/vis. I noticed such a verbatim WKT2 written to a GPKG, which was wrong - the data were GIS/vis, but the WKT2 said verbatim 4326. Another application probably would assume that the WKT2 was correct, and misplace the axes.

Maybe the BBOX is a PROJ issue, I'm using the PROJ function (latest rgdal rev.) to enforce GIS/vis if `rgdal::get_enforce_xy()`. I'll check.

PR #1482, opened 2020-09-14
Change default precision to 1e8

edzer:
Up for discussion: this PR changes the default precision value used by `sf` from 0.0 (don't do any rounding) to 1e8 (round as `round(x * 1e8)/1e8` ). This means (as far as I understand) roughly to rounding to the nearest cm for geographic coordinates (if you're close to the equator), or to the nearest 10 nanometer for coordinates with unit meters (most projected coordinates).

I believe this corresponds to what `rgeos` does, and it is time that `sf` takes a (non-zero) default too.


edzer:
There is now a global variable, which can be retrieved by `st_precision()` and changed (for upcoming use, not for already created objects) e.g. by `st_set_precision(1e10)`.

tim-salabim:
My go-to reference for degree vs. length precision conversion is 
https://en.wikipedia.org/wiki/Decimal_degrees

Seems with 8 digits, you're rounding to the nearest mm (at the equator)

edzer:
> Seems with 8 digits, you're rounding to the nearest mm (at the equator)
Of course, you're right!

dbaston:
What problem is the automatic rounding intended to solve? I agree that the removed precision is almost certainly meaningless, but there is value in having data pass through `sf` unchanged. Rounding could create confusion in other cases, such as polygons seen as "valid" by PostGIS / shapely / etc. but invalid by `sf`. Invalidities introduced through rounding could go on to cause incorrect results or exceptions from GEOS. In my experience, users are much more likely to blame this on the tool they're using instead of the data they're working with.

edzer:
@dbaston do you have a (real) example of an invalidity introduced by rounding? The only things I can think of are mm / nm sized things.

dbaston:
@edzer Only dataset I have on this laptop is GADM country boundaries. They're all valid at a tolerance of 1e-8, but Rwanda is invalid at a tolerance of `1.2e-8`:
```
SELECT ST_IsValid(geom), ST_IsValid(ST_SnapToGrid(geom, 1.2e-8)) FROM gadm36_0 WHERE GID_0 = 'RWA'; 
NOTICE:  Self-intersection at or near point 28.927768703999998 -2.3901689039999998
 st_isvalid | st_isvalid 
------------+------------
 t          | f
```

edzer:
Thanks @dbaston , point taken. I think the reason to do this is that doing geometrical operations with doubles in general is not robust, and the error messages users get are hard to catch and don't point them in the direction of solutions. That, with the fact that rgeos has done this for maybe 10 years now, although I don't know whether @rsbivand or @rundel would do the same thing again, now?

Would the OverlayNG activity in GEOS be of influence on this discussion?

dbaston:
If the goal is to improve robustness, you could consider rounding on the fly if an overlay operation fails. This has the advantage of keeping lossless operations like `st_read(x) %>% filter(abc == 'def') %>% st_write(y)`.

And yes, the OverlayNG work in GEOS should definitely improve the story here. Unfortunately I haven't had much (any) time to work with it yet. If I understand right, it will not round by default, but could round on error (possibly to 1e-12: https://github.com/libgeos/geos/blob/master/include/geos/operation/overlayng/OverlayNGSnapIfNeeded.h#L71, but again, I haven't really worked with it yet)

edzer:
I just learned from https://github.com/r-spatial/s2/issues/78#issuecomment-693641625 that `GEOSGeom_setPrecision_r` has a flag to preserve topology while rounding a complete geometry.

dbaston:
It does, but under the hood the "topology preservation" is just a potentially destructive `buffer(0)`.

PR #1514, opened 2020-10-18
Add zip directory support to st_read

arthurgailes:
A quick edit to allow st_read to process zip files by extracting them into the R temp directory. It's common practice in particular to zip ESRI shapefiles, and saves directory space, so I suspect many users would find this useful.


edzer:
Still fails on:
```
  ── 1. Error: we can read a zipped shapefile using st_read (@test_zip.R#5)  ─────
  Cannot open "/tmp/RtmpL0YZMl/shp"; The file doesn't seem to exist.
  Backtrace:
   1. sf::st_read("inst/zip/nc.zip", "nc", crs = 4267, quiet = TRUE)
   2. sf:::st_read.character("inst/zip/nc.zip", "nc", crs = 4267, quiet = TRUE)
   3. sf:::CPL_read_ogr(...)
  
  ══ testthat results  ═══════════════════════════════════════════════════════════
  [ OK: 735 | SKIPPED: 60 | WARNINGS: 13 | FAILED: 1 ]
  1. Error: we can read a zipped shapefile using st_read (@test_zip.R#5) 
```

arthurgailes:
Fixed the R checks, I don't think I broke the AppVeyor test.

R/read.R
```diff
@@ -242,7 +242,7 @@ st_read.character = function(dsn, layer, ..., query = NA, options = NULL, quiet
 proc_zip <- function(dsn){
 	# create unique directory so it can be deleted
 	tmp <- paste0(tempdir(),"/shp")
```
edzer, line 244: I think that using `tempfile()` here would be safer than `paste0(tempdir(), "/shp")`, since the latter is constant for the whole R session, and may already exist and contain data. `exdir = ...` creates the directory if it doesn't already exist.

arthurgailes: Cool! Updated the file.

dbaston:
`sf` already supports reading zipped data via the `/vsizip` prefix...maybe just update `st_read` to tack on `/vsizip` automatically before handing the DSN over to GDAL?

```
> st_read('/vsizip/NUTS_BN_01M_2021_4326.shp.zip')
Reading layer `NUTS_BN_01M_2021_4326' from data source `/vsizip/NUTS_BN_01M_2021_4326.shp.zip' using driver `ESRI Shapefile'
Simple feature collection with 20091 features and 8 fields
geometry type:  LINESTRING
dimension:      XY
bbox:           xmin: -63.15119 ymin: -21.38885 xmax: 55.83578 ymax: 80.83402
geographic CRS: WGS 84
```

Issue #794, opened 2018-07-13
TopologyException in multipolygons (intersection function)


camolina10:
Hi,
I have a large sfc (MULTIPOLYGON) objects which throw several TopologyException errors when I try to use the intersection function on them.  I have a couple of questions so, If anyone has a moment, I would really appreciate it if you could take a look at this. I've attached the files in RDS format ([all](https://www.dropbox.com/s/n7p3yco5rkq1l9v/all.rds?dl=0)). Below is some example code.

`all<-readr::read_rds("all.rds")`

None of these lines seems to work:

```
all %>% st_intersection
all %>% st_set_precision(1000000) %>% lwgeom::st_make_valid %>% st_intersection
all %>% st_set_precision(1000) %>% lwgeom::st_make_valid %>% st_intersection

```

1. `sf::st_is_valid` returns `TRUE` for all objects. I've tried using `mapview`, `lwgeom::st_make_valid` and modifying the precision of the calculations. Only when I set precision to 100 or lower, the "rounded coordinates" become valid geometries to execute the `st_intersection` function. This issue is related to discussions in [#600](https://github.com/r-spatial/sf/issues/600) and [#603](https://github.com/r-spatial/sf/issues/603). The problem for me is that rounding the coordinates (two decimal places) does not seem appropriate for the work I'm doing. I tried several strategies including rounding (and using st_make_valid) the coordinates just for next rows (sfg):
`all %>% st_set_precision(100) %>% lwgeom::st_make_valid %>% sf::st_is_valid`
However, the issue remains. 

2.  A subset of the data that generates one of the TopologyException errors is:

```
subset<-all %>% subset(id %in% c(47,98,158))
st_intersection(subset)
```
I imagine I am off-track here but It seems tricky to me that the st_intersection of these three multipolygons do not work but pairwise-intersections of these rows do not generate any error. Could you please let me know what I am missing?
```
int12<-st_intersection(subset[c(1,2),])
int13<-st_intersection(subset[c(1,3),])
int23<-st_intersection(subset[c(2,3),])
```

Apologies if I've posted this in the wrong place. I hope I am not completely off-track. In case, feel free to delete this.

dbaston:
If you can output the offending geometries in WKB format, you could submit it as a bug report to [GEOS](https://trac.osgeo.org/geos/).

edzer:
Read the documentation section starting with "when called with a missing ‘y’,", and consider calling
```r
out <- st_intersection(all, all)
```
[...]

PR #1516, opened 2020-10-20
Detect overlay-ng

rsbivand:
I suggest using the overlay-ng output to complete the test, and re-using that case as a prototype for testing whether we are overlay-ng or not. I haven't exposed the prototype, but did edit DESCRIPTION by  hand (Collate:). `sf:::is_overlayng()` is `TRUE` where I think it should be, and `FALSE` on Windows, which is also correct.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -153,3 +153,4 @@ Collate:
     'shift_longitude.R'
     'make_grid.R'
     's2.R'
+    'geos-overlayng.R'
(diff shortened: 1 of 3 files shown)


edzer:
I still don't see why we'd need this (and who would need this): doesn't testing on the GEOS version tell you whether overlayNG is used?

rsbivand:
No, not at all. The configure flag must be set and can be unset. I've asked Martin Davis if I can CC you on fallout from #794 but haven't heard back. It may be that JTS is OK but GEOS not, or that my 13.7GB WKT test file is not a good representation of `all.rds`. We need a runtime test, and if GEOS will not give us one, we need to be defensive, I think.

rsbivand:
Any progress on this?

PR #1520, opened 2020-10-28
revise crs_a == crs_b, in the light of axis order

edzer:


edzer:
See https://twitter.com/edzerpebesma/status/1321405831531945985 - this gets us:
```r
> st_axis_order(FALSE) # default: non-authority compliant
> st_crs("OGC:CRS84") == st_crs("OGC:WGS84")
[1] TRUE
> st_axis_order(TRUE)
> st_crs("OGC:CRS84") == st_crs("OGC:WGS84")
[1] FALSE
```
but I'm not sure this is a good idea. For instance:
```r
> library(sf)
Linking to GEOS 3.8.1, GDAL 3.1.3, PROJ 7.1.1
> st_axis_order(TRUE)
> st_sfc(st_point(0:1), crs = 'OGC:WGS84') %>% st_transform('OGC:CRS84') # doesn't swap axes:
Geometry set for 1 feature 
geometry type:  POINT
dimension:      XY
bbox:           xmin: 0 ymin: 1 xmax: 0 ymax: 1
geographic CRS: WGS 84
POINT (0 1)
```

rsbivand:
I'm not sure. I think that == is in PROJ: `proj_is_equivalent_to(..., PJ_COMP_STRICT)` rather than `PJ_COMP_EQUIVALENT` and is not `PJ_COMP_EQUIVALENT_EXCEPT_AXIS_ORDER_GEOGCRS`. `rgdal::compare_CRS()` reports all three, but that is not good for a logical operator (maybe with `all()`). Details at https://proj.org/development/reference/datatypes.html#_CPPv423PJ_COMPARISON_CRITERION.

edzer:
Thanks for that link. Given that EQUIVALENT also means e.g. only different units, we should go for STRICT.

PR #1544, opened 2020-11-20
Update README.md

prdm0:
yay today is the most used program in Arch Linux and derived distributions for installing community packages Arch. For example, yaourt has been discontinued.


edzer:
Thanks!

PR #1545, opened 2020-11-24
Add is_coverage argument to summarize method

dbaston:
This is passed through to st_union and allows for better performance in
cases where the inputs are known to have no overlaps.


edzer:
Thanks, makes sense!

Issue #1559, opened 2020-12-14
What should st_as_sfc() do with a SpatialLines object with one coordinate?


rsbivand:
This comes from coercing the output of gdistance::shortestPath(..., output="SpatialLines") to `sfc` when no shortest path is found @kauedesousa. The **sp** code did once briefly check for > 1 coordinate but was commented out in 2007 because of **maps** returning single point lines https://r-forge.r-project.org/scm/viewvc.php/pkg/sp/R/Class-SpatialLines.R?root=rspatial&r1=563&r2=593.

```
library(sp)
crds <- matrix(c(1,1), ncol=2)
ll <- SpatialLines(list(Lines(list(Line(crds)), ID="1")))
library(sf)
st_make_valid((st_as_sfc(ll)))
# Error in CPL_geos_make_valid(x) : 
#   Evaluation error: IllegalArgumentException: point array must contain 0 or >1 elements.
```
It would be possible to edit `st_as_sfc.SpatialLines()` either to duplicate the coordinate (probably not a good idea) or to return an empty linestring on both branches of linestring or multilinestring if `nrow(.@coords) == 1L`. It would also be possible to change **sp** by reintroducing the `nrow > 1` condition, or `shortestPath()`, but the latter is not logical if **sp** accepts `nrow == 1L` as valid.

edzer:
I think the proper way would be to create an empty linestring.

rsbivand:
```
library(sp)
crds <- matrix(c(1,1), ncol=2)
ll <- SpatialLines(list(Lines(list(Line(crds)), ID="1")))
library(sf)
st_make_valid(st_as_sfc(ll))
st_make_valid(st_as_sfc(ll, forceMulti = TRUE))
crds1 <- matrix(c(1,1,2,2), ncol=2, byrow=TRUE)
ll1 <- SpatialLines(list(Lines(list(Line(crds), Line(crds1)), ID="1")))
st_make_valid(st_as_sfc(ll1))
st_make_valid(st_as_sfc(ll1, forceMulti = TRUE))
ll0 <- SpatialLines(list(Lines(list(Line(crds1)), ID="1")))
st_make_valid(st_as_sfc(ll0))
st_make_valid(st_as_sfc(ll0, forceMulti = TRUE))
```
These now pass with commit to rsbivand/sf - will PR shortly. Case not added to tests.

edzer:
Closed by https://github.com/r-spatial/sf/pull/1560

PR #1560, opened 2020-12-15
handle coercion to sf of degenerate sp single coord. lines #1559

rsbivand:
This works for the simple test case in the issue.


edzer:
Thanks!

PR #1585, opened 2021-01-27
Allowing class unions by exporting sf class

### LINKED ISSUE #1584: Exporting class "sf" to allow class unions
Hello, 
I am writing a package in which I find myself needing to set a union between `data.frame` and `sf` as such:
```r
#' @import sf
setOldClass(c("data.frame", "sf"))
setClassUnion("spatial.data.frame", c("data.frame", "sf"))
```
Yet doing so I get a the following warning when I `devtools::document` or when I build the package:
```r
Warning: class "sf" is defined (with package slot ‘sf’) 
but no metadata object found to revise superClass information---not imported?  
Making a copy in package ‘spaspm’
```
I think this is due to the fact that the class `sf` is not exported by the `sf` package. 

Therefore my question is: would you consider exporting the `sf` class to allow for class unions?

VLucet:
Exports the `sf` class by adding the appropriate `roxygen` tag. 

Fixes #1584


VLucet:
Tests pass, but we'd need to document the class to avoid warnings.

PR #1588, opened 2021-01-31
Remove wk methods

paleolimbot:
Very cool that there is an `wk::as_wkb()` method in sf! I'm taking the approach in wk that I should provide full sf support so that I can fix all the bugs as they pop up! I believe the latest CRAN version has an `as_wkb()` method for sfc, and as far as I can tell, you don't use `as_wkb()` internally or in s2, so I think it is safe to remove these.


edzer:
I'm generally ok with this, but there are still some differences between `sf:::as_wkb.sfc` and `wk:::as_wkb.sfc` (besides docs errors) that need to be resolved. For instance, `sf`'s version transforms non-geographic coordinate data to geographic.

edzer:
>  although I'd prefer if it didn't convert to lon/lat first.

From my perspective, this is about using s2 as a geometry back-end instead of GEOS. From your perspective, this is about making wk a clean back-end, irrespective what happens next. If not here, where would you put the (error / warning / auto-conversion to lng/lat) when someone tries to use s2 routines on projected data?

edzer:
Thanks a lot!

PR #1596, opened 2021-02-05
Update to new spatstat structure.

rubak:
This should work now, but I'm not sure whether it is the correct time for integrating into `sf`, so maybe we should have a quick discussion before you actually accept this PR. When is the next CRAN submission planned for `sf`? I can't figure out whether this PR should wait until the new umbrella spatstat is on CRAN or at least very close to. If you have a combination of old spatstat and the new sub-packages you can get unwanted behaviour as described [in this issue](https://github.com/spatstat/spatstat.geom/issues/1).

Can you squash the 4 commits when you accept or do I need to do it and push a squashed version?


edzer:
Thanks! No need to squash AFAICT. What is the risk merging this? It passes tests with the CRAN packages, what do I overlook?

rubak:
Assume user with old spatstat 1.64-1 (current CRAN) installed along side spatstat.geom.

``` r
library(sf)
#> Linking to GEOS 3.8.0, GDAL 3.0.4, PROJ 6.3.1
triangle <- st_sfc(st_polygon(list(cbind(c(1,1,2,1),c(1,2,2,1)))))
```

With current CRAN version of `sf` this would do the right thing, but with the
suggested changes it returns the bounding box:

``` r
W_spst <- spatstat::as.owin(triangle)
plot(W_spst)
```



With current CRAN version of `sf` this would return the bounding box. With the
suggested changes it returns the correct triangle:

``` r
W_spst.geom <- spatstat.geom::as.owin(triangle)
plot(W_spst.geom)
```



Once the user has the new umbrella spatstat the command `spatstat::as.owin()`
doesn’t exist and it will correctly fail when called with `spatstat::`. It works
when called after `library(spatstat)` because umbrella `spatstat` loads
`spatstat.geom` and friends via `Depends`.

The main problem is that we will have silent wrong behaviour, which is
always bad. So the question is whether we need more checking of spatstat
versions etc. in the sf code. The problematic cases arise when the user has
installed `spatstat.geom` but at the same time has `spatstat < 1.65`, so one
option is to detect this and ask the user to update `spatstat`. But that is
not possible before umbrella spatstat is on CRAN.

<sup>Created on 2021-02-06 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

edzer:
Thanks - all clear, yes then it is best to wait for umbrella `spatstat` to arrive on CRAN, and then put version requirements in `sf` before we merge this & submit to CRAN.

rsbivand:
@rubak do you have a timeline for the umbrella **spastat**? May I commit my local **maptools** adaptations to R-Forge - do you need that to run revdep checks?

rubak:
@edzer OK. Regarding the version requirement in `sf`: Would it be OK if I make a function in `R/spatstat.R` called `check_spatstat` and use that in all the functions that start with checking for flat coordinates and that spatstat.xxxx is installed? Something like:

```r
check_spatstat <- function(pkg){
  if(!requireNamespace(pkg, quietly = TRUE)){
    stop("package ", pkg, " required, please install it (or the full spatstat package) first")
  } else{
    spst_ver <- try(packageVersion("spatstat"), silent = TRUE)
    if(!inherits(spst_ver, "try-error") && spst_ver < 1.65){
      stop("You have an old version of spatstat installed which is incompatible with ", pkg, 
           ". Please update spatstat (or uninstall it).")
    }
  }
  if (isTRUE(st_is_longlat(X)))
    stop("Only projected coordinates may be converted to spatstat class objects")
}

```

rubak:
@rsbivand You can definitely go ahead and commit to R-Forge. I will test your new version out then. Do you need any input from me? Unless unexpected things happen Adrian expects to submit umbrella spatstat (spatstat 2.0-0) in a few days when we also have Windows binaries for spatstat.linnet.

rsbivand:
@rubak **maptools** 1.1-1  on R-Forge - please keep me updated on things that still need fixing when you run reverse dependency checks involving the coercion functions.

rubak:
@rsbivand I know of some people have had success with your dev version and I haven't heard about issues yet, so that is a good sign. Would you mind putting something like `check_spatstat()` in the spatstat conversion functions in `maptools` to make sure that people don't have old spatstat around? I can check out the dev version from svn and make the changes if you prefer?

@edzer are you OK with this type of check in `sf` as well? Then I will modify the pull request.

edzer:
Yes, that's OK! (The function still needs an argument `X`.)

rsbivand:
**maptools** rev. 385 now uses `check_spatstat()` but does not export it. Please let me know if this is as required.

rubak:
@rsbivand thanks for putting it in. You are right that `check_spatstat()` is not meant to be exported. Here in `sf` I have added a `Remotes: baddstats/spatstat` to `DESCRIPTION` to get spatstat 2.0 installed for checking. I don't know if that is an option with your R-forge workflow? Otherwise, I guess the failing test is not a big problem as spatstat 2.0 is going on CRAN this month (fingers crossed that it will be accepted without too long a delay).

rsbivand:
There is no `Remotes:` field in a valid DESCRIPTION file. Please advise if you think that R CMD check should not return:
```
Unknown, possibly mis-spelled, fields in DESCRIPTION:
  ‘Remotes’
```

rubak:
You are absolutely right @rsbivand. We use it for development versions of spatstat to help automatic scripts get the newest development versions of packages and then we delete the line before CRAN submission. I don't know if @edzer has a similar workflow with `sf` and other spatial packages? Otherwise I can also delete the line from DESCRIPTION in this pull request and then let the spatstat tests fail until spatstat 2.0 gets on CRAN.

Using this remotes specification is the easiest way I know of letting users relatively worry free install development versions that possibly use other development versions (well apart from the amazing https://r-universe.dev that Jeroen recently launched, which I hope we can adopt for the development workflow of spatstat in the future).

rsbivand:
I worry more about "worry free" approaches than having to avoid superfluous external dependencies (so never any RS innovations in my workflows, especially for development).

rubak:
I get your point @rsbivand and I also sometimes worry about these "automagic" things. We only started feeling the pain of needing users to install a development version of one package to get the development version of another when we first split out `spatstat.utils` from `spatstat` in the process of satisfying CRAN. It may be that since they are development versions it would be OK to ask a bit more of the end user, and give them instructions to install development versions of another package themselves rather than having this done automatically by a field in DESCRIPTION and a function in `remotes`. Anyway, with https://r-universe.dev I foresee that we can get rid of this usage of `remotes` once spatstat 2.0-0 is on CRAN and the split-up mess is complete. Maybe @edzer has a better way to get spatstat 2.0-0 installed in the `sf` check system here on GitHub so that this pull request turns "green" and seems safe for him to merge. I'm happy to adapt to which ever workflow is used if I can mange figure out the usage.

edzer:
Yes, I also use remotes often e.g. to get stars pass CI checks when it depends on an sf version not yet on CRAN.

rsbivand:
I've forked `baddstats/spatstat`, built it under R 4.0.3, and put it in a `drat` repo at "https://rsbivand.github.io/drat/". So `install.packages("spatstat", repos="https://rsbivand.github.io/drat/") seems to work. Had the package also been micro-versioned (2.0-0.100, etc), it would have been possible to tell which reference point in the fluid git repo had been used. Please feel free to point people at this drat repo rather than using remotes, or to fork the repo to baddstats, which is the logical place. Of course, `Additional_repositories:` is for the package itself rather than reverse dependencies, I think, am still waiting for R-Forge to rebuild rev. 389, but expect failure.

rubak:
Great. Thanks a lot @rsbivand. I have been meaning to have a look at Dirk's drat package but never got around to it. I was hoping that I would skip it all together in favor of the r-universe.dev service, but I will consider how to handle it.

rubak:
Btw. @rsbivand are you indicating this yields a simple way to automatically install development versions of a hard dependency? Are you allowed to have a package in `Imports`/`Depends` which is not on CRAN/Bioconductor but only available through your own repository? I have only seen this used for packages in `Suggests`. If this is allowed I can simply have `spatstat (>=2.0-0)` as `Imports`/`Depends` of a package and set `Additional_repositories` to a drat or r-universe.dev repository and then it would pass checks?

rsbivand:
@rubak I'm not sure how to interpret WRE or Dirk's comments. Might be worth trying?

rubak:
OK. Thanks for all your input @rsbivand. I will discuss with Adrian how to proceed from here.

rsbivand:
@rubak I ran spatial revdeps today (with PROJ 8.0.0 - the main reason). But I also installed my draft maptools and spatstat 2.0.0, and most of the observed failures were caused by CRAN packages (obviously) not ready. I attach a CSV - sf and maptools fail because they are the CRAN packages, draft maptools passes check fine. I apologise for the Excel file, but github doesn't like CSV:

[spatstat_revdeps.xlsx](https://github.com/r-spatial/sf/files/6018421/spatstat_revdeps.xlsx)

rubak:
Thanks for the file @rsbivand. There are indeed quite a few packages that require an update once spatstat 2.0 is on CRAN. We (Adrian) will submit to CRAN tomorrow (Monday). I hope both you and @edzer will submit new versions of `sf` and `maptools` soon thereafter. These are the main packages in terms of interconnections between packages that use `spatstat`, so they need to be in place before many others will pass checks.

edzer:
Great work. Thanks so much!

edzer:
(I mistakenly merged this before spatstat 2.0 was released on CRAN; is there an E.T.A. for spatstat 2.0?)

rubak:
We are stuck in the CRAN queue and don't know why. On Tuesday 23 Feb we
received a message about reverse dependencies failing and we confirmed
that this was expected and maintainers had been notified. Since then
there has been radio silence. We requested a status yesterday, but
still haven't heard back. We are sorry that this final step is taking
so long and hope that CRAN will proceed soon.

edzer:
OK, thanks!

edzer:
spatstat 2.0.1 now on CRAN - congratulations!

PR #1600, opened 2021-02-10
fix `delete_layer` default in write_sf.DBIObject

### LINKED ISSUE #1599: write_sf() Ignores delete_layer Argument
When using `write_sf()` I kept getting an error about `overwrite` being used multiple times.

```r
Error in .local(conn, name, value, ...) : 
  formal argument "overwrite" matched by multiple actual arguments
```

So I realized that the `overwrite` argument had been changed to `delete_layer`. But then I checked the source code for `write_sf()` and it appears to ignore what I am setting and forces `delete_layer=TRUE`.

```r
function (..., quiet = TRUE, append = FALSE, delete_layer = TRUE) 
{
    st_write(..., quiet = quiet, append = append, delete_layer = TRUE)
}
```

This occurs at [https://github.com/r-spatial/sf/blob/master/R/read.R#L502](https://github.com/r-spatial/sf/blob/master/R/read.R#L502).

I imagine 

```r
st_write(..., quiet = quiet, append = append, delete_layer = TRUE)
```

should be 

```r
st_write(..., quiet = quiet, append = append, delete_layer = delete_layer)
```

etiennebr:
- The default is to delete a table if it exists (`append = FALSE`)
- Using `append = TRUE` will set `delete_layer = FALSE` and append to an existing table or create one
- Using `append = TRUE` and `delete_layer  = TRUE` will raise an error using `delete_layer` rather than `overwrite`

fix #1599
related #1266


etiennebr:
I'll have a look at the failures later today

edzer:
regenerate docs.

etiennebr:
Looks good now.

edzer:
Thanks! Could you please double check this is still consistent with the current documentation?
```
#' @param append logical; should we append to an existing layer, or replace it?
#' if \code{TRUE} append, if \code{FALSE} replace; default \code{NA} will
#' raise an error if the layer exists. See also next two arguments.
#' @param delete_dsn logical; delete data source \code{dsn} before attempting
#' to write?
#' @param delete_layer logical; delete layer \code{layer} before attempting to
#' write?
```

etiennebr:
Good point Edzer. I'll see what I can do to use `NA` to preserve a consistent behavior, but it might be tricky with the database.

etiennebr:
I'll need to look at this more closely because I'm not sure what would make the most sense from a user perspective. In the `sf1.Rmd` vignette, it is mentioned that `write_sf` overwrites by default (which matches other `write_*` functions). 
https://github.com/r-spatial/sf/blob/61efbdaf163fdc38fcb092cef2c453322ac4feb5/vignettes/sf1.Rmd#L477-L485

However, what seems to be the agreement in https://github.com/r-spatial/sf/issues/1266#issuecomment-596201629 is:
> this commit renames `update` into `append`, deprecates update, raises an error if the layer exists but `append` has not been specified to either `TRUE` (do append) or `FALSE` (overwrite layer) .
> 
> I decided (for now) against a separate overwrite argument, as setting `append=FALSE` overwrites. In addition, we do have the `delete_dsn` and `delete_layer` logic of GDAL, which does different things to different drivers, which I think we want to keep.

Also As you mentioned: 
https://github.com/r-spatial/sf/blob/61efbdaf163fdc38fcb092cef2c453322ac4feb5/R/read.R#L329-L337

etiennebr:
Sorry for the hesitation. I would like your approval on this @edzer. 

I went back to original [`write_*()`](https://readr.tidyverse.org/reference/write_delim.html#arguments) functions form `readr`, because if I remember correctly this is what the inspiration was for the behaviour.

> __append__ If `FALSE`, will overwrite existing file. If `TRUE`, will append to existing file. In both cases, if the file does not exist a new file is created.

The default is `FALSE`, so it overwrites by default. 

If we decide to follow this behaviour for `write_sf` the strict implementation would be to set `delete_layer=!append`, but I think it would make sense to expose it in the arguments (as in a603d39f4dd21b05123bb080d63e3c6fbd27aa76). In that case, we could document the parameter as:

```r
#' @param append logical; should we append to an existing layer, or replace it? 
#' if \code{TRUE} append, if \code{FALSE} replace. 
#' The default for \code{st_write} is \code{NA} which raises an error if the layer exists. 
#' The default for \code{write_sf} is \code{FASE}, which overwrites by default.
#' See also next two arguments for more control on overwrite behavior. 
```

Meanwhile I'll go with this version for the PR so you can merge it, but let me know if you would prefer another behaviour.

etiennebr:
@edzer, let me know if there's anything missing to merge

edzer:
Thanks a lot, Etienne!

PR #1608, opened 2021-02-22
Fix gdal_polygonize() labelling of polygons and contour lines

JoshOBrien:
`gdal_polygonize()` "labels" the spatial features it returns in the
sf object's first column. That column is used to indicate the
contour (or contours, in the case of polygons) corresponding to
each spatial feature. Currently, the labels that gdal_polygonize is
assigning to each spatial feature aren't what would be hoped
for. The issues involved depend on whether one is using the
function to return polygons or the contour lines themselves. Here,
I'll briefly describe the two cases in turn.

- MULTIPOLYGONs (returned when setting `contour_lines = FALSE`)

  At least where I've used it (as in the example below), all of the
  resultant MULTIPOLYGON features are labeled as "NA". (The seems to
  happen because `process_cpl_read_ogr(CPL_polygonize(...))` is
  returning an object with a column full of "NA"s.

- LINESTRINGs (returned when setting `contour_lines = TRUE`)

  In this case there are several issues. First, the values that get
  returned are ranges (e.g. "[30,60]") instead of the scalar
  values (e.g. "30" or "60") that we'd expect as identifiers for a
  contour line.

  Worse yet, when the bottom two contour levels fall within the range
  of the raster's values, the results are unambigously incorrect. In
  one example below, I show that contours with the lowest two levels,
  at 30 and 60, both get labeled as having the same `"[30,60]"` level.

This PR proposes some modifications to `gdal_polygonize()` that at
least partially fix the situation. I'm fairly well satisfied with the
fix it applies to LINESTRINGs labeling. The fix for MULTIPOLYGONs
works fine in the example below but is maybe less robust, as it relies
on sampling of raster values within the polygons to identify the range
of values with which they are associated. A better fix may involve
modifications to CPL_polygonize or process_cpl_read_ogr, but I'm not
well-equipped to figure that out. Still, this first stab at a fix at
least demonstrates what I think would be better behavior.

I will attach images of the figures produced by the code below to a comment 
in the PR's thread.

```r
library(stars)
library(ggplot2)
library(ggspatial)

## Prepare an example raster layer
tif <- system.file("tif/L7_ETMs.tif", package = "stars")
x6 <- read_stars(tif)[,,,6]
r <- st_warp(x6, cellsize = c(4*28, -4*28),
             crs = st_crs(x6))
range(r[[1]])
## [1]   2 218

## Contour polygons
CP <- st_contour(r, breaks = c(0, 30, 60, 90, 300), contour_lines = FALSE)
names(CP)[[1]] <- "range"

## Contour lines, case 1: bottom contour value outside (below) data range
CL1 <- st_contour(r, breaks = c(0, 30, 60, 90, 300), contour_lines = TRUE)
names(CL1)[[1]] <- "contour"

## Contour lines, case 2: bottom contour values within data range
CL2 <- st_contour(r, breaks = c(30, 60, 90), contour_lines = TRUE)
names(CL2)[[1]] <- "contour"

##------------------------------------------------------------------------------
## Behavior without this PR
##------------------------------------------------------------------------------
CP
## Simple feature collection with 4 features and 1 field
## geometry type:  MULTIPOLYGON
## dimension:      XY
## bbox:           xmin: 288776.3 ymin: 9110681 xmax: 298744.3 ymax: 9120761
## projected CRS:  UTM Zone 25, Southern Hemisphere
##   range                      geometry
## 1  <NA> MULTIPOLYGON (((289840.3 91...
## 2  <NA> MULTIPOLYGON (((294432.3 91...
## 3  <NA> MULTIPOLYGON (((290512.3 91...
## 4  <NA> MULTIPOLYGON (((298744.3 91...
ggplot() + layer_spatial(CP, aes(fill = range))

## IDs for contours are ranges (e.g. "[0,30]") when they should be
## single values (e.g. "30").
CL1[4:6,]
## Simple feature collection with 6 features and 1 field
## geometry type:  LINESTRING
## dimension:      XY
## bbox:           xmin: 289827.8 ymin: 9120615 xmax: 297968.3 ymax: 9120761
## projected CRS:  UTM Zone 25, Southern Hemisphere
##   contour                       geometry
## 4  [0,30] LINESTRING (297814.7 912076... ## contour at 30
## 5 (30,60] LINESTRING (294374.3 912076... ## contour at 60
## 6 (60,90] LINESTRING (290456.3 912076...
ggplot() + layer_spatial(CL1, aes(col = contour))


## Even worse, when the bottom two contour levels fall within the
## range of the raster's values, the results are unambigously
## incorrect. Here, contours with the lowest levels, , at 30 and 60,
## both get labeled as having the same "[30,60]" level.
CL2[4:6,]
## Simple feature collection with 6 features and 1 field
## geometry type:  LINESTRING
## dimension:      XY
## bbox:           xmin: 289827.8 ymin: 9120615 xmax: 297968.3 ymax: 9120761
## projected CRS:  UTM Zone 25, Southern Hemisphere
##   contour                       geometry
## 4 [30,60] LINESTRING (297814.7 912076... ## contour at 30
## 5 [30,60] LINESTRING (294374.3 912076... ## contour at 60
## 6 (60,90] LINESTRING (290456.3 912076...
ggplot() + layer_spatial(CL2, aes(col = contour))


##------------------------------------------------------------------------------
## Behavior *with* this PR
##------------------------------------------------------------------------------

CP
## Simple feature collection with 4 features and 1 field
## geometry type:  MULTIPOLYGON
## dimension:      XY
## bbox:           xmin: 288776.3 ymin: 9110681 xmax: 298744.3 ymax: 9120761
## projected CRS:  UTM Zone 25, Southern Hemisphere
##      range                       geometry
## 1   [0,30] MULTIPOLYGON (((289840.3 91...
## 2  (30,60] MULTIPOLYGON (((294432.3 91...
## 3  (60,90] MULTIPOLYGON (((290512.3 91...
## 4 (90,300] MULTIPOLYGON (((298744.3 91...
ggplot() + layer_spatial(CP, aes(fill = range))

CL1[4:6,]
## Simple feature collection with 3 features and 1 field
## geometry type:  LINESTRING
## dimension:      XY
## bbox:           xmin: 290456.3 ymin: 9120615 xmax: 297968.3 ymax: 9120761
## projected CRS:  UTM Zone 25, Southern Hemisphere
##   contour                       geometry
## 4      30 LINESTRING (297814.7 912076... ## contour at 30
## 5      60 LINESTRING (294374.3 912076... ## contour at 60
## 6      90 LINESTRING (290456.3 912076...
ggplot() + layer_spatial(CL1, aes(col = contour))

CL2[4:6,]
## Simple feature collection with 3 features and 1 field
## geometry type:  LINESTRING
## dimension:      XY
## bbox:           xmin: 290456.3 ymin: 9120615 xmax: 297968.3 ymax: 9120761
## projected CRS:  UTM Zone 25, Southern Hemisphere
##   contour                       geometry
## 4      30 LINESTRING (297814.7 912076... ## contour at 30
## 5      60 LINESTRING (294374.3 912076... ## contour at 60
## 6      90 LINESTRING (290456.3 912076...
ggplot() + layer_spatial(CL2, aes(col = contour))
```

### DIFF
--- a/R/stars.R
+++ b/R/stars.R
@@ -12,7 +12,7 @@
 #' @export
 gdal_read = function(x, ..., options = character(0), driver = character(0), read_data = TRUE, NA_value = NA_real_,
 		RasterIO_parameters = list())
-	CPL_read_gdal(as.character(x), as.character(options), as.character(driver), 
+	CPL_read_gdal(as.character(x), as.character(options), as.character(driver),
 		as.logical(read_data), as.double(NA_value), RasterIO_parameters)
 
 #' @name gdal
@@ -20,7 +20,7 @@ gdal_read = function(x, ..., options = character(0), driver = character(0), read
 #' @param type gdal write type
 #' @param geotransform length 6 numeric vector with GDAL geotransform parameters.
 #' @param update logical; \code{TRUE} if in an existing raster file pixel values shall be updated.
-gdal_write = function(x, ..., file, driver = "GTiff", options = character(0), type = "Float32", 
+gdal_write = function(x, ..., file, driver = "GTiff", options = character(0), type = "Float32",
 		NA_value = NA_real_, geotransform, update = FALSE) {
 
 	if (!requireNamespace("stars", quietly = TRUE))
@@ -71,7 +71,7 @@ gdal_write = function(x, ..., file, driver = "GTiff", options = character(0), ty
 gdal_inv_geotransform = function(gt)
 	CPL_inv_geotransform(as.double(gt))
 
-## @param x two-column matrix with columns and rows, as understood by GDAL; 0.5 refers to the first cell's center; 
+## @param x two-column matrix with columns and rows, as understood by GDAL; 0.5 refers to the first cell's center;
 ## FIXME: this is now duplicate in sf and stars
 xy_from_colrow = function(x, geotransform, inverse = FALSE) {
 # http://www.gdal.org/classGDALDataset.html , search for geotransform:
@@ -84,7 +84,7 @@ xy_from_colrow = function(x, geotransform, inverse = FALSE) {
 			stop("geotransform not invertible") # nocov end
 	}
 	stopifnot(ncol(x) == 2)
-	matrix(geotransform[c(1, 4)], nrow(x), 2, byrow = TRUE) + 
+	matrix(geotransform[c(1, 4)], nrow(x), 2, byrow = TRUE) +
 		x %*% matrix(geotransform[c(2, 3, 5, 6)], nrow = 2, ncol = 2)
 }
 
@@ -135,7 +135,7 @@ st_as_sfc.dimensions = function(x, ..., as_points = NA, use_cpp = TRUE, which = 
 			c(x[1] - 0.5 * d[1], x + 0.5 * c(d, tail(d, 1)))
 		}
 		if (raster$curvilinear) { # expand jointly:
-			if (!as_points && all(dim(xd$values) == dim(x)[xy_names])) { # expand from points to cells/polygons: 
+			if (!as_points && all(dim(xd$values) == dim(x)[xy_names])) { # expand from points to cells/polygons:
 				xd$values = apply((apply(xd$values, 1, expand)), 1, expand)
 				yd$values = apply((apply(yd$values, 1, expand)), 1, expand)
 			}
@@ -161,7 +161,7 @@ st_as_sfc.dimensions = function(x, ..., as_points = NA, use_cpp = TRUE, which = 
 	}
 	dims = dim(x) + !as_points
 	if (use_cpp)
-		structure(CPL_xy2sfc(cc, as.integer(dims), as_points, as.integer(which)), 
+		structure(CPL_xy2sfc(cc, as.integer(dims), as_points, as.integer(which)),
 			crs = st_crs(xd$refsys), n_empty = 0L, bbox = bbox.Mtrx(cc))
 	else
 		st_sfc(xy2sfc(cc, dims, as_points), crs = xd$refsys)
@@ -240,7 +240,7 @@ gdal_subdatasets = function(file, options = character(0), name = TRUE) {
 #' @name gdal
 #' @export
 gdal_polygonize = function(x, mask = NULL, file = tempfile(), driver = "GTiff", use_integer = TRUE,
-		geotransform, breaks = classInt::classIntervals(na.omit(as.vector(x[[1]])))$brks, 
+		geotransform, breaks = classInt::classIntervals(na.omit(as.vector(x[[1]])))$brks,
 		use_contours = FALSE, contour_lines = FALSE, connect8 = FALSE, ...) {
 	gdal_write(x, file = file, driver = driver, geotransform = geotransform) # nocov start
 	on.exit(unlink(file))
@@ -270,12 +270,23 @@ gdal_polygonize = function(x, mask = NULL, file = tempfile(), driver = "GTiff", 
 	pol = CPL_polygonize(file, mask_name, "GTiff", "Memory", "foo", options, 0, contour_options, use_contours, use_integer)
 	out = process_cpl_read_ogr(pol, quiet = TRUE)
 	names(out)[1] = names(x)[1]
-	if (use_contours) {
-		m = as.integer(cut(out[[1]], breaks = nbreaks, include.lowest = TRUE)) # FIXME: add coverage when GDAL 2.4.0 is here
-		if (any(is.na(m)) && !all(is.na(m)))
-			warning("range of breaks does not cover range of cell values")
-		out[[1]] = structure(m, levels = levels(cut(breaks, breaks, include.lowest = TRUE)), class = "factor")
-	}
+        if (use_contours) {
+            if (!contour_lines) {
+                out[[1]] =
+                    sapply(seq_len(nrow(out)),
+                           FUN = function(i) {
+                               mean(st_extract(x, st_sample(out[i, ], size = 50))[[1]], na.rm = TRUE)
+                           })
+                m = as.integer(cut(out[[1]], breaks = nbreaks, include.lowest = TRUE))
+                if (any(is.na(m)) && !all(is.na(m)))
+                    warning("range of breaks does not cover range of cell values")
+                out[[1]] = structure(m, levels = levels(cut(breaks,
+                                                            breaks, include.lowest = TRUE)), class = "factor")
+            }
+            else {
+                out[[1]] = factor(out[[1]], levels = breaks)
+            }
+        }
 	out # nocov end
 }
 


JoshOBrien:
### Figures produced by the MRE above, with existing version of `gdal_polygonize()`

----
```r
ggplot() + layer_spatial(CP, aes(fill = range))
```

(Fails entirely, as polygon IDs are all `NA`)

----
```r
ggplot() + layer_spatial(CL1, aes(col = contour))
```
CL_plot-current
(Note that contour IDs are all ranges.)

---
```r
ggplot() + layer_spatial(CL1, aes(col = contour))
```
CL_plot-current-2
(Two sets of contours -- those at levels 30 and 60 -- are ID'd as `[30,60]`)

JoshOBrien:
### Figures produced by the MRE above, with PR's fixes applied

----
```r
ggplot() + layer_spatial(CP, aes(fill = range))
```

CP_plot-fixed

----
```r
ggplot() + layer_spatial(CL1, aes(col = contour))
```

CL_plot-fixed
(Note that contour IDs are scalar levels)

---
```r
ggplot() + layer_spatial(CL1, aes(col = contour))
```

(Produces same figure as that just above, as we would hope it would)

edzer:
I think I get your point. I don't get why `st_sample()` could be part of an acceptable solution, though.

JoshOBrien:
> I don't get why st_sample() could be part of an acceptable solution, though.

Do you mean that you don't get why `st_sample()` could **not** be part of an acceptable solution?

I think it might be acceptable. However, from the way the existing code is written (with labels constructed as ranges), and from the fact that labeling for polygons doesn't currently work at all, I assumed (maybe wrongly) that at some point in the past `process_cpl_read_ogr(CPL_polygonize())` **did** return a first column with useful (non-`NA`) values, but that it is no longer working that way. If that's were so, then a fix to `CPL_polygonize()` would be the way to go. But was that a wrong assumption? Did labeling of polygons by `gdal_polygonize()` not ever work?

Using `st_sample()` seems to work OK. I just don't know if it's always reliable. (I first tried using `st_sample()` with `size=1`, and sometimes got incorrect results because some cells (or parts of cells) intersected by a polygon can have values that don't reflect the region's intended range of values. Setting `size=50` and taking the mean of the returned cell values seems fairly reliable, but I didn't know if there are corner cases where it will fail.)

Is that any clearer?

edzer:
Ah, I misunderstood: I thought you were sampling the whole area, but you are sampling the resulting polygons, as a way of querying their attribute values.

JoshOBrien:
Exactly. Well put.

edzer:
I tried to solve this at another point in the code; please check.

JoshOBrien:
Nice. I was hoping an upstream (C++) fix just like this would be possible.

I do notice one problem with polygon labeling behavior, that occurs when the lowest-valued user-supplied break falls within (rather than below) the raster's data range. (I just wrote a much longer description of the problem and then lost it all when unintentionally closing my browser window. If this description is a bit short, that's why!)

### Demonstration of the problematic behavior

```r
library(stars)
library(ggplot2)
library(ggspatial)
trace(gdal_polygonize, edit = TRUE)
delta = min(diff(sort(breaks))) * 0.1
m = as.integer(cut(out[[1]] - delta, breaks = nbreaks, include.lowest = TRUE))

## Prepare an example raster layer
tif <- system.file("tif/L7_ETMs.tif", package = "stars")
x6 <- read_stars(tif)[,,,6]
r <- st_warp(x6, cellsize = c(4*28, -4*28),
             crs = st_crs(x6))
range(r[[1]])
## [1]   2 218

## When the bottom break falls below the lowest value in the raster, everything works well
CP1 <- st_contour(r, breaks = c(0, 60, 90, 300), contour_lines = FALSE)
names(CP1)[[1]] <- "range"
ggplot() + layer_spatial(CP1, aes(fill = range)) +
    scale_fill_manual(values = scales::hue_pal()(3))
```
Range-covered-current


```r
## When the lowest break lies *within* the raster's data range, the MULTIPOLYGON surrounding 
## regions with values lower than that break is mis-identified
CP2 <- st_contour(r, breaks = c(60, 90), contour_lines = FALSE)
names(CP2)[[1]] <- "range"
ggplot() + layer_spatial(CP2, aes(fill = range)) +
    scale_fill_manual(values = scales::hue_pal()(3)[2])
```
Range-not-covered-current

edzer:
Thanks, this should've fixed it.

JoshOBrien:
That's still not quite right, unfortunately. I see two issues with it, one you'd definitely want to fix, and the other that's I guess more a matter of taste.

First, while the fix properly leads to the upper end of the range of breaks being labeled as `Inf`, it labels the bottom end as `0` rather than `-Inf` (even though `-Inf` is really what is meant). Here is an example, using a raster like the one used used above but with all of its values shifted downward by 50 so that its range includes negative values:


```r
library(stars)
library(ggplot2)
library(ggspatial)

## Prepare an example raster layer
tif <- system.file("tif/L7_ETMs.tif", package = "stars")
x6 <- read_stars(tif)[,,,6]
r <- st_warp(x6, cellsize = c(4*28, -4*28),
             crs = st_crs(x6))
r2 <- r - 50
range(r2[[1]])

## CP <- st_contour(r2, breaks = c(0, 60, 90), contour_lines = FALSE) ## another demo of the issue
CP <- st_contour(r2, breaks = c(-10, 60, 90), contour_lines = FALSE)
names(CP)[[1]] <- "range"
ggplot() + layer_spatial(CP, aes(fill = range))
```
(The bottom range in this figure's legend should instead be labeled `[-Inf,-10)`)
New-version-mislabels-neg-Inf

edzer:
The `0` seems to be a gdal problem, when writing `r2` to `r2.tif`:
```
gdal_contour -p -amin min -amax max -fl -100 -20 -10 20 50 100 300 r2.tif r2.gpkg
ogrinfo r2.gpkg  contour > x
grep '(Real)' x
```
gives
```
  min (Real) = 0
  max (Real) = -20
  min (Real) = -20
  max (Real) = -10
  min (Real) = -10
  max (Real) = 20
  min (Real) = 20
  max (Real) = 50
  min (Real) = 50
  max (Real) = 100
  min (Real) = 100
  max (Real) = 300
```
so I propose to replace that with the data minimum.

edzer:
I can see your second point, but would like to stay close to what `gdal_contour` does; you now have the `Min` and `Max` fields on which you can filter.

JoshOBrien:
Alternatively, replacing it with `-Inf` preserves symmetry with the way that values falling outside of and **above** the data range are labeled. Using `-Inf` also conveys (to me) that the data fell outside the range of main interest. Finally, having `-Inf` in the label would make it easier for me (by string pattern matching on the labels) to filter out `MULTIPOLYGON`s that represent data outside the range of interest.

I still think it'd be nicest to only return `MULTIPOLYGON`s within the range of breaks the user supplied. Users can, after all, always include `-Inf` and `Inf` in `breaks=` if they affirmatively want those polygons, so it'd be nice let them also indicate that they don't want them. But if the decision is to always return those polygons, I think it'd be helpful for them to include `-Inf` and `Inf` in their labels.

JoshOBrien:
Didn't see your last reply before writing the comment above. Having those fields to filter on does make this much handier. I still think best would be to make the minimum on the lowest range be `-Inf` (when it's including data that fall outside the range of supplied breaks), rather than the lowest data value, for ease of identifying that case and for symmetry with the `Inf` in the `Max` column of the top range.

JoshOBrien:
Commit a139f48 still doesn't get things quite right. I'm happy to explain why, but I think demonstrating the behavior and showing my proposed fix will make the situation clear enough.

First, here's my proposed fix (again to that same code block near the bottom of `gdal_polygonize()`). (Because the current approach isn't working right, I couldn't tell for sure what you are wanting to happen when the raster's data range extends below the lowest supplied break. My proposed replacement sets the lower bound of that bottom interval to `0` when the raster is fully positive and `-Inf` when the raster contains some negative values. If you'd instead like the lower bound to be set to `-Inf` whatever the values in the raster, a small tweak will fix that.)

```r
                ## CODE TO REPLACED
		## if (out$Min[1] == 0 && out$Min[1] > min(breaks))
		## 	out$Min[1] = -Inf

                ##  PROPOSED REPLACEMENT CODE
                i <- match(out$Max[1], sort(breaks))
                if (i > 1) {
                    out$Min[1] = sort(breaks)[i - 1]
                }
                else if (min(x[[1]])) {
                    out$Min[1] = -Inf
                }
```

Here is a reproducible example showing some cases where the current code fails, alongside what I think the results should be. Following each call to `st_contour()`, the first line of results shows output from the current code, while the second line shows results after applying the patch above.

```r
library(stars)

## Prepare two raster layer, one fully positive, the other with negative values
tif <- system.file("tif/L7_ETMs.tif", package = "stars")
x6 <- read_stars(tif)[,,,6]
r_pos <- st_warp(x6, cellsize = c(4*28, -4*28),
                 crs = st_crs(x6))
r_neg <- r_pos - 50

##------------------------------------------------------------------------------
## Breaks for raster with minimum value of 2
##------------------------------------------------------------------------------

range(r_pos[[1]])
## [1]   2 218

levels(st_contour(r_pos, breaks = c(10, 90), contour_lines = FALSE)[[1]])
## [1] "[0,10)"   "[10,90)"  "[90,Inf)"
## [1] "[0,10)"   "[10,90)"  "[90,Inf)"

levels(st_contour(r_pos, breaks = c(1, 90), contour_lines = FALSE)[[1]])
## [1] "[0,90)"   "[90,Inf)"                ## <- Incorrect
## [1] "[1,90)"   "[90,Inf)"

levels(st_contour(r_pos, breaks = c(-1, 90), contour_lines = FALSE)[[1]])
## [1] "[-Inf,90)" "[90,Inf)"               ## <- Incorrect
## [1] "[-1,90)"  "[90,Inf)"


##------------------------------------------------------------------------------
## Breaks for raster with minimum value of -48
##------------------------------------------------------------------------------

range(r_neg[[1]])
## [1] -48 168

levels(st_contour(r_neg, breaks = c(10, 90), contour_lines = FALSE)[[1]])
## [1] "[0,10)"   "[10,90)"  "[90,Inf)"     ## <- Incorrect
## [1] "[-Inf,10)" "[10,90)"   "[90,Inf)"

levels(st_contour(r_neg, breaks = c(-40, 90), contour_lines = FALSE)[[1]])
## [1] "[-Inf,-40)" "[-40,90)"   "[90,Inf)"
## [1] "[-Inf,-40)" "[-40,90)"   "[90,Inf)"

levels(st_contour(r_neg, breaks = c(-100, 90), contour_lines = FALSE)[[1]])
## [1] "[-Inf,90)" "[90,Inf)"               ## <- Incorrect
## [1] "[-100,90)" "[90,Inf)"
```

JoshOBrien:
... and just for the sake of completeness, to **always** get `-Inf` as the bottom interval's lowest bound when the raster contains values below the lowest user-supplied break, use this replacement code instead:

```r
                ##  ALTERNATIVE REPLACEMENT CODE
                i <- match(out$Max[1], sort(breaks))
                if (i > 1) {
                    out$Min[1] = sort(breaks)[i - 1]
                }
                else {
                    out$Min[1] = -Inf
                }
```

edzer:
I followed that last suggestion; trying `min(x[[1]])` may fail when `x` has more than 2 dimensions and fails when it is a `stars_proxy` object.

JoshOBrien:
Cool. I wondered about that.

One other thing that occurred to me while looking at this code is that it might be desirable to enforce that breaks passed in by users are in ascending order. Not sure whether it'd be better to test and warn users if they aren't in order, or to just do `breaks = sort(breaks)` as a first step.

Supplying breaks in reversed order returns truly messed up `MULTIPOLYGON` results (many of which error out with a `"Points of LinearRing do not form a closed linestring."` message when plotted), so there doesn't seem to be much reason to think folks will want to use it that way.

PR #1626, opened 2021-03-11
use proj 7.1 for proj_context_set_url_endpoint

### LINKED ISSUE #1625: Build problem with Proj 7.0
I'm having isssues building the latest sf (ce439eb89cb9265f9bc2a2fb795d5422a6a24659) with proj 7.0.0. 
```
proj.cpp:186:40: note: suggested alternative: ‘proj_context_set_url_endpoint’
   return Rcpp::CharacterVector::create(proj_context_get_url_endpoint(PJ_DEFAULT_CTX));
                                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                        proj_context_set_url_endpoint
/usr/lib/R/etc/Makeconf:177: recipe for target 'proj.o' failed
make: *** [proj.o] Error 1
```

Using bisect, the problem starts at e6a6a4d3e1f1ca3c42c0b0f6feaa245fc3530bae, which introduces 
https://github.com/r-spatial/sf/blob/e6a6a4d3e1f1ca3c42c0b0f6feaa245fc3530bae/src/proj.cpp#L29-L43
But, it seems that `proj_context_get_url_endpoint` is only available in 7.1
image

https://proj.org/development/reference/functions.html#_CPPv429proj_context_get_url_endpointP10PJ_CONTEXT

etiennebr:
fix #1625

### DIFF
--- a/src/proj.cpp
+++ b/src/proj.cpp
@@ -178,7 +178,7 @@ Rcpp::LogicalVector CPL_is_network_enabled(bool b = false) {
 }
 
 Rcpp::CharacterVector CPL_enable_network(Rcpp::CharacterVector url, bool enable = true) {
-#if PROJ_VERSION_MAJOR >= 7
+#if PROJ_VERSION_MAJOR >= 7 && PROJ_VERSION_MINOR >=1
 	if (enable) {
 		proj_context_set_enable_network(PJ_DEFAULT_CTX, 1);
 		if (url.size())


src/proj.cpp
```diff
@@ -178,7 +178,7 @@ Rcpp::LogicalVector CPL_is_network_enabled(bool b = false) {
 }
 
 Rcpp::CharacterVector CPL_enable_network(Rcpp::CharacterVector url, bool enable = true) {
-#if PROJ_VERSION_MAJOR >= 7
+#if PROJ_VERSION_MAJOR >= 7 && PROJ_VERSION_MINOR >=1
```
edzer, line 181: That doesn't work for 8.0.x

etiennebr: ... and we can't run preprocessor on float :thinking:

edzer: GDAL does [this](https://github.com/r-spatial/sf/blob/master/src/gdal.cpp#L662) kind of things, and it's a nightmare too.

PR #1631, opened 2021-03-16
Fix st_as_grob issue when plotting mix of MULTI and base type

thomasp85:
This PR fixes an issue that surfaced in https://github.com/thomasp85/gganimate/issues/358

My initial implementation forgot to forward grab specific settings pertaining to `pointGrob` when forwarding calls


edzer:
Great, thanks!

PR #1654, opened 2021-04-21
fixes #1653 on branch

### LINKED ISSUE #1653: sf slightly broken by GEOS 3.10.0
When testing the https://github.com/pramsey/geos/tree/main-geometry-fixer proposal, I ran into version number testing issues, also found this in **sf**. In `sf/tests/testthat/test_geos.R`, line 192:

```
sf_extSoftVersion()["GEOS"] < "3.9.0"
# GEOS 
# TRUE 
sf_extSoftVersion()["GEOS"]
#       GEOS 
# "3.10.0dev" 
```
My **rgeos** fix to the same problem is to post-process the returned string through `base::package_version()` and then to use its operations:
```
version_GEOS0() < "3.9.0"
# [1] FALSE
```
in `rgeos/R/AAA.R` line 91, returning:
```
class(version_GEOS0())
# [1] "package_version" "numeric_version"
```
I probably need to be more systematic in other places, but when branching on versions, this helped.

rsbivand:
Light-touch conversion in-test for version checking, in this branch because that was where I was.


edzer:
Thanks!

PR #1656, opened 2021-04-22
extent length of graphical parameters after unlisting a geometrycollections

thomasp85:
This PR addresses the issue reported in https://github.com/tidyverse/ggplot2/issues/4275

When calling `st_cast_sfc_default()` on a geometry collection the length of the output will (likely) change and the graphical parameters need to be matched up to the new length in order to get rendered as expected


edzer:
Great, thanks!

PR #1669, opened 2021-05-04
Pass an `sfc` object to lapply() instead than `sfg`

ranghetti:
Running the following code I am encountering the error reported below:
```r
library(s2)
library(sf)
packageVersion("sf")
# [1] ‘0.9.9’
packageVersion("s2")
# [1] ‘1.0.4.9000’
sf_use_s2(TRUE)

txt <- c("POLYGON ((0 0, 0 1, 1 1, 1 0, 0 0))",
         "POLYGON ((0.1 0.1, 0.1 0.5, 0.5 0.5, 0.5 0.1, 0.1 0.1))")
st_difference(st_as_sfc(as_s2_geography(txt[1])), st_as_sfc(as_s2_geography(txt[2])))
# Error in UseMethod("as_s2_geography") : 
#   no applicable method for 'as_s2_geography' applied to an object of class "c('XY', 'POLYGON', 'sfg')"
```

I see that the error occurs here: https://github.com/r-spatial/sf/blob/4911f5a888311384e5e2fb3c781836b0e4f981a0/R/geom-transformers.R#L650-L651
The proposed PR, although not elegant, allows passing a `sfc` geometry to `fn()` within `lapply()` instead than a `sfg` object.

Feel free to reject if it is not the most suitable solution.

Ps. the PR is toward branch `master`, but it should also affect `sf_1_0`... If it would be better to pull to `sf_1_0` before, you can make a PR from `ranghetti/sf:sf_1_0` which includes the commit.

### DIFF
--- a/R/geom-transformers.R
+++ b/R/geom-transformers.R
@@ -647,8 +647,9 @@ geos_op2_geom = function(op, x, y, s2_model = "semi-open", ...) {
 				sym_difference = s2::s2_sym_difference,
 				union = s2::s2_union, stop("invalid operator"))
 		# to be optimized -- this doesn't index on y:
-		lst = structure(unlist(lapply(x, fn, y, s2::s2_options(model = s2_model, ...)),
-			recursive = FALSE), class = "s2_geography")
+		lst = structure(unlist(lapply(seq_along(x), function(i) {
+			fn(x[i], y, s2::s2_options(model = s2_model, ...))
+		}), recursive = FALSE), class = "s2_geography")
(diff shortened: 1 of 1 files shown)


edzer:
I can't reproduce the error you see; what is your output of `sessionInfo()`?

edzer:
Ah, I see - this is with `s2` from github; the problem might lie there.

ranghetti:
Oops, I did not try with GitHub sf and CRAN s2, sorry. Thank you for addressing the issue.

Issue #1666, opened 2021-05-01
st_is_valid performance


pbaylis:
This is may be a dumb question, but I was surprised to find that `st_is_valid` not only runs more slowly than `st_make_valid`, but it also scales non-linearly as the number of features increases. 

I include a MWE demonstrating the two observations below (I ran this with `sf_0.9-8`). For my actual use case, I have around 11m features, and at least one of them is causing some sort of downstream validity problem when I try to join or crop to other datasets. Unfortunately it's not one of the problems that `st_make_valid` can resolve (invalid number of points exception), so I'm reliant on `st_is_valid`, which hasn't been able to complete in a reasonable amount of time. 

I know these functions are reliant on backend code, so maybe this is not a resolvable problem, but if anyone has any ideas for improving performance here I'd be very interested. Also happy to relocate this to StackOverflow if it's more appropriately asked there.

```
library(sf)
library(dplyr)
library(tictoc)

b0 <- st_polygon(list(rbind(c(-1,-1), c(1,-1), c(1,1), c(-1,1), c(-1,-1))))
b1 <- st_polygon(list(rbind(c(-1,-1), c(1,-1), c(1,1), c(0,-1), c(-1,-1))))

b <- st_sfc(b0, b1) %>% st_sf()

tic()
invisible(st_is_valid(b %>% sample_n(5000, replace = T)))
toc()
# 1.2s on my laptop            

tic()
invisible(st_make_valid(b %>% sample_n(5000, replace = T)))
toc()
# 0.7s s on my laptop

tic()
invisible(st_is_valid(b %>% sample_n(50000, replace = T)))
toc()
# 34s on my laptop            

tic()
invisible(st_make_valid(b %>% sample_n(50000, replace = T)))
toc()
# 6.5s on my laptop
```

edzer:
As you suggested, work here does not involve actual implementing these functions, but interfacing existing implementations. Hopeful news is that there's both activity in this area going on in GEOS (3.10-dev) and the R s2 package (for spherical validity).

dbaston:
With trivial inputs, the timing is dominated by R overhead rather than the underlying implementations. in the 34s example, less than 10% of the time is spent in GEOS. It looks like a (primary?) difference here is that `st_is_valid` uses the subsetting operator for an `sfc`, which does a bit of work, whereas `st_make_valid` does not.

[...]

PR #1673, opened 2021-05-04
Remove unclass call from [.sfc

dbaston:
References #1666. This does not replace the solution in 4911f5a888311384e5e2fb3c781836b0e4f981a0 but greatly improves runtime in other cases where you need to index into an `sfc`. There appear to be other cases that could could benefit from the same technique.

### DIFF
--- a/R/sfc.R
+++ b/R/sfc.R
@@ -145,7 +145,7 @@ sfg_is_empty = function(x) {
 "[.sfc" = function(x, i, j, ..., op = st_intersects) {
 	if (!missing(i) && (inherits(i, "sf") || inherits(i, "sfc") || inherits(i, "sfg")))
 		i = lengths(op(x, i, ...)) != 0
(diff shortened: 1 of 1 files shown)


edzer:
Makes total sense, I'm amazed that this matters so much. Thanks!

dbaston:
From `?unclass`, "unclass returns (a copy of) its argument with its class attribute removed. (It is not allowed for objects which cannot be copied, namely environments and external pointers.)"

So if you loop over 50,000 geometries, you copy 50,000 geometries 50,000 times :)

dbaston:
@edzer So apparently this commit broke the following:

```
x <- st_as_sfc('POINT (3 8)')
x[1]   # works
x[1, ] # used to work, now gives an "incorrect number of dimensions" error
```

Was the second usage supposed to work? Is the behavior change a problem?

edzer:
No, that was not supposed to work. `sfc` is a list, and needs a single index. Let's see what the revdep checks give, it may be necessary to postpone this change if things are bad.

PR #1701, opened 2021-06-21
A few more failing GEOS version tests for 3.10

rsbivand:
I tried rebuilding SDSR, but failed on:

```
uitting from lines 1679-1691 (sds.Rmd) 
Error in st_voronoi.sfc(st_sfc(x), st_sfc(envelope), dTolerance, bOnlyEdges = bOnlyEdges) : 
  for voronoi, GEOS version 3.5.0 or higher is required
Calls: <Anonymous> ... st_voronoi.sfg -> get_first_sfg -> st_voronoi -> st_voronoi.sfc
```
To unblock it, I need these updates.


edzer:
Thanks!

PR #1702, opened 2021-06-23
removes --static

edzer:


rsbivand:
Could I suggest changing the test to try with `--static` first as a separate test, and if it fails remove it from `PROJ_LIB_PATH` and try again? Failing `--static` would be a message or warning, not an error.

edzer:
I have no recollection why `--static` was added, and cannot find any reason in the issues or PRs. Maybe it was done by accident, like a wild guess, by accident (..) not in a branch? If nobody remembers this, I suggest reverting.

PR #1709, opened 2021-06-27
Update `update-tic.yml`

pat-s:
Required for future executions of the WF as the `@master` ref will stop working after June 30th for the `peter-evans/create-pull-request` action.


edzer:
Thanks!

PR #1712, opened 2021-07-02
Added brackets to uname command

pascalgulikers:


edzer:
`configure` is created automatically from `configure.ac` by `autoconf`, so changing this does not make sense.

pascalgulikers:
Well, it's broken since this merge: https://github.com/r-spatial/sf/pull/1702/files
You can see the differences between configure and configure.ac here

edzer:
I can see that, but you can see my point too, right? If the line in `configure` is
```
    if test `uname` = "Darwin"; then
```
does that work on Fedora? I know how to generate that from `configure.ac`.

pascalgulikers:
Yes, that would work

pascalgulikers:
I have changed it

pascalgulikers:
@edzer Do you have any idea when the patches will be available at CRAN?

edzer:
> @edzer Do you have any idea when the patches will be available at CRAN?

No. But this PR can be closed, right?

pascalgulikers:
Fixed in https://github.com/r-spatial/sf/commit/03210a8504673d5f900607fa679c196e80d9b1ba

fritzratna:
hi, I am trying to install sf from CRAN but i am getting the same error mentioned earlier
```
configure: GDAL: 2.4.1
configure: pkg-config proj exists, will use it
configure: using proj.h.
./configure: line 3769: Linux: command not found
configure: PROJ: 6.3.2
checking PROJ: checking whether PROJ and sqlite3 are available for linking:... no
configure: error: libproj or sqlite3 not found in standard or given locations.
ERROR: configuration failed for package ‘sf’
```
I have installed GDAL, GEO and PROJ lib but I have no idea how to fix this issue

edzer:
Have you tried `remotes::install_github("r-spatial/sf")` ?

PR #1716, opened 2021-07-06
Support terra objects in st_crs

dbaston:


edzer:
Yes, why not. The recommended way to do this is found here: https://github.com/r-spatial/stars/blob/master/R/raster.R#L194 (and further down that file).

PR #1734, opened 2021-07-23
Tweak to ensure rwinlib compatibity with ucrt

jeroen:
Small tweak for future compatibility.

### DIFF
--- a/src/Makevars.ucrt
+++ b/src/Makevars.ucrt
@@ -1,0 +1,2 @@
+CRT=-ucrt
+include Makevars.win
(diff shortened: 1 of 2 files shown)


edzer:
Great! I hope this will resolve the [error](https://www.r-project.org/nosvn/R.check/r-devel-windows-x86_64-gcc10-UCRT/stars-00check.html) stars currently has on UCRT when reading a netcdf file (it does this by calling GDAL routines in `sf`).

jeroen:
Yes I think it will. With this, the behaviour between the ucrt and non-ucrt builds should be 100% consistent.

rsbivand:
Thanks Jeroen! I made the same tweak for **rgeos** rev 672 https://r-forge.r-project.org/scm/viewvc.php?root=rgeos&view=rev and **rgdal**  rev 1135 https://r-forge.r-project.org/scm/viewvc.php?root=rgdal&view=rev. Does that make sense?

jeroen:
Yes that looks correct, thank you!

PR #1754, opened 2021-08-09
Correct small typo in R/read.R

rhurlin:
Although only a temporarily needed comment, nevertheless corrected even better understandable ;)


edzer:
Thanks!

PR #1755, opened 2021-08-12
should have endif here?

mdsumner:
looks like a dangling if, not sure


mdsumner:
ah no ... just a big long condition nvm

PR #1775, opened 2021-08-31
vignette sf7 (s2): minor fixes

florisvdh:
Made a few typo and language fixes, while reading the vignette on S2 last week. A very nice vignette!

Some other notes:

- line 288: something is wrong with the last sentence
- line 310: `uk_sfc = st_as_sfc(uk)`
  - the result `uk_sfc` gets 'WGS 84' CRS by default. I presume that this is intended, but I couldn't find it documented somewhere (not for `st_as_sfc()`). Perhaps it is behaviour specific to `s2_geography` objects? Then my suggestion, not for the vignette but for the function docs, would be to document this case.

### DIFF
--- a/vignettes/sf7.Rmd
+++ b/vignettes/sf7.Rmd
@@ -23,7 +23,7 @@ knitr::opts_chunk$set(collapse = TRUE)
 This vignette describes what spherical geometry implies, and how
 package `sf` uses the s2geometry library (https://s2geometry.io)
 for geometrical measures, predicates and transformations.
-The code in this vignette only run if `s2` has been installed, e.g. with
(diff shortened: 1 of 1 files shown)


edzer:
Thanks - the `st_as_sfc` and `st_as_sf` docs should now be in their right places.

florisvdh:
Great, behaviour for `s2_geography` objects was documented indeed! 

Still something wrong here (my first note): a half sentence in line 288, maybe some lines have switched or so:

https://github.com/r-spatial/sf/blob/2c53a4eaec86a79e9ef720557abd3d29e3a13aea/vignettes/sf7.Rmd#L285-L300

PR #1801, opened 2021-09-22
Change st_simplify

agila5:
The current implementation of `st_simplify` always returns a warning message when the input object is specified with lat-long coordinates and `sf_use_s2` returns `TRUE` since the argument `preserveTopology` is never evaluated as missing: 

``` r
remotes::install_github("r-spatial/sf", quiet = TRUE)
library(sf)
#> Linking to GEOS 3.9.1, GDAL 3.2.1, PROJ 7.2.1
nc = st_read(system.file("shape/nc.shp", package="sf"), quiet = TRUE)
nc = st_simplify(nc)
#> Warning in st_simplify.sfc(st_geometry(x), preserveTopology, dTolerance):
#> argument preserveTopology is ignored
```

<sup>Created on 2021-09-22 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

I created this PR to propose a solution that (I think) preserves the original behaviour and removes that warning message when is not needed. I also added some examples and wrote more details in the docs. This is the figure created in the examples: 

``` r
library(sf)
#> Linking to GEOS 3.9.1, GDAL 3.2.1, PROJ 7.2.1
nc = st_read(system.file("shape/nc.shp", package="sf"))
nc_g = st_geometry(nc)
op = par(mfrow = c(2, 3), mar = rep(0, 4))
plot(nc_g[1])
plot(st_simplify(nc_g[1], dTolerance = 1e3)) # 1000m
plot(st_simplify(nc_g[1], dTolerance = 5e3)) # 5000m
nc_g_planar = st_transform(nc_g, 2264) # planar coordinates, US foot
plot(nc_g_planar[1])
plot(st_simplify(nc_g_planar[1], dTolerance = 1e3)) # 1000 foot
plot(st_simplify(nc_g_planar[1], dTolerance = 5e3)) # 5000 foot
```

<sup>Created on 2021-09-22 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -113,7 +113,7 @@ LinkingTo:
 VignetteBuilder: 
     knitr
 Encoding: UTF-8
-RoxygenNote: 7.1.1
+RoxygenNote: 7.1.2
 SystemRequirements: C++11, GDAL (>= 2.0.1), GEOS (>= 3.4.0),
     PROJ (>= 4.8.0), sqlite3
 Collate: 
--- a/R/geom-transformers.R
+++ b/R/geom-transformers.R
@@ -194,23 +194,49 @@ st_convex_hull.sf = function(x) {
 #' @name geos_unary
 #' @export
 #' @details \code{st_simplify} simplifies lines by removing vertices
-#' @param preserveTopology logical; carry out topology preserving simplification? May be specified for each, or for all feature geometries. Note that topology is preserved only for single feature geometries, not for sets of them.
-#' @param dTolerance numeric; tolerance parameter, specified for all or for each feature geometry.
-st_simplify = function(x, preserveTopology = FALSE, dTolerance = 0.0)
+#' @param preserveTopology logical; carry out topology preserving
+#'   simplification? May be specified for each, or for all feature geometries.
+#'   Note that topology is preserved only for single feature geometries, not for
+#'   sets of them. Ignored when the input data is specified with long-lat
+#'   coordinates and \code{sf_use_s2()} returns \code{TRUE} since, in that case,
+#'   \code{st_simplify} implicitly calls \code{s2::s2_simplify} which always
+#'   preserve topology (per single feature).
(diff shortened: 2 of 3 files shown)


edzer:
I see the problem; but setting it to `TRUE` for data with ellipsoidal coordinates should not lead to a warning, right?

agila5:
I'm sorry but I don't understand the question.

edzer:
With your patch, we have
```r
library(sf)
# Linking to GEOS 3.9.0, GDAL 3.2.1, PROJ 7.2.1
nc = st_read(system.file("shape/nc.shp", package="sf"), quiet = TRUE)
nc = st_simplify(nc, preserveTopology = TRUE)
# Warning message:
# In st_simplify.sfc(st_geometry(x), preserveTopology, dTolerance) :
#   argument preserveTopology is ignored
```
but in that case I would rather not see a warning, because it is assumed to be `TRUE`.

agila5:
Should be fixed with the last two commits: 

``` r
library(sf)
#> Linking to GEOS 3.9.1, GDAL 3.2.1, PROJ 7.2.1
nc = st_read(system.file("shape/nc.shp", package="sf"), quiet = TRUE)
nc2 = st_simplify(nc)
nc3 = st_simplify(nc, preserveTopology = TRUE)
nc4 = st_simplify(nc, preserveTopology = FALSE)
#> Warning in st_simplify.sfc(st_geometry(x), preserveTopology, dTolerance):
#> argument preserveTopology is ignored
```

<sup>Created on 2021-09-22 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Maybe I can slightly adjust the warning message? Something like `argument preserveTopology cannot be set to FALSE since the algorithm behind st_simplify always preserves topological relationships`?

edzer:
Yes, but add that this is only the case for ellipsoidal coordinates.

agila5:
Done 🎉

edzer:
Great work - thanks!

Issue #1802, opened 2021-09-23
st_buffer ignores nQuadSegs when sf_use_s2(TRUE)


juansebastianl:
I'm not sure if this is an issue or expected behavior (sort of like the st_crop issue) but it took me a long time to figure out what was going on, so even if it is expected behavior I think it should be documented. 

If I have the default `sf_use_s2(TRUE)` then `st_buffer` behaves completely differently than expected. For example running 

    tibble(id = 1, lon = 0, lat = 0) %>% 
      st_as_sf(coords = c("lon", "lat")) %>% 
      st_set_crs(4326) %>% 
      st_buffer(111000, nQuadSegs = 1, endCapStyle = "SQUARE") %>% 
      ggplot() + 
         geom_sf()

produces a circle of radius 1 degree no matter the value of `nQuadSegs` or `endCapStyle`, If I run this after running `sf_use_s2(FALSE)` I would get a square of radius larger than the whole earth.

edzer:
Yes, agreed this needs improved documentation; see e.g. #1801 for a successful attempt to clear up a similar problem for `st_simplify()`.

juansebastianl:
Okay, it seems `nQuadSegs` is completely ignored because the s2 library doesn't reproduce the same ability to buffer as geos, right? Specifically this issue https://github.com/r-spatial/s2/issues/42 . There is code in the `st_buffer.sfc` function which is commented out which outputs a warning. Uncommenting that would be helpful, but it doesn't seem to make sense to me why this function should use `s2` as a default when `s2` can't actually produce the same results. 

Would it be possible to have it use geos and have a warning and a flag along the lines of `warning: st_buffer uses geos instead of s2 for calculation by default, use s2 = T to use s2` and then a warning when using `s2` along the lines of `warning: st_buffer ignores all arguments when using s2 = T` . 

When the issue in `s2` is resolved then it can be reverted to the current default behavior.

edzer:
> but it doesn't seem to make sense to me why this function should use s2 as a default when s2 can't actually produce the same results.

because GEOS can only create _round_ buffers on the equator, everywhere else they are ellipses, getting worse the further you go away from the equator. s2 creates round shapes on the sphere, regrettably not with smooth borders.

> When the issue in s2 is resolved then it can be reverted to the current default behavior.

[...]

PR #1803, opened 2021-09-25
St buffer changes

juansebastianl:
This is a change made to solve issue #1802 which updates the documentation and outputs a warning if any non-default parameters are used. 

I ran roxygen2 to update the documentation and a bunch of files show like they have changes in git, but they didn't actually change it seems.


edzer:
Sorry, I wrote this review but never finished it! Happy to accept after some small mods.

Issue #1814, opened 2021-10-11
Suggestion: use issue template 


rCarto:
Hi, 
A common problem in some of the issues submitted to this repo is the lack of valid reprex and/or contextual information. 
A suggestion to improve the completeness of such issues would be to use (simple) [issue templates](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository). 
The template could look like this:

```md
**Describe the bug**
A clear and concise description of what the bug is.

**To Reproduce**
It is easier for developers to help you if they can reproduce your problem.   
Can you please provide a minimal [reproducible example](https://community.rstudio.com/t/faq-whats-a-reproducible-example-reprex-and-how-do-i-create-one/5219).

**Additional context**
Add any other context about the problem here.

<details>
Paste the output of your `sessionInfo()`
</details>
```

Of course this template should be optional and "free text" issues should remain available. 

What do you think?

PR #1818, opened 2021-10-13
add issue templates

rCarto:
As discussed in #1814
Issue templates are just simple md files in the .github folder. They can be a lot more sophisticated but I don't think that's useful here. 
You can can try it on my fork to see what it looks like.


edzer:
Thanks, @rCarto !

Issue #1817, opened 2021-10-13
st_intersection in s2 mode does not accept model as argument name / expects s2_model


jlacko:
The st_intersection function seems not to pass the dot dot dot arguments to s2::s2_options() as expected.

As a reproducible example consider this piece of code meant to count the neighboring counties of cnty Mecklenburg in the nc shapefile; five are expected (and delivered in the GEOS mode); in s2 mode getting the same result requires tweaking the model argument, as s2 polygons are by default "semi-closed" and thus do not share borders with their neighbours; forcing the expected behaviour requires setting the model to "closed". This is a well documented and perfectly reasonable difference between s2 and geos behaviour and I do not complain.

The setting of model option however does not behave as expected; I believe the problem is in the internal `geos_op2_geom()`  function, where the `model` argument of `sf::s2_options()` gets overriden by the `s2_model` argument; see the line 663 bellow.

https://github.com/r-spatial/sf/blob/ea94efe83cf2ef575bf6e9d59e04a07301d06b28/R/geom-transformers.R#L663

This behaviour seems not to be in line with the package documentation (which implies that the ... is passed to s2_options unchanged). I suggest either renaming the `s2_model` argument of the `geos_op2_geom()` function to `model` to align the naming convention, or updating the documentation of sf::st_intersection() to explicitly mention the `s2_model` argument and its behaviour.

I would be happy to provide a pull request with either of these, if agreeable to the package maintainers.

``` r
library(sf)

shape <- st_read(system.file("shape/nc.shp", package="sf"))

mecklenburg <- shape[shape$NAME == 'Mecklenburg', ]
rest_of_nc <- shape[shape$NAME != 'Mecklenburg', ]

sf_use_s2(F) # GEOS regime as a baseline

st_intersection(mecklenburg, rest_of_nc) |>
  nrow() # 5 borders 

sf_use_s2(T) # s2 regime

st_intersection(mecklenburg, rest_of_nc) |>
  nrow() # 0 borders, because s2 polygons are by default semi-open

st_intersection(mecklenburg, rest_of_nc, model = "closed") |>
  nrow() # throws an error / but the argument is named as per docs

st_intersection(mecklenburg, rest_of_nc, s2_model = "closed") |>
  nrow() # works fine / but the argument name is not documented
```

PR #1819, opened 2021-10-13
in geos_op2_geom() renaming argument (s2_model >> model)

jlacko:
This should resolve the problem raised in  #1817 / passes the reprex in the issue and seems not to break any unit tests


jlacko:
I'd be happy to! 

I noticed the mismatch in st_intersection first - I needed to force the "closed" model behaviour to reproduce earlier results from GEOS era, and ended up with two model arguments in `s2::s2_options`; one from my dot dot dot and the other from the built in `s2_model` - and did not delve too much into the rest of the codebase.

To cut long story short: I will update the pull request, hopefully later today.

PR #1825, opened 2021-10-20
Vignette sf7 (spherical): stress condition of GEOGCRS for s2

florisvdh:
In the vignette introduction, it is said once already, but IMHO rather inconspicuously (boldface added here):

> Most `sf` functions automatically use `s2` functions **when working with ellipsoidal coordinates**; [...]

I suggest to stress this aspect again where the vignette says:

> Starting with `sf` version 1.0, sf uses the new package `s2` (Dunnington, Pebesma, Rubak 2020) for spherical geometry [...]

For some readers it may be not clear enough that `s2` is not going to be used for projected coordinates. It may seem as if `s2` is always going to be used, which isn't the case, so I suggest to make that more visible.

### DIFF
--- a/vignettes/sf7.Rmd
+++ b/vignettes/sf7.Rmd
@@ -96,7 +96,8 @@ and whether it is
 * a straight line, cutting through the Earth's surface, or
 * a curved line following the Earth's surface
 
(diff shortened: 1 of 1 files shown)


edzer:
Thanks!

PR #1849, opened 2021-11-19
CPL_extract: initialize xsize and ysize

ateucher:
This appears to fix r[spatial/stars#459](https://github.com/r-spatial/stars/issues/459) by initializing variables for `poDataset->GetRasterXSize()` and `poDataset->GetRasterYSize()`, before they are accessed in an `if()` statement.

### DIFF
--- a/src/stars.cpp
+++ b/src/stars.cpp
@@ -165,7 +165,7 @@ NumericVector read_gdal_data(GDALDataset *poDataset,
 		if (has_offset)
 			offset = poBand->GetOffset(NULL);
 		units[i] = poBand->GetUnitType();
-		// if (! NumericVector::is_na(nodatavalue[0]) || has_offset || has_scale) { 
+		// if (! NumericVector::is_na(nodatavalue[0]) || has_offset || has_scale) {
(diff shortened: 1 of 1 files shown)


edzer:
Do you understand why this could fix it? I don't see it.

ateucher:
I haven't a clue - I don't know enough about C++, this was just trial and error. When I put in a `Sprintf()` to see the value of it, it stopped segfaulting. Seems akin to using `force()` in R to force greedy evaluation.

edzer:
If it fixes it, it fixes it, but all I can think of is that that would indicate an error in the compiler.

ateucher:
Yeah it doesn't seem right... it may be worth holding merging until we figure out the other issue with read.

edzer:
Well, the code is more readable your way IMO.

ateucher:
Ok I think it's worth merging this if you want. The read error I'm almost certain is a compilation issue, as you said. This may be as well but at least it will do no harm. I don't have the time to investigate the read error further until later this week, so if you want to take this one now feel free!

Issue #1782, opened 2021-09-03
How to silence sf_use_s2?


MikkoVihtakari:
Related to [this issue](https://github.com/r-spatial/sf/issues/1780), I need to add `sf::sf_use_s2(FALSE)` to a function to do cropping, and add `sf::sf_use_s2(TRUE)` once the cropping is done. This leads to the following print on console:

`Spherical geometry (s2) switched off
  |===========================================================================     |  93%Spherical geometry (s2) switched on`

In my opinion, the user does not need to know every time I switch s2 on or off. How to silence the print? `invisible(sf::sf_use_s2(FALSE))` does not seem to work.

edzer:
This is intentional: as `sf_use_s2()` changes global behaviour for the user, the user needs to be informed. 

Of course you can hack it: see what it does, and do that without informing the user; alternatively, set `sink(/dev/null)` (or sth appropriate for your platform) and set it back, but that's rigorous!

We may need to find an alternative for cropping (i.e., add an argument), that does this inside the function, without warning, but clear from the calling code.

MikkoVihtakari:
Why force the print? I'd just print a message instead. The user is informed in a similar way and one can avoid sink hacks to turn the message off. That said, I would not write a function, which modifies global behavior without telling the user. In this case, it's a temporary hack and s2 is switched back on afterward (should probably check whether s2 is on to begin with, though).

rCarto:
I use something like that in `mapsf::mf_worldmap()`:
https://github.com/riatelab/mapsf/blob/8f6167dbc5aec46b593a3c8b7a9f3d2986d52542/R/mf_worldmap.R#L16

But a 'quiet=FALSE' or 'verbose=TRUE' default arg in `sf_use_s2()` would be nice.

edzer:
Clever, but if that function breaks halfway, it will leave the R session of the user in a completely different mode than when it doesn't. It might be safer to set the s2 mode back with on.exit(), directly after setting it to the desired mode.

rCarto:
Oh yes, I see that now.

eliocamp:
R has a standard way of suppressing messages and warnings. `sf_use_s2()` should use `mesage()` instead of `cat()` and then the user can just use `suppressMessages(sf_use_s2(TRUE))`.

If the devs agree, I can open a PR with the change.

edzer:
This is intentional: if your code switches off s2 usage, the user needs to be notified because it has consequences for anything they do subsequently with sf.

[...]

PR #1864, opened 2021-12-13
add quiet param to sf_use_s2

mpadge:
There is currently no way to suppress the message, `Spherical geometry (s2) switched on/off`. This PR adds a `quiet` option to suppress if changed from default `FALSE` to `TRUE`.


edzer:
continuing the discussion on #1782

mpadge:
Thanks @edzer, i hadn't seen that discussion and am happy to close in favour of replacing `cat` with `message` (or similarl).

PR #1869, opened 2021-12-20
quietening CMD check test output

rsbivand:
bringing many `*.Rout.save` up to current GEOS (3.10.1), PROJ (8.2.0) and GDAL (3.4.0), adding `quiet=TRUE` to `st_read()` where missing.


edzer:
Thanks! I merged the essential bits manually.

Issue #1873, opened 2022-01-04
Should coercion from ppp to sf respect window scaling?


rsbivand:
For:

```
library(spatstat)
data(japanesepines)
japanesepines
#Planar point pattern: 65 points
#window: rectangle = [0, 1] x [0, 1] units (one unit = 5.7 metres)
str(japanesepines)
#List of 5
# $ window    :List of 4
# ..$ type  : chr "rectangle"
#  ..$ xrange: num [1:2] 0 1
#  ..$ yrange: num [1:2] 0 1
#  ..$ units :List of 3
#  .. ..$ singular  : chr "metre"
#  .. ..$ plural    : chr "metres"
#  .. ..$ multiplier: num 5.7
#  .. ..- attr(*, "class")= chr "unitname"
#  ..- attr(*, "class")= chr "owin"
# $ n         : int 65
# $ x         : num [1:65] 0.09 0.29 0.38 0.39 0.48 0.59 0.65 0.67 0.73 0.79 ...
# $ y         : num [1:65] 0.09 0.02 0.03 0.18 0.03 0.02 0.16 0.13 0.13 0.03 ...
# $ markformat: chr "none"
# - attr(*, "class")= chr "ppp"
```
**maptools** has:
```
library(maptools)
spjpines0 <- as(japanesepines, "SpatialPoints")
library(sp)
bbox(spjpines0)
#     min max
#[1,]   0 5.7
#[2,]   0 5.7
head(coordinates(spjpines0))
#        mx    my
#[1,] 0.513 0.513
#[2,] 1.653 0.114
#[3,] 2.166 0.171
#[4,] 2.223 1.026
#[5,] 2.736 0.171
#[6,] 3.363 0.114
```
that is both the bounding box set from the owin object and the coordinates themselves are rescaled by `window$units$multiplier`. In **sf**:
```
library(sf)
oo <- st_as_sf(japanesepines)
oo
#Simple feature collection with 66 features and 1 field
#Geometry type: GEOMETRY
#Dimension:     XY
#Bounding box:  xmin: 0 ymin: 0 xmax: 1 ymax: 1
#CRS:           NA
#First 10 features:
#    label                           geom
#1  window POLYGON ((0 0, 1 0, 1 1, 0 ...
#2   point              POINT (0.09 0.09)
#3   point              POINT (0.29 0.02)
#4   point              POINT (0.38 0.03)
#5   point              POINT (0.39 0.18)
#6   point              POINT (0.48 0.03)
```

[...]

PR #1875, opened 2022-01-05
Check for spatstat multiplier and warn

rubak:
Trying a quick fix for issue #1873 raised by @rsbivand. All coded in browser with quick testing in a terminal. Fingers crossed for the automated checks.


rsbivand:
@rubak Excellent! I'll consider adding a message to the **maptools** coercion method, but with this resolution, I should be able to use the **sf** coercion methods in our ASDAR book once **maptools** is retired. When that happens, `"ppp"` -> `"SpatialPoints"` will be `"ppp"` -> (rescale if desired) -> `"sf"` -> (drop window) -> `"SpatialPoints"`.

edzer:
Thanks, Ege!

PR #1892, opened 2022-01-26
Adapt to new spatstat.random package.

rubak:
Random generators are being moved from `spatstat.core` to a new package `spatstat.random`.
The proposed changes should hopefully fix the upcoming problem for your package.


edzer:
Ah, I see that `spatstat.random` is already on CRAN; thanks a lot, @rubak !

PR #1895, opened 2022-02-02
Update `pivot_longer()` signature

DavisVaughan:
In tidyr 1.2.0 we updated `pivot_longer()`'s `names_transform`, `values_transform`, `names_ptypes` and `values_ptypes` arguments to allow a _single_ function or ptype which will be applied to all columns, rather than requiring a named list. See https://github.com/tidyverse/tidyr/blob/main/NEWS.md#pivoting

This required us to change the default argument from `list()` to `NULL`. In particular, for `names/values_ptypes`, this was important because `list()` should now be interpreted as "use a list ptype for all columns".

For backwards compatibility with sf and a few other packages that extend `pivot_longer()`, we have temporarily forced `list()` to be interpreted as `NULL`, but we would like to remove this behavior. See https://github.com/tidyverse/tidyr/issues/1296.

This PR is a step towards that by updating the signature used in sf to use `NULL` rather than `list()`. We would appreciate if you could merge this in and get it on CRAN at your earliest convenience!

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -101,7 +101,7 @@ Suggests:
     terra,
     testthat,
     tibble (>= 1.4.1),
-    tidyr (>= 1.0-0),
+    tidyr (>= 1.2.0),
     tidyselect (>= 1.0.0),
     tmap (>= 2.0),
     vctrs,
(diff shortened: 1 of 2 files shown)


edzer:
Thanks, Davis!

PR #1896, opened 2022-02-03
CRS_from_crs() simplification

rsbivand:
(only edits in `R/sp.R`) In creating an `evolution` branch on my **sp** repo, I inadvertently created infinite recursion because of the way `CRS_from_crs()` was set up. This PR simplifies `CRS_from_crs()` by removing special casing.


edzer:
Thanks; merged manually (to avoid the test output mods)

PR #1911, opened 2022-03-02
Update tic templates [ci-skip]

edzer:
{tic} templates can be updated :rocket: :robot:


rsbivand:
Did you see this thread on Rtools42: https://stat.ethz.ch/pipermail/r-package-devel/2022q2/007975.html ?

edzer:
Thanks; we also see now that the [windows GA](https://github.com/r-spatial/sf/runs/6169412990?check_suite_focus=true) breaks on having R 4.2 but not the required / corresponding winlibs things.

PR #1919, opened 2022-03-15
Windows: update GDAL 3.4.1 for R < 4.2

jeroen:
Wrap-up GDAL update for R 3.3.0 - 4.1.3 on Windows. Also adds support for zstd tiff files. This update does not affect the new R-4.2 ucrt builds which you have hardcoded separately in [src/Makevars.ucrt](src/Makevars.ucrt).

### DIFF
--- a/src/Makevars.win
+++ b/src/Makevars.win
@@ -1,24 +1,22 @@
-VERSION = 3.2.1
+VERSION = 3.4.1
 RWINLIB = ../windows/gdal3-$(VERSION)
 TARGET = lib$(subst gcc,,$(COMPILED_BY))$(R_ARCH)
 
 PKG_CPPFLAGS =\
-	-I$(RWINLIB)/include/gdal-3.2.1 \
-	-I$(RWINLIB)/include/geos-3.9.0 \
-	-I$(RWINLIB)/include/proj-7.2.1 \
+	-I$(RWINLIB)/include \
 	-DHAVE_PROJ_H
 
 PKG_LIBS = \
 	-L$(RWINLIB)/$(TARGET) \
-	-L$(RWINLIB)/lib$(R_ARCH)$(CRT) \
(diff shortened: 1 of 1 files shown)


rsbivand:
Thanks! Curiously, 3.4.2 was just released. The big bump will be 3.6; 3.5 is coming soon and will have both cmake and configure, 3.6 drops configure, like GEOS 3.10 and PROJ 9.

jeroen:
OK great I had also tested this for rgdal and rgeos on all R versions. 

If there is interest I can keep providing updates in the future, but it looks like the new cran policies will mandate you use the gdal provided by TK instead of rwinlib for R 4.2 and up. So we'll see.

rsbivand:
This is my understanding too. Typically, migration to new R minor versions seems to have been slowing, so I expect 4.1 to be in active use for another year at least. I'm not aware that TK is planning to provide CRAN MXE UCRT support for `r-oldrel`, so up to the R 4.3 release, R 4.1 will be needed, I think. rwinlib has served the geospatial community very well!

edzer:
Thanks a lot, @jeroen !!

jeroen:
@rsbivand you committed a typo on rgdal: you commented out some of the old lines, but that does not work with makefiles: a line that ends with a `\`  will comment out the entire block:

Screenshot 2022-03-15 at 2 54 14 PM

Can you remove the commented lines from the Makevars.win, or alternatively move them out of the definition of PKG_LIBS and PKG_CPPFLAGS?

You can also inspect the build failure on r-universe: https://r-forge.r-universe.dev/

jeroen:
@rsbivand you may also benefit from this same change in rgeos because the new gdal stack includes geos 3.10.2.

rsbivand:
Thanks, **rgeos** seems to build and check fine on R-forge.

PR #1927, opened 2022-04-06
Don't strip whitespaces in RStudio project

kadyb:
If we edit project code in RStudio, the option to strip trailing whitespace is enabled by default. Currently, this generates large code differences. Disabling this option will make it easier to track changes in the code.


kadyb:
> StripTrailingWhitespace: Yes
> StripTrailingWhitespace: No

Not many changes, but they make a difference :smile:

PR #1928, opened 2022-04-06
Remove names in `st_sample()` output

kadyb:
This PR removes unnecessary coordinate names in the `st_sample()` so that the size of the returned object is about 40% smaller.


edzer:
Great idea, thanks!

Issue #1925, opened 2022-04-02
Low-hanging performance improvement opportunities


paleolimbot:
As hinted on Twitter and as per our discussion earlier, I think there are a few easy things that could speed up a number of things considerably! The main one is `st_sfc()`, which `vapply()`s along the list a few times. For a million points, the identity subset takes over two seconds:

``` r
library(sf)
#> Linking to GEOS 3.9.1, GDAL 3.2.3, PROJ 7.2.1; sf_use_s2() is TRUE
n <- 1e6
points <- data.frame(id = seq_len(n), x = runif(n), y = runif(n))
points_sf <- st_as_sf(points, coords = c("x", "y"))
system.time(points_sf$geometry[])
#>    user  system elapsed 
#>   2.009   0.070   2.078
```

The `vapply()` calls are here:

https://github.com/r-spatial/sf/blob/4b455d5edae1af1b65764636e0fbe8ecdcd1c3ed/R/sfc.R#L51

https://github.com/r-spatial/sf/blob/4b455d5edae1af1b65764636e0fbe8ecdcd1c3ed/R/sfc.R#L56

https://github.com/r-spatial/sf/blob/4b455d5edae1af1b65764636e0fbe8ecdcd1c3ed/R/sfc.R#L84

https://github.com/r-spatial/sf/blob/4b455d5edae1af1b65764636e0fbe8ecdcd1c3ed/R/sfc.R#L124

https://github.com/r-spatial/sf/blob/4b455d5edae1af1b65764636e0fbe8ecdcd1c3ed/R/sfc.R#L128

Given the amount of times `[.sfc` and `st_sfc()` get called, I think it's worth some C++ to calculate the sfc attributes in a single pass. I'm happy to contribute this if it would be welcome.

One other quality-of-life improvement would be removing the `vapply()` in `print.sfc()`, which takes about a second for the above points example:

https://github.com/r-spatial/sf/blob/4b455d5edae1af1b65764636e0fbe8ecdcd1c3ed/R/sfc.R#L199

It seems like that information is already in the `class()` anyway? I'm also happy to contribute that fix.

edzer:
This was a little more involved than I'd hoped! The searching for `NULL` values is still needed, but dropping the need for looking for `NA`s in the list (generated by `merge`) helps quite a bit; this is now obsolete by the added `is.na<-.sfc` method which `merge.data.frame()` calls.

paleolimbot:
Awesome!

[...]

PR #1938, opened 2022-05-02
Small optimizations for `st_sfc()`

paleolimbot:
Following discussion on #1925, this PR eliminates `vapply()`s in most cases when constructing or subsetting, which mostly happens in `st_sfc()`. The most dramatic speedup is on points, where calculating the bounding box was very slow, but it looks like the speedup for polygons is about the same (about 8x faster, for both). 

I tried to only swap out the vapplys to avoid accidentally changing the structure of sfc objects! In #1925 @edzer experimented with a few ways to skip some of these checks, which may help even more. That said, I think this PR reduces the subset/construction overhead to the point where it's not noticeable compared to other things (like actual computations/reading from disk).

Current dev:

``` r
library(sf)
#> Linking to GEOS 3.10.1, GDAL 3.3.3, PROJ 8.1.0; sf_use_s2() is TRUE
n <- 1e6
points <- data.frame(id = seq_len(n), x = runif(n), y = runif(n))
points_sf <- st_as_sf(points, coords = c("x", "y"))
bench::mark(
  points_sf$geometry[],
  check = F,
  iterations = 10
)
#> Warning: Some expressions had a GC in every iteration; so filtering is disabled.
#> # A tibble: 1 × 6
#>   expression                min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>           <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 points_sf$geometry[]    1.67s    1.76s     0.563     123MB     2.76
```

``` r
library(sf)
#> Linking to GEOS 3.10.1, GDAL 3.3.3, PROJ 8.1.0; sf_use_s2() is TRUE
nc <- read_sf(system.file("shape/nc.shp", package = "sf"))
nc_big <- rep(nc$geometry, 1000)
bench::mark(nc_big[], check = FALSE)
#> Warning: Some expressions had a GC in every iteration; so filtering is disabled.
#> # A tibble: 1 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 nc_big[]      333ms    350ms      2.85    54.8MB     9.99
```

After this PR:

``` r
library(sf)
#> Linking to GEOS 3.10.1, GDAL 3.3.3, PROJ 8.1.0; sf_use_s2() is TRUE
n <- 1e6
points <- data.frame(id = seq_len(n), x = runif(n), y = runif(n))
points_sf <- st_as_sf(points, coords = c("x", "y"))
bench::mark(
  points_sf$geometry[],
  check = F,
  iterations = 10
)
#> Warning: Some expressions had a GC in every iteration; so filtering is disabled.
#> # A tibble: 1 × 6
#>   expression                min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>           <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 points_sf$geometry[]    169ms    247ms      3.25    61.3MB     4.87
```

``` r
library(sf)
#> Linking to GEOS 3.10.1, GDAL 3.3.3, PROJ 8.1.0; sf_use_s2() is TRUE
nc <- read_sf(system.file("shape/nc.shp", package = "sf"))
nc_big <- rep(nc$geometry, 1000)
bench::mark(nc_big[], check = FALSE)
#> # A tibble: 1 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 nc_big[]     41.5ms   41.5ms      24.1    48.2MB     145.
```


edzer:
Thanks a lot, @paleolimbot !

PR #1963, opened 2022-06-20
implement filter, slice, arrange, and mutate through dplyr_col_modify and dplyr_row_slice

huizezhang-sherry:
The four tidyverse verbs have been implemented through `dplyr_col_modify` and `dplyr_row_slice`, to replace the original `xxx.sf` methods.

`select` and `transmute` are not implemented in such a way since it would require changing on `[.sf`.


huizezhang-sherry:
Any idea why the ubuntu-latest(devel) checks fail? I'm not familiar with this error msg: 

> Error: Error: getaddrinfo EAI_AGAIN api.r-hub.io

edzer:
That seems a false positive (CI problem, not a problem of this PR)

PR #1975, opened 2022-07-17
Tidyr drop na

gregleleu:
Adding drop_na to the tidyverse functions for sf objects


edzer:
Thanks - makes sense!

Issue #2005, opened 2022-10-05
st_interpolate_aw returns error - replacement has x rows, data has y


phcantis:
I am working to re-aggregate population counts from one administrative level to another. One particular row keeps returning the error "replacement has 4 rows, data has 2". I have tried validating the geometries, but it has not worked. 

Even better, I was preparing the dummy data for postin this question at StackOverflow (which I didnt do in the end because of an excess in characters due to using dput()). After saving the dput() of the sf layers into a text file, the command line works after re-assigning the variables. This only works when using the dput output saved in a textfile, but it does not work if I directly do 
`st_interpolate_aw(dput(geography1),dput(geography2), extensive = TRUE)`.  

This returns the same error.

Is there any known reason why the error might be happening, and how **saving** the dput() output into a .txt file might be removing the error? I have tried saving to a geojson file and then reloading, but the error remains. I have also tried using st_buffer(0) and st_make_valid(). Nothing worked so far.

I am attaching the two files with the output of dput() that were giving me the error, but since I have saved them, they will probably not reproduce it.

These [.geojson](https://drive.google.com/file/d/1wF6AB1oHVEkcuI4iRAloLO56DL_8EZBr/view?usp=sharing) files might reproduce the error when running
`st_interpolate_aw(SA1_suburb, suburb, extensive = TRUE)`

[SA1_suburb.txt](https://github.com/r-spatial/sf/files/9716975/SA1_suburb.txt)
[suburb.txt](https://github.com/r-spatial/sf/files/9716977/suburb.txt)

phcantis:
As an update - I got to share the geojson files in StackOverflow and somebody pointed that it might be an issue related to S2. See link:
https://stackoverflow.com/questions/73962700/st-interpolate-aw-returns-error-replacement-has-x-rows-data-has-y/73963282#73963282

kadyb:
> Is there any known reason why the error might be happening, and how saving the dput() output into a .txt file might be removing the error?

I suspect that when saving using `dput()` you lost the precision of the coordinates and that solved this problem.

```r
SA1_suburb = read_sf("SA1_suburb_1221.geojson")
print(st_coordinates(SA1_suburb)[1, 1:2], digits = 18)
#> 142.7316964610000696 -36.5496727679999367
tmp = tempfile()
dput(SA1_suburb, tmp)
SA1_suburb = dget(tmp)
print(st_coordinates(SA1_suburb)[1, 1:2], digits = 18)
[...]

PR #2011, opened 2022-10-11
another shot at #2005 - asking s2 to ignore line & point intersections

jlacko:
I believe this pull request resolves the underlying issue behind #2005  better than (now closed) #2006 

If we ask S2 to return only polygon intersections - which is possible, I just did not realise it when composing the previous pull - it will not be necessary to do any filtering.

In case of planar intersection the `dimensions = "polygon"` part of the intersection call would be safely ignored.

### DIFF
--- a/R/aggregate.R
+++ b/R/aggregate.R
@@ -122,7 +122,7 @@ st_interpolate_aw.sf = function(x, to, extensive, ..., keep_NA = FALSE) {
 
 	if (! all_constant(x))
 		warning("st_interpolate_aw assumes attributes are constant or uniform over areas of x")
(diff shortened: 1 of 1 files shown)


edzer:
Wow - the power of `...` !

jlacko:
quite amazing, actually... 
> Wow - the power of `...` !
I first got the impression that the entire function would need a revision because of the way `{s2}` handles line & point intersection differently from original GEOS and the `idx` concept was no longer reliable; but at the end the `s2_options` may offer a solution. 
This code passes the tricky geojsons mentioned in #2006 and tests I could think of using standard `nc.shp` dataset, both in S2 and GEOS mode.

Issue #2024, opened 2022-10-31
Faster Convex Hull


kadyb:
I just found that there is `grDevices::chull()` function in R for convex hull computing and it seems to be pretty fast compared to `sf::st_convex_hull()` for planar CRS using GEOS. This may be due to the fact that `chull()` uses different algorithm proposed by Eddy (1977). I think it's worth mentioning this function in the `{sf}` documentation. **Edit:** I proposed PR #2025.

```r
library("sf")
set.seed(1)

n = 5000000
df = data.frame(x = rnorm(n), y = rnorm(n))
pts = st_as_sf(df, coords = c("x", "y"), crs = 2180) # planar CRS
pts = st_combine(pts)
system.time({ch_sf = st_convex_hull(pts)}) # uses GEOS
#>  user  system elapsed
#> 11.42    0.33   11.77
plot(ch_sf)

system.time({
  ch_r = chull(df)
  ch_r = st_as_sf(df[ch_r, ], coords = c("x", "y"), crs = 2180)
  ch_r = st_cast(st_combine(ch_r), "POLYGON")
})
#> user  system elapsed
#> 0.17    0.01    0.18
plot(ch_r)
```

PR #2025, opened 2022-10-31
Mention `grDevices::chull()` in docs (#2024)

kadyb:


edzer:
Thanks!

kadyb:
Sorry, this link for `chull()` is still not properly generated (https://r-spatial.github.io/sf/reference/geos_unary.html#see-also-1).
This should work as [documented](https://roxygen2.r-lib.org/articles/rd-formatting.html#default-link-text), but maybe we should remove `grDevices`? Just write: `[chull()]`

Issue #2033, opened 2022-11-09
8-bit int for GDAL rasters coming


rsbivand:
Heads-up: https://github.com/OSGeo/gdal/pull/6634

edzer:
Future proof!

rsbivand:
This will be a divergence for rgdal, which I don't intend to patch.

PR #2036, opened 2022-11-10
Experimental support for GDAL 3.6 columnar API

paleolimbot:
This PR is a proof-of-concept for future support for the new columnar access API that is introduced in GDAL 3.6. The API exposes a pull-style iterator as an [`ArrowArrayStream`](https://arrow.apache.org/docs/format/CStreamInterface.html). The pyogrio package has a PR up to support this and has noticed a ~2x improvement on their test data set ( https://github.com/geopandas/pyogrio/pull/155 ). I've spent some time implementing conversions for Arrow C data interface objects ( https://github.com/apache/arrow-nanoarrow/pull/65 ) and am curious to see if we can get any of that speed in R too!

To give it a try:

```
# (Requires development GDAL!)
# remotes::install_github("apache/arrow-nanoarrow/r#65")
# remotes::install_github("paleolimbot/sf@stream-reading")
library(sf)
#> Linking to GEOS 3.8.0, GDAL 3.7.0dev-908498a4d8, PROJ 6.3.1; sf_use_s2() is
TRUE
read_sf(system.file("shape/nc.shp", package = "sf"), use_stream = TRUE)
#> Simple feature collection with 100 features and 15 fields
#> Geometry type: GEOMETRY
#> Dimension:     XY
#> Bounding box:  xmin: -84.32385 ymin: 33.88199 xmax: -75.45698 ymax: 36.58965
#> Geodetic CRS:  NAD27
#> # A tibble: 100 × 16
#>    OGC_FID  AREA PERIMETER CNTY_ CNTY_ID NAME   FIPS  FIPSNO CRESS…¹ BIR74 SID74
#>      <dbl> <dbl>     <dbl> <dbl>   <dbl> <chr>  <chr>  <dbl>   <int> <dbl> <dbl>
#>  1       0 0.114      1.44  1825    1825 Ashe   37009  37009       5  1091     1
#>  2       1 0.061      1.23  1827    1827 Alleg… 37005  37005       3   487     0
#>  3       2 0.143      1.63  1828    1828 Surry  37171  37171      86  3188     5
#>  4       3 0.07       2.97  1831    1831 Curri… 37053  37053      27   508     1
#>  5       4 0.153      2.21  1832    1832 North… 37131  37131      66  1421     9
#>  6       5 0.097      1.67  1833    1833 Hertf… 37091  37091      46  1452     7
#>  7       6 0.062      1.55  1834    1834 Camden 37029  37029      15   286     0
#>  8       7 0.091      1.28  1835    1835 Gates  37073  37073      37   420     0
#>  9       8 0.118      1.42  1836    1836 Warren 37185  37185      93   968     4
#> 10       9 0.124      1.43  1837    1837 Stokes 37169  37169      85  1612     1
#> # … with 90 more rows, 5 more variables: NWBIR74 <dbl>, BIR79 <dbl>,
#> #   SID79 <dbl>, NWBIR79 <dbl>, wkb_geometry <GEOMETRY [°]>, and abbreviated
#> #   variable name ¹​CRESS_ID
#> # ℹ Use `print(n = ...)` to see more rows, and `colnames()` to see all variable names
```

All sorts of options aren't supported but I'm mostly trying to find out whether or not the nanoarrow conversions are going to be helpful in this context. I will follow up here with more experiments!

### DIFF
--- a/src/gdal_read_stream.cpp
+++ b/src/gdal_read_stream.cpp
@@ -106,7 +106,14 @@ Rcpp::List CPL_read_gdal_stream(
     }
 
     GDALStreamWrapper::Make(&stream_temp, prep, stream_out);
-    double num_features = (double) poLayer->GetFeatureCount(false);
+
+    // The reported feature count is incorrect if there is a query
+    double num_features;
+    if (query.size() == 0) {
+        num_features = (double) poLayer->GetFeatureCount(false);
+    } else {
+        num_features = -1;
+    }
 
     return Rcpp::List::create(wkt_str, Rcpp::NumericVector::create(num_features));
 }
--- a/src/gdal_read_stream.cpp
+++ b/src/gdal_read_stream.cpp
@@ -79,7 +79,7 @@ Rcpp::List CPL_read_gdal_stream(
 		bool dsn_isdb,
         Rcpp::CharacterVector fid_column,
 		int width) {
-    
+
     const char* array_stream_options[] = {"INCLUDE_FID=NO", nullptr};
     if (fid_column.size() == 1) {
         array_stream_options[0] = "INCLUDE_FID=YES";
@@ -89,7 +89,6 @@ Rcpp::List CPL_read_gdal_stream(
 							quiet,  drivers,
 							wkt_filter,
 							dsn_exists, dsn_isdb, width);
-    OGRDataSource* poDS = (OGRDataSource*)(R_ExternalPtrAddr(prep[0]));
 	OGRLayer* poLayer = (OGRLayer*)R_ExternalPtrAddr(prep[1]);
     auto stream_out = reinterpret_cast<struct ArrowArrayStream*>(
         R_ExternalPtrAddr(stream_xptr));
--- a/src/gdal_read_stream.cpp
+++ b/src/gdal_read_stream.cpp
@@ -69,7 +69,7 @@ private:
 };
 
 // [[Rcpp::export]]
-Rcpp::CharacterVector CPL_read_gdal_stream(
+Rcpp::List CPL_read_gdal_stream(
         Rcpp::RObject stream_xptr,
         Rcpp::CharacterVector datasource, Rcpp::CharacterVector layer,
 		Rcpp::CharacterVector query,
@@ -99,7 +99,9 @@ Rcpp::CharacterVector CPL_read_gdal_stream(
     }
 
     GDALStreamWrapper::Make(&stream_temp, prep, stream_out);
-    return Rcpp::CharacterVector::create(wkt_str);
+    double num_features = (double) poLayer->GetFeatureCount(false);
+
+    return Rcpp::List::create(wkt_str, Rcpp::NumericVector::create(num_features));
 }
 
 #else
--- a/src/gdal_read_stream.cpp
+++ b/src/gdal_read_stream.cpp
@@ -68,6 +68,8 @@ private:
     }
 };
 
+#endif
+
 // [[Rcpp::export]]
 Rcpp::List CPL_read_gdal_stream(
         Rcpp::RObject stream_xptr,
@@ -79,6 +81,8 @@ Rcpp::List CPL_read_gdal_stream(
 		bool dsn_isdb,
         Rcpp::CharacterVector fid_column,
 		int width) {
+
+#if GDAL_VERSION_NUM >= GDAL_COMPUTE_VERSION(3,6,0)
 
     const char* array_stream_options[] = {"INCLUDE_FID=NO", nullptr};
     if (fid_column.size() == 1) {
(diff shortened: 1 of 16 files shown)


rsbivand:
Thanks, interesting! Is it feasible to use libgdal without needing **nanoarrow**? I think https://gdal.org/development/rfc/rfc86_column_oriented_api.html implements locally without requiring access to Arrow itself?

src/stars.cpp
```diff
@@ -509,7 +509,7 @@ void CPL_write_gdal(NumericMatrix x, CharacterVector fname, CharacterVector driv
 		eType = GDT_Byte; // #nocov
 #if GDAL_VERSION_NUM >= 3070000
 	else if (Type[0] == "Int8")
-		eType = GDT_Int8; // #nocov
+		eType = GDT_Byte; // #nocov
```
edzer, line 512: This doesn't look good IMO, see https://github.com/r-spatial/sf/issues/2033 - do you actually run into this code? That would mean the #if condition is wrong.

paleolimbot: I'm sure it's wrong...but yes, this failed to compile for me with `GDT_Int8`.

rsbivand:
Is @edzer 's point to do with WIP RFC87 https://github.com/rspatial/terra/issues/885#issuecomment-1309213171?

edzer:
We can also outcomment the whole `Int8` section and wait until that moves in GDAL.

src/stars.cpp
paleolimbot: I fixed this - it's because I'm running from the latest dev branch which has moved on to 3.7.xx versioning because there's a release candidate.

rouault: GDT_Int8 implementation has now landed in GDAL master

edzer:
Looks great - trains, airports and plains are always the best place for some decent development!

src/gdal_read_stream.cpp
```diff
@@ -0,0 +1,133 @@
+
+#include <ogrsf_frmts.h>
+
+#define RCPP_DEFAULT_INCLUDE_CALL false
+#include "Rcpp.h"
+
+#if (GDAL_VERSION_MAJOR >= 3 && GDAL_VERSION_MINOR >= 6)
```
rouault, line 7: would break for GDAL 4.0:

```suggestion
#if GDAL_VERSION_NUM >= GDAL_COMPUTE_VERSION(3,6,0)
```

mdsumner:
also dialect for when running sql, that's just a string input doesn't have any layer cleanup implications (so can run SQLITE st_funs on OGRSQL drivers, or vice versa, for example)

kadyb:
Thanks, great to see this feature! Did you try do some benchmarks on synthetic data? In the example below, the differences in timings are quite small. What could be the reason?

<details>
<summary>Benchmark</summary>

```r
library("sf")

n = 500000
df = data.frame(x = rnorm(n), y = rnorm(n),
                col_1 = sample(c(TRUE, FALSE), n, replace = TRUE), # logical
                col_2 = sample(letters, n, replace = TRUE),        # character
                col_3 = runif(n),                                  # double
                col_4 = sample(1:100, n, replace = TRUE))          # integer

## points ##
pts = st_as_sf(df, coords = c("x", "y"), crs = "EPSG:2180")
write_sf(pts, "pts.gpkg")

bench::mark(
  check = FALSE, iterations = 10,
  stream = read_sf("pts.gpkg", use_stream = TRUE),
  non_stream = read_sf("pts.gpkg", use_stream = FALSE)
)
#>   expression      min   median `itr/sec` mem_alloc
#> 1 stream         1.8s    2.14s     0.491    87.8MB
#> 2 non_stream     1.7s    1.79s     0.545    72.5MB

## buffers ##
buff = st_buffer(pts, dist = 1000)
write_sf(buff, "buffers.gpkg")

bench::mark(
  check = FALSE, iterations = 5,
  stream = read_sf("buffers.gpkg", use_stream = TRUE),
  non_stream = read_sf("buffers.gpkg", use_stream = FALSE)
)
#>   expression      min   median `itr/sec` mem_alloc
#> 1 stream        4.22s    5.97s     0.181    1.93GB
#> 2 non_stream    4.21s    6.23s     0.182    1.91GB
```

</details>

kadyb:
This doesn't work for me:

```r
n = 10
df = data.frame(x = rnorm(n), y = rnorm(n)) # without attributes
pts = st_as_sf(df, coords = c("x", "y"), crs = "EPSG:2180")
write_sf(pts, "pts.gpkg")
x = read_sf("pts.gpkg", use_stream = TRUE)
#> Error in st_as_sfc.WKB(x) : 
#>   cannot read WKB object from zero-length raw vector
```

kadyb:
Good to hear, I will wait for further news! In which repository should I create the issue?

edzer:
> In which repository should I create the issue?

here.

kadyb:
Here is an interesting comparison of vector data loading performance in `sf` and `gdalraster` by @ctoney: https://firelab.github.io/gdalraster/articles/vector-read-benchmarks.html

Issue #2008, opened 2022-10-06
`st_distance` `by_element=TRUE` slow depending on projection 


bart1:
Ths speed of `st_distance` is very slow depending on the projection:
``` r
require(sf)
#> Loading required package: sf
#> Linking to GEOS 3.8.0, GDAL 3.0.4, PROJ 6.3.1; sf_use_s2() is TRUE
x<-st_as_sf(data.frame(a=1:10000,x=runif(10000),y=runif(10000)),coords=c('x','y'), crs=st_crs("+proj=aeqd +units=km"))
system.time(a<-st_distance(x[-1,],x[-nrow(x),], by_element = T))
#>    user  system elapsed 
#>   5.543   0.023   5.566
xx<-x |> st_transform(st_crs(4326))         
system.time(b<-st_distance(xx[-1,],xx[-nrow(xx),], by_element = T))
#>    user  system elapsed 
#>   0.152   0.000   0.153
cor(a,b)
#> [1] 0.9999859
system.time(st_distance(x[1:100,]))
#>    user  system elapsed 
#>   0.038   0.000   0.038
system.time(st_distance(xx[1:100,]))
#>    user  system elapsed 
#>   0.025   0.000   0.025
```

<sup>Created on 2022-10-06 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Here the distance calculation is more then 40 times slower when a non longlat projection is used. The difference does only occur when using `by_element=T`. I think this is due to the recursive call for each element here: https://github.com/r-spatial/sf/blob/9f4574bcbd0c4b4a3a71a65b4b75f9180d7034a2/R/geom-measures.R#L172 . I'm not sure its avoidable, but it would be more elegant if the speed were more comparable.

<details>
``` r
require(sf)
#> Loading required package: sf
#> Linking to GEOS 3.8.0, GDAL 3.0.4, PROJ 6.3.1; sf_use_s2() is TRUE
sessionInfo()
#> R version 4.2.1 (2022-06-23)
#> Platform: x86_64-pc-linux-gnu (64-bit)
#> Running under: Ubuntu 20.04.5 LTS
#> 
#> Matrix products: default
#> BLAS:   /usr/lib/x86_64-linux-gnu/atlas/libblas.so.3.10.3
#> LAPACK: /usr/lib/x86_64-linux-gnu/atlas/liblapack.so.3.10.3
#> 
#> locale:
[...]

Issue #1879, opened 2022-01-10
Don't return names in `st_sample()`?


kadyb:
What do you think about not returning the `"lon"`, `"lat"`, `"x"`, `"y"` names in `st_sample()` to reduce object size? I suspect they are not needed, but please correct me if I am wrong. 

```r
library("sf")

nc = read_sf(system.file("shape/nc.shp", package = "sf"))
size = 10000

smp = st_sample(nc, size)
format(object.size(smp), units = "auto")
#> "6.9 Mb"

## remove names to reduce object size
for (i in seq_along(smp)) names(smp[[i]]) = NULL
format(object.size(smp), units = "auto")
#> "4.1 Mb"
```

These lines are relevant:
https://github.com/r-spatial/sf/blob/9f4574bcbd0c4b4a3a71a65b4b75f9180d7034a2/R/sample.R#L182
https://github.com/r-spatial/sf/blob/9f4574bcbd0c4b4a3a71a65b4b75f9180d7034a2/R/sample.R#L251
so to correct this, just use `colnames() = NULL` in next lines.

PR #2038, opened 2022-11-12
drop the rownames in `st_coordinates()`

kadyb:
This will improve https://github.com/r-spatial/sf/issues/2008#issuecomment-1309530848 and https://github.com/r-spatial/sf/issues/1879#issuecomment-1009185571. I ran the local and GHA tests and nothing seems to break. With this PR, I see a reduced output size of ~9 times in `st_distance(by_element = TRUE, which = "Euclidean")` and ~5 times in `st_coordinates()`.

```r
n = 10000
df1 = st_as_sf(data.frame(x = runif(n), y = runif(n)), coords = 1:2, crs = 3857)
df2 = st_as_sf(data.frame(x = runif(n), y = runif(n)), coords = 1:2, crs = 3857)
dist = st_distance(df1, df2, by_element = TRUE, which = "Euclidean")
format(object.size(x), units = "auto")
format(object.size(st_coordinates(df1)), units = "auto")
```

| **Function**      | **Before [Kb]** | **After [kB]** |
|-------------------|-----------------|----------------|
| `st_distance`    | 704.4           | 79.2           |
| `st_coordinates` | 781.9           | 156.8          |

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # version 1.0-10
 
+* drop the rownames in `st_coordinates()` reducing the output size
+
 # version 1.0-9
 
 * adjust for changes how R-devel handles `POSIXlt`; #2028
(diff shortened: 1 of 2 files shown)


edzer:
Great, thanks!

PR #2047, opened 2022-11-24
Ignores the `.keep` of `group_split.sf()` for rowwise_df

UchidaMizuki:
`group_split.sf()` for rowwise_df warns that `.keep` is ignored as follows, 
so I modified the behavior of `group_split.sf()` for rowwise_df.

``` r
suppressMessages(require(sf, quietly = TRUE))
suppressMessages(require(dplyr, quietly = TRUE))

nc <- st_read(system.file("shape/nc.shp", package="sf"), quiet = TRUE)

out <- nc %>%
  rowwise() %>%
  group_split()
#> Warning: .keep is ignored in group_split(<rowwise_df>)
```

<sup>Created on 2022-11-24 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/tidyverse.R
+++ b/R/tidyverse.R
@@ -25,7 +25,11 @@ dplyr_reconstruct.sf = function(data, template) {
 
 group_split.sf <- function(.tbl, ..., .keep = TRUE) {
 	 class(.tbl) = setdiff(class(.tbl), "sf")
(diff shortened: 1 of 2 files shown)


edzer:
Thanks!

Issue #2063, opened 2022-12-16
sf/R/sp.R: process_pl_comment() failing - replace rgeos::createSPComment etc.


rsbivand:
https://github.com/r-spatial/evolution/issues/9 https://github.com/rstudio/leaflet/issues/833

Something is de-railing in sf/R/sp.R: process_pl_comment()  in st_as_sfc.SpatialPolygons()  ... unsure how to proceed.

PR #2069, opened 2022-12-21
for #2063; reproducing rgeos workflow for assigning interior to exterior rings

rsbivand:
https://github.com/r-spatial/evolution/issues/9, https://github.com/rstudio/leaflet/issues/833

This is not optimized at all, but since it is unlikely to be entered unless someone has canned, stale **sp** polygon objects in their workflows, speed doesn't matter. Here the workhorse is: `st_contains_properly(raw0[!holes], raw0[holes])`, where `holes` are got from the input object. If no holes were assigned initially, the flow of control reverts effectively to the previous code. `st_contains_properly()` may fail if a quasi-hole shares a border segment with its containing exterior ring. Holes are assigned as interior rings to  exterior rings in order of exterior ring area from smallest to largest (the nested island/holes case), and assigned holes the dropped from the larger exterior rings.

### DIFF
--- a/R/sp.R
+++ b/R/sp.R
@@ -137,6 +137,7 @@ st_as_sfc.SpatialLines = function(x, ..., precision = 0.0, forceMulti = FALSE) {
 
 #' @name st_as_sfc
 #' @export
+
 st_as_sfc.SpatialPolygons = function(x, ..., precision = 0.0, forceMulti = FALSE) {
 	lst = if (forceMulti || any(sapply(x@polygons, function(x) moreThanOneOuterRing(x@Polygons)))) {
 		if (is.null(comment(x)) || comment(x) == "FALSE") {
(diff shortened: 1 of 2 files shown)


edzer:
According to the simple feature standard, holes do not share a border segment with the exterior ring. Does this PR bring in non-SF logic into the package? What speaks against making such objects (SF-)valid at this point?

rsbivand:
I'll check and report back - `st_contains_properly()` checks that all of B is in the interior of A, so perhaps using `st_contains()` would catch the case. Now I think that such a hole would warn that no exterior ring could be found, but if  `st_contains()` was used, it would assign the interior ring to the correct exterior ring, and `st_make_valid()` used later should restore SF conformance.

rsbivand:
With `st_contains()` instead of `st_contains_properly()`, no warning of orphaned hole and the hole on the edge of the central island removed and the edge notched:
image
Will update the PR; in neither case would coercion to `"sf"` create an invalid object, but some geometry details would have been lost with `st_contains_properly()`.

edzer:
Great, so this is ready to go, right?

rsbivand:
Yes, as far as I can see. Should resolve https://github.com/r-spatial/evolution/issues/9,  and https://github.com/rstudio/leaflet/issues/833, where `rgeos::createPolygonsComment()` isn't exactly patched by `as(sf::st_as_sfc(x), "Spatial")` because `x` is a `"Spatial*"`, so this round trip corresponds to  `rgeos::createSPComment()`. I don't think it is necessary to create coercions from `"Polygons"` to `"sfg"` and back, but https://github.com/rstudio/leaflet/blob/51daa695419fe2bd9bab3d9fadf69d59bbf795ee/R/normalize-sp.R#L85-L109 might well be re-configured to coerce input `"Spatial*"` objects into `"sf"` objects then use `normalize-sf.R`, probably for all the vector objects.

edzer:
Strange enough this passes locally, but fails on windows, see [here](https://win-builder.r-project.org/qcMfSWiljzeG/00check.log). Is it OK if I outcomment the test for the 1.0-10 release?

rsbivand:
Which **sp** version were you using? Winbuilder is on Rtools43, but I'm sure with released **sp**. My local testing was on Fedora with **sp** fixed for the `main()` problem - do we also need a release https://cran.r-project.org/web/checks/check_results_roger.bivand_at_nhh.no.html to remove the rgdal and rgeos warnings. I'll submit rgdal shortly with updated src/Makevars.ucrt for the Rtools43 error.

rsbivand:
Did you also update to: https://svn.r-project.org/R-dev-web/trunk/WindowsBuilds/winutf8/ucrt3/r_packages/patches/CRAN/sf.diff for GDAL 3.6.2 ? I think Win-builder has an intermediate version with older GEOS - maybe a cause? Trying now ...

rsbivand:
With R Under development (unstable) (2023-01-03 r83550 ucrt), matching Rtools with GDAL 3.6.2 and GEOS 3.11.1:

[00check.log](https://github.com/r-spatial/sf/files/10362640/00check.log)
[00install.zip](https://github.com/r-spatial/sf/files/10362643/00install.zip)
[testthat.zip](https://github.com/r-spatial/sf/files/10362653/testthat.zip)

rsbivand:
With released CRAN **sf** and GEOS 3.9.3, I see (in debug and after pasting in the new process_pl_comment()):
```
>   spls_rt <- as(spls_sfc, "Spatial")
> comment(slot(spls_rt, "polygons")[[1]])
[1] "0 1 1 1"
> sf_extSoftVersion()
          GEOS           GDAL         proj.4 GDAL_with_GEOS     USE_PROJ_H 
       "3.9.3"        "3.5.2"        "8.2.1"         "true"         "true" 
          PROJ 
       "8.2.1" 
```
so maybe a dependency on GEOS versions later than 3.9.3? Shall I try to establish which? Should we condition this part of `test_sp.R` on say GEOS >= 3.11?

rsbivand:
Confirmed that with GEOS 3.9.4 on Fedora, `st_make_valid()` returns two `GEOMETRYCOLLECTION` features and four `POLYGON` features, with 3-.11.1, six `POLYGON` features.

PR #2079, opened 2023-01-09
Coerce comments

rsbivand:
This should work with older GEOS; not checked for `lwgeom` for #2069 . Looks like  

> CAPI: GEOSMakeValidWithParams new validity enforcement approach from https://github.com/locationtech/jts/pull/704, uses GeometryFixer (Paul Ramsey, Martin Davis)

https://github.com/libgeos/geos/blob/3.11.1/NEWS.md in 3.10.0.


rsbivand:
Maybe comment out https://github.com/r-spatial/sf/blob/5334d9c367341ad2d48e53e6aff7adbd6652eafb/tests/testthat/test_sp.R#L27, as with the fix the test also passes on earlier GEOS?

PR #2084, opened 2023-01-13
Reenable CI/CD on Windows

krlmlr:
`autoconf.ac` is precious, needs to be checked out with LF style.


edzer:
Thanks! The process still breaks, but on another issue now.

Issue #2062, opened 2022-12-15
st_distance does not recycle y when by_element = TRUE in 1.0.9


majr-red:
From [the `st_distance` documentation](https://r-spatial.github.io/sf/reference/geos_measures.html):

>If by_element is FALSE
>st_distance returns a dense numeric matrix of dimension length(x) by length(y); <i>otherwise it returns a numeric vector of length x or y, the shorter one being recycled.</i>

(my italics) This was the behaviour in 1.0.8, but in 1.0.9 if `by_element = TRUE` then supplying unequal length x and y arguments produces an error. This appears to be because v1.0.9 includes:

```R	
if (by_element)
		stopifnot(!missing_y, length(x) == length(y))
```
Is this now the desired behaviour (and so the documentation needs updating), or a bug?

``` r
# remotes::install_version("sf", version = "1.0.9",
#                          dependencies = "Depends",
#                          upgrade = "never"
# )

library(sf)
#> Linking to GEOS 3.10.2, GDAL 3.4.1, PROJ 7.2.1; sf_use_s2() is TRUE
packageVersion("sf")
#> [1] '1.0.9'

pt_a <- st_point( c(0, 1) )
pt_b <- st_point( c(1, 0) )

x <- st_sf(geom = st_sfc(pt_a, pt_b),
           placename = c("a", "b"))

y <- st_sf(geom = st_sfc( st_point( c(2, 3) )),
           placename = c("c"))

st_distance(x, y, by_element=TRUE)
#> Error in st_distance(x, y, by_element = TRUE): length(x) == length(y) is not TRUE

sessionInfo()
#> R version 4.1.2 (2021-11-01)
#> Platform: x86_64-w64-mingw32/x64 (64-bit)
#> Running under: Windows Server x64 (build 14393)
#> 
#> Matrix products: default
#> 
#> locale:
#> [1] LC_COLLATE=English_United Kingdom.1252 
#> [2] LC_CTYPE=English_United Kingdom.1252   
#> [3] LC_MONETARY=English_United Kingdom.1252
#> [4] LC_NUMERIC=C                           
#> [5] LC_TIME=English_United Kingdom.1252    
#> 
#> attached base packages:
#> [1] stats     graphics  grDevices utils     datasets  methods   base     
#> 
#> other attached packages:
#> [1] sf_1.0-9
#> 
#> loaded via a namespace (and not attached):
[...]

PR #2106, opened 2023-02-24
Update docs for st_distance

majr-red:
Update docs for `st_distance()` to reflect behaviour of `by_element` parameter since v1.0.9 (#2062)


edzer:
Thanks!

PR #2108, opened 2023-03-01
Update Homebrew instructions for Mac on Apple Silicon

### LINKED ISSUE #1982: configure error on M1 Mac (12.5) w/ homebrew libs
**Describe the bug**
Attempting to install `sf` (and `terra`) I get the following error from `configure`:

```
~/tmp/sf » R CMD install .                                                                                                                                            rundel@magsafe
* installing to library ‘/Users/rundel/Library/R/arm64/4.2/library’
* installing *source* package ‘sf’ ...
file ‘configure’ has the wrong MD5 checksum
** using staged installation
configure: CC: clang
configure: CXX: clang++ -std=gnu++11
checking for gdal-config... /opt/homebrew/bin/gdal-config
checking gdal-config usability... yes
configure: GDAL: 3.5.1
checking GDAL version >= 2.0.1... yes
checking for gcc... clang
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables...
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether the compiler supports GNU C... yes
checking whether clang accepts -g... yes
checking for clang option to enable C11 features... none needed
checking for stdio.h... yes
checking for stdlib.h... yes
checking for string.h... yes
checking for inttypes.h... yes
checking for stdint.h... yes
checking for strings.h... yes
checking for sys/stat.h... yes
checking for sys/types.h... yes
checking for unistd.h... yes
checking for gdal.h... yes
checking GDAL: linking with --libs only... yes
checking GDAL: /opt/homebrew/Cellar/gdal/3.5.1_1/share/gdal/pcs.csv readable... no
checking GDAL: checking whether PROJ is available for linking:... yes
checking GDAL: checking whether PROJ is available fur running:... yes
configure: GDAL: 3.5.1
configure: pkg-config proj exists, will use it
configure: using proj.h.
configure: PROJ: 9.0.1
checking PROJ: checking whether PROJ and sqlite3 are available for linking:... no
configure: error: libproj or sqlite3 not found in standard or given locations.
ERROR: configuration failed for package ‘sf’
```

The specific issue is caused by the inclusion of `-ljpeg` in `${LIBS}` without the inclusion of the necessary `-L` flag to find this library, so the following error is reported:

```
~/tmp/sf » more errors.txt                                                                                                                                        1 ↵ rundel@magsafe
ld: library not found for -ljpeg
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

This seems in part related to how homebrew has configured `pkg-config` which is pulling this specific library in due to proj -> libtiff-4 -> jpeg with libtiff bringing in the `-l` flag and not the `-L` flag. While this could be take up with the homebrew folks this seems like an unnecessary use of the `--static` flag with pkg-config for proj - see https://github.com/r-spatial/sf/blob/e0eeaa1e70522f4a7cdb439909e6151dbae5b029/configure.ac#L398

Removing `--static` here allows for configure to complete successfully and the package to be installed and everything seems to be working.

**Additional context**

This seems to be related to #1696 and potentially some of the changes made to address it. From a brief survey of `configure.ac` I'm not sure why static would be needed since the package and everything else are being dynamically linked.

The homebrew on the laptop configuration is standard - happy to provide specific versions and details if helpful.

<details>
```
> sessionInfo()
R version 4.2.1 (2022-06-23)
Platform: aarch64-apple-darwin21.5.0 (64-bit)
Running under: macOS Monterey 12.5

Matrix products: default
BLAS:   /opt/homebrew/Cellar/openblas/0.3.21/lib/libopenblasp-r0.3.21.dylib
LAPACK: /opt/homebrew/Cellar/r/4.2.1_2/lib/R/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base

loaded via a namespace (and not attached):
 [1] Rcpp_1.0.9         magrittr_2.0.3     units_0.8-0        tidyselect_1.1.2
 [5] R6_2.5.1           rlang_1.0.4        fansi_1.0.3        dplyr_1.0.9
 [9] tools_4.2.1        grid_4.2.1         KernSmooth_2.23-20 utf8_1.2.2
[13] cli_3.3.0          e1071_1.7-11       DBI_1.1.3          class_7.3-20
[17] assertthat_0.2.1   tibble_3.1.8       lifecycle_1.0.1    sf_1.0-8
[21] purrr_0.3.4        vctrs_0.4.1        glue_1.6.2         proxy_0.4-27
[25] compiler_4.2.1     pillar_1.8.0       generics_0.1.3     classInt_0.4-7
[29] pkgconfig_2.0.3
```

```
> sf:::sf_extSoftVersion()
          GEOS           GDAL         proj.4 GDAL_with_GEOS     USE_PROJ_H
      "3.11.0"        "3.5.1"        "9.0.1"         "true"         "true"
          PROJ
       "9.0.1"
```
</details>

### LINKED ISSUE #1984: global polygon is, by default, seen as empty polygon in s2.
**Describe the bug**

a global polygon, like `POLYGON ((-180 -90, 180 -90, 180 90, -180 90, -180 -90))` is seen by `s2` as an empty polygon when `orientation` is `FALSE` in `st_as_s2` (which is the default). This leads to an infinite loop in the example below:

**To Reproduce**

```
> library(stars)
Loading required package: abind
Loading required package: sf
Linking to GEOS 3.10.2, GDAL 3.4.3, PROJ 8.2.0; sf_use_s2() is TRUE

> bb = st_as_stars() |> st_bbox() |> st_as_sfc() # global:

> bb
Geometry set for 1 feature 
Geometry type: POLYGON
Dimension:     XY
Bounding box:  xmin: -180 ymin: -90 xmax: 180 ymax: 90
Geodetic CRS:  WGS 84
POLYGON ((-180 -90, 180 -90, 180 90, -180 90, -...

> bb[[1]]
POLYGON ((-180 -90, 180 -90, 180 90, -180 90, -180 -90))

> st_sample(bb, 10)
```

jcheng5:
Fixes #1982.

The Intel and Apple Silicon versions of Homebrew use different directory prefixes (/usr/local and /opt/homebrew, respectively). This change uses the brew command to determine the prefix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,20 +80,20 @@ brew install gdal
 Once gdal is installed, you will be able to install `sf` package from source in R. With the current version of `proj` on homebrew, installation requires additional configuration:
 
 ```r
-install.packages("sf", configure.args = "--with-proj-lib=/usr/local/lib/")
+install.packages("sf", type = "source", configure.args = "--with-proj-lib=$(brew --prefix)/lib/")
 ```
 
 Or the development version:
 
 ```r
 library(remotes)
-install_github("r-spatial/sf", configure.args = "--with-proj-lib=/usr/local/lib/")
+install_github("r-spatial/sf", configure.args = "--with-proj-lib=$(brew --prefix)/lib/")
 ```
 
 If you are using `sf` and `rgdal` together, it is necessary to install `rgdal` from source using this configuration:
 
 ```r
-install.packages("rgdal", configure.args = c("--with-proj-lib=/usr/local/lib/", "--with-proj-include=/usr/local/include/"))
+install.packages("rgdal", type = "source", configure.args = c("--with-proj-lib=$(brew --prefix)/lib/", "--with-proj-include=$(brew --prefix)/include/"))
 ```
 
 Alternatively, [these instructions](https://stat.ethz.ch/pipermail/r-sig-mac/2017-June/012429.html) explain how to install gdal using kyngchaos frameworks.


edzer:
Happy to merge this, but I don't see how it fixes #1984?

jcheng5:
🤦‍♂️  Sorry, I meant #1982.

jcheng5:
Let me amend the commit description so it doesn’t auto-close 1994

edzer:
Thanks, Joe!

jcheng5:
Ah, you were quicker than me--you'll need to reopen 1984.

Thanks!

kadyb:
This fix should also be applied to [terra README](https://github.com/rspatial/terra#macos) if both packages use the same base? CC: @rhijmans

PR #2127, opened 2023-03-19
Use registration with dynamic library and update RcppExports

eddelbuettel:
The PR contains two commits.  The first adds the missing `, registration=TRUE` to the roxygen directive that write the corresponding (and important!) line in `NAMESPACE`.

With registration declared, we then re-run `Rcpp::compileAttributes()` which reflects the use of registrations and uses symbols in lieu of characters when calling compiled functions.  This should take care of parts of your ongoing update for CRAN.


edzer:
Fantastic, thanks a lot!!

PR #2136, opened 2023-03-29
Describing plotting order for polygons

jskoien:
Here is an attempt to describe plotting order and how to modify it. I did not include any examples, as I dont know any overlapping polygons without importing from other packages, such as rtop. Do you find it useful?


edzer:
Thanks!

kadyb:
It would be great to put the code in the `\code{}` tag. And it looks like this text was not generated correctly in .Rd:

`x %>% arrange(desc(st_area(x))) |> ggplot(aes(fill = column_name)) + geom_sf()`

Only this is printed:

`x `

edzer:
I'll take care of that; will move the examples to the @examples.

PR #2140, opened 2023-03-31
Make duplicated and unique work (fixes #2138)

### LINKED ISSUE #2138: `duplicated` fails with a factor column
I just encountered the case where `duplicated` on an `sf` object fails because the geometry column is "sticky". It seems to go wrong when `duplicated.data.frame` attempts to convert the factor columns to numeric:
``` r
require(sf)
#> Loading required package: sf
#> Linking to GEOS 3.8.0, GDAL 3.0.4, PROJ 6.3.1; sf_use_s2() is TRUE
p<-data.frame(x=1, y=2, a=gl(2,2), h=3) |> st_as_sf(coords=c('x','y'))
# non factor columns work as expected
duplicated(p %>% dplyr::mutate(a=as.character(a)))
#> [1] FALSE  TRUE FALSE  TRUE
# factor column fail
duplicated(p)
#> Error in lapply(x[i], as.numeric): 'list' object cannot be coerced to type 'double'
duplicated.data.frame
#> function (x, incomparables = FALSE, fromLast = FALSE, ...) 
#> {
#>     if (!isFALSE(incomparables)) 
#>         .NotYetUsed("incomparables != FALSE")
#>     if (length(x) != 1L) {
#>         if (any(i <- vapply(x, is.factor, NA))) 
#>             x[i] <- lapply(x[i], as.numeric)
#>         if (any(i <- (lengths(lapply(x, dim)) == 2L))) 
#>             x[i] <- lapply(x[i], split.data.frame, seq_len(nrow(x)))
#>         duplicated(do.call(Map, `names<-`(c(list, x), NULL)), 
#>             fromLast = fromLast)
#>     }
#>     else duplicated(x[[1L]], fromLast = fromLast, ...)
#> }
#> <bytecode: 0x56032210eb38>
#> <environment: namespace:base>
```
I'm not sure if this relates to #2131, but maybe for this specific case the easiest solution would be to implement a `duplicated.sf` function that already does the conversion to numeric and thus prevents `duplicated.data.frame` from failing? That solution should also directly fix `unique(p)`

bart1:
add a duplicated.sf method so `unique` and `duplicated` work with factor column in the data.frame


bart1:
@edzer I guess the tic error is unrelated to the pull request right? or is there a way to prevent it?

edzer:
That's right.

edzer:
Thanks!

jmaspons:
Does the function take into account the geometries to determine duplicated rows? I could not find any docs

bart1:
> Does the function take into account the geometries to determine duplicated rows? I could not find any docs

yes:
``` r
require(sf)
#> Loading required package: sf
#> Linking to GEOS 3.10.2, GDAL 3.4.1, PROJ 8.2.1; sf_use_s2() is TRUE
st_as_sf(data.frame(x=c(1,1),y=2), coords = 1:2) |> duplicated()
#> [1] FALSE  TRUE
st_as_sf(data.frame(x=c(1,3),y=2), coords = 1:2) |> duplicated()
#> [1] FALSE FALSE
```

<sup>Created on 2025-06-27 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

PR #2144, opened 2023-04-11
Integrate GeosNG into sf

edzer:
See e.g. https://lin-ear-th-inking.blogspot.com/2020/05/jts-overlay-next-generation.html

I think this can only be considered progress, so suggest to merge with main.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # version 1.0-13
 
+* use GEOS' overlayNG routines for (GEOS) Intersection, Difference, Union and SymDifference; #2143
+
 * added `duplicated.sf()`; #2138, #2140, thanks to @bart1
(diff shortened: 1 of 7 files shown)


rsbivand (approved):
Yes, does make sense; I didn't spot the `*Prec` versions when updating `rgeos` for which this is now too late with only six months to archiving.

edzer:
@paleolimbot I see from the geos sources that `...Prec_r()` functions were introduced in 3.9.1-1, so that needs updating of the #ifdefs

edzer:
> There is an analog in s2 as well (you can do an overlay with explicit precision there using the various snap options).

Now that you mention this - worth a post in the docs, or on r-spatial.org! Thanks,

Issue #2179, opened 2023-06-21
st_read() could "just work" for simple zipped spatial datasets (e.g. .shp)?


MichaelChirico:
I find public .shp datasets are often retrieved as .zip files containing the requisite bundle of files: LAYER.{dbf,prj,shp,shx}.

The most typical case is those files are alone in the .zip file, so it would make sense to me if `st_read()` would, when encountering a .zip file (inferred either from the extension or the [magic number](https://en.wikipedia.org/wiki/ZIP_(file_format))), attempt to `unzip(list = TRUE)` the file & check for a unique spatial dataset, then try `st_read()`, erroring if there's any ambiguity in the process.

```r
# setup
tmp <- tempfile(fileext = ".zip")
withr::with_dir(
  system.file("shape", package = "sf"),
  zip(tmp, list.files(pattern = "^nc"))
)
```

Here's an example of what would be "nice":

```r
st_read(tmp)
```

Currently gives a cryptic error:

```
Error: Cannot open "/tmp/RtmpgmVSJ1/file21e78753714e7.zip"; The source could be corrupt or not supported. See `st_drivers()` for a list of supported formats.
```

Workaround:

```r
unzip(tmp, exdir = tempdir())
st_read(file.path(tempdir(), "nc.shp"))
```

Alternatively, we can permanently unzip the file, but that has its own downsides, e.g. cluttering the data directory (4x the number of files laying around), introducing more issues about data organization (do I put each .shp in its own subdirectory? or create a 'geo' subdirectory and put all the .shp files there?), consuming more space (unzipping means decompressing, after all). We also have to memorize more stuff since the contained .shp file often doesn't match the name of the .zip file (i.e., "nc.shp" is not easy to infer automatically and often entails some manual inspection by the analyst).

For example, [this](https://data.sfgov.org/Geographic-Locations-and-Boundaries/Analysis-Neighborhoods/p5b7-5n3h) dataset has `Analysis Neighborhoods.zip` (:+1:) which contains `geo_export_742b8aeb-ec47-4467-8f7b-27b165ac89e1.shp` (:-1:).

---

Edit: Just realized `st_read()` already _does_ "just work" for directories:

```r
st_read(tempdir())
```

Works for the above example if there's no other `.shp` in `tempdir()`.

So working for .zip is only a small step beyond that, I think.

edzer:
See #1514 for a solution

MichaelChirico:
[...]

PR #2181, opened 2023-06-21
further clarify st_read() of .zip files

MichaelChirico:
Follow-up to 3362108a005ab77ba83953430dbaa6edf4b26075 for #2179

I think this more completely fills out the documentation. It makes sense to mostly punt upstream to GDAL here, so sf doesn't need to keep updating this man page as the VSI interface evolves. zip/gz/curl pieces should be pretty stable, so safe to mention here.


edzer:
Great, thanks!

PR #2195, opened 2023-06-27
Fix issue with key in plot.R

henryrscharf:
When plotting multiple character/factor-valued columns of an `sf` object, an error occurs when requesting a key because a portion of the code intended for numeric `values` gets executed (`classInt::classIntervals()`). I added lines to convert character `values` to a factor, and to check for numeric `values` before using `classIntervals()`. The changes solve the issue for me.


edzer:
There's one too many `)` on line 120.

henryrscharf:
Sorry about that. I removed the extra one.

edzer:
Thanks!

Issue #2199, opened 2023-07-05
Subsetting `sf` using an `sf` as an index does not error


bart1:
I just was pointed to this example where `df[df,]` does not fail (`df` being of the class `sf`). As for any other kind of indexing `i` needs to be character, logical or numeric I think this should fail (unless I miss why this should work?). It would be more intuitive if the failure behavior is similar to normal dataframe subsetting 
``` r
require(sf)
#> Loading required package: sf
#> Linking to GEOS 3.10.2, GDAL 3.4.1, PROJ 8.2.1; sf_use_s2() is TRUE
pt1 = st_point(c(0,1))
pt2 = st_point(c(1,1))
st_sfc(pt1, pt2)
#> Geometry set for 2 features 
#> Geometry type: POINT
#> Dimension:     XY
#> Bounding box:  xmin: 0 ymin: 1 xmax: 1 ymax: 1
#> CRS:           NA
#> POINT (0 1)
#> POINT (1 1)
d = data.frame(a = 1:2)
d$geom = st_sfc(pt1, pt2)
df = st_as_sf(d)
df[df,]
#> Simple feature collection with 2 features and 1 field
#> Geometry type: POINT
#> Dimension:     XY
#> Bounding box:  xmin: 0 ymin: 1 xmax: 1 ymax: 1
#> CRS:           NA
#>   a        geom
#> 1 1 POINT (0 1)
#> 2 2 POINT (1 1)
d[d,]
#> Error in xj[i]: invalid subscript type 'list'
d[df,]
#> Error in xj[i]: invalid subscript type 'list'
df[d,]
#> Error in `[.default`(geom, i): invalid subscript type 'list'
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.1.2 (2021-11-01)
#>  os       Ubuntu 22.04.2 LTS
#>  system   x86_64, linux-gnu
#>  ui       X11
#>  language (EN)
#>  collate  en_US.UTF-8
#>  ctype    en_US.UTF-8
#>  tz       Europe/Amsterdam
#>  date     2023-07-05
#>  pandoc   2.19.2 @ /usr/lib/rstudio/resources/app/bin/quarto/bin/tools/ (via rmarkdown)
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version date (UTC) lib source
[...]

PR #2200, opened 2023-07-05
Clarify documentation #2199

bart1:
As I looked at the documentation but overlooked this option for `i` I thought it might be useful to improve the documentation


edzer:
Thanks!

PR #2223, opened 2023-08-29
Force field type when appending to unexisting table (Fix #2206)

### LINKED ISSUE #2206: [Cross-Post] Polygon/multipolygon columns: incorrect write, correct read
https://github.com/r-dbi/RPostgres/issues/439

etiennebr:
This fix only applies to `RPostgres` driver for now since I could not get RPostgreSQL to work on my machine. This PR adds:

- Fix to the appending table as suggested by @dpprdan (thanks!) 
- Test to make sure that `append=TRUE` uses the right field type

Let me know if you want any modification.


edzer:
Great, thanks!

PR #2238, opened 2023-09-30
Paleolimbot stream reading

edzer:
@paleolimbot This leaves one test that breaks: 
```r
  > if ("GML" %in% st_drivers()$name) {
  +   gml = system.file("gml/fmi_test.gml", package = "sf")
  +   print(dim(st_read(gml, quiet = TRUE)))
  +   gml = system.file("gml/20170930_OB_530964_UKSH.xml.gz", package = "sf")
  +   print(dim(st_read(gml, layer = "Parcely", quiet = TRUE)))
  +   print(dim(st_read(gml, layer = "Parcely", int64_as_string=TRUE, quiet = TRUE)))
  + }
  [1] 22 11
  Error in st_as_sfc.WKB(x) : vapply(x, is.raw, TRUE) are not all TRUE
  Calls: print ... lapply -> FUN -> st_as_sfc -> st_as_sfc.WKB -> stopifnot
```
we can instrument this to use `use_stream=FALSE` but maybe this is of interest to look into.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -108,7 +108,7 @@ Suggests:
     tmap (>= 2.0),
     vctrs,
     wk
-Remotes: paleolimbot/arrow-nanoarrow@r-pkg-conversions
+Remotes: apache/arrow-nanoarrow#65
 LinkingTo: 
     Rcpp
 VignetteBuilder: 
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -73,6 +73,7 @@ Imports:
     utils
 Suggests:
     blob,
+    nanoarrow,
     covr,
     dplyr (>= 0.8-3),
     ggplot2,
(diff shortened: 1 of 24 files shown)


edzer:
Most changes are because the stream reader returns a geom column called `wkb_geometry`, and I didn't want this to be always renamed to `geometry` - some drivers have a different name (GPKG using `geom`) and others have named, or multiple named geometry columns (databases) where the name matters. All test changes involve shapefiles that assumed the geom column to be called `geometry`, I relaxed that. But the stream PR is good to go, from your side?

edzer:
Great! with `R_SF_ST_READ_USE_STREAM=true R CMD check` there are still errors in `testthat`, but most of it is the different geometry column name, and not converting to multi. We can still see what to do with that.

PR #2241, opened 2023-10-10
Use pkg-config on Windows when available.

kalibera:
This will use pkg-config to establish the libraries to link when pkg-config is available (which will soon be in Rtools). This should reduce the frequency of needed updates to linking when new features are linked/enabled in gdal.

### DIFF
--- a/src/Makevars.ucrt
+++ b/src/Makevars.ucrt
@@ -1,9 +1,14 @@
 
 PKG_CPPFLAGS = \
 	-DHAVE_PROJ_H
-LIBSHARPYUV = $(or $(and $(wildcard $(R_TOOLS_SOFT)/lib/libsharpyuv.a),-lsharpyuv),)
-PKG_LIBS =  \
+
+ifeq (,$(shell pkg-config --version 2>/dev/null))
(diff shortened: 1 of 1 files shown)


kadyb:
@edzer, it may be worth waiting until the Rtools will be update, because GHA and R-Universe will stop working (see: https://github.com/rspatial/terra/pull/1306).

kalibera:
Updated Rtools with pkg-config are already available at https://cran.r-project.org/bin/windows/Rtools/rtools43/rtools.html
Depending on which `make` your github actions use, you however may have to create a `pkg-config.bat` wrapper - more in https://cran.r-project.org/bin/windows/base/howto-R-devel.html, "Establishing the linking order in R packages via pkg-config". But, certainly, this is not urgent.

edzer:
@kalibera would merging this prevent the error we now see on [r-devel-windows](https://www.r-project.org/nosvn/R.check/r-devel-windows-x86_64/sf-00install.html)?

kalibera:
> @kalibera would merging this prevent the error we now see on [r-devel-windows](https://www.r-project.org/nosvn/R.check/r-devel-windows-x86_64/sf-00install.html)?

Yes

edzer:
Thanks!

PR #2253, opened 2023-11-03
Eliminate partial matching of args in plot methods

ateucher:
This is minor, but I have `options(warnPartialMatchArgs = TRUE)` set locally. This generates warnings when I plot with sf, as there are a couple of places where argument names are not fully spelled out. This PR fixes that in the places that I could find.


edzer:
Great, thanks!

PR #2266, opened 2023-11-17
Use roxygen_md docs

olivroy:
This will help fix rendering issues. see https://r-spatial.github.io/sf/reference/geos_binary_pred.html#details

While doing that, I noticed there was a typo in DESCRIPTION testhat instead of testthat, so I fixed that as well.

This required changing `expect_is()` by `expect_s3_class()` in tests.

I did some minor changes to the R files to ensure that x[i] was not wrongfully converted to \link by using back-ticks.

Ref: https://roxygen2.r-lib.org/articles/rd-formatting.html


olivroy:
According to test check, the package is not ready to transition to [testthat 3rd edition](https://testthat.r-lib.org/articles/third-edition.html). There seem to have messages that are not properly captured, according to the logs. Likely due to this change

testthat no longer silently swallows messages; you now need to deliberately handle them.

I do not have time (nor enough sf knowledge) to investigate that. the `expect_s3_class()`  changes are still worth it, since a future switch to testthat 3 will require them.

edzer:
Thanks for exploring!

DESCRIPTION
```diff
@@ -114,7 +114,8 @@ VignetteBuilder:
     knitr
 Encoding: UTF-8
 RoxygenNote: 7.2.3
-Config/testhat/edition: 3
+Roxygen: list(markdown = TRUE)
+Config/testthat/edition: 3
```
olivroy, line 118:
```suggestion
Config/testthat/edition: 2
```

As the failures are unrelated to the main purpose of this PR, maybe making this change would be sufficient to reduce warnings in tests, given that you agree with the other changes made.

olivroy:
@edzer do you agree with this, or I should close? 

I was thinking that I could work on improving some documentation if this is merged. 

Let me know if this requires adjustments

edzer:
Great, thanks!

PR #2269, opened 2023-11-21
Fix cross-compiling

jeroen:
Small fix for when we are cross compiling, for example linux to macos or linux to webassembly.
In those cases we can't run test programs (because they are for another architecture).

Autoconf automatically sets the `cross_compiling` variable based on `--host` or `--build` arguments being passed (see the output configure for details).


edzer:
Thanks, @jeroen !

PR #2283, opened 2023-12-07
Add bbox method to st_sample()

JosiahParry:
This PR adds `st_sample.bbox()` to provide an S3 method to st_sample() for bbox class object. 

Very often I find myself wanting sample random points over the extent of an sfc object. This requires me to write code such as 

```r
st_bbox(nc) |> 
  st_as_sfc() |> 
  st_sample(42)
```

It would be quite nice to simplify this to `st_sample(st_bbox(nc), 42)`. This PR makes it possible. 

Again, sorry for not updating `man/st_sample.Rd`. I used to be able to compile sf. I'll figure it out again at some point.

### DIFF
--- a/R/sample.R
+++ b/R/sample.R
@@ -139,6 +139,19 @@ st_sample.sfg = function(x, size, ...) {
 	st_sample(st_geometry(x), size, ...)
 }
 
+#' @export
+#' @name st_sample
+#' @examples
+#' bbox = st_bbox(
+#'	c(xmin = 16.1, xmax = 16.6, ymax = 48.6, ymin = 47.9),
+#' 	crs = st_crs(4326)
+#' )
+#' 
+#' st_sample(bbox, 5)
+st_sample.bbox = function(x, size, ...) {
+	st_sample(st_as_sfc(x), size, ...)
+}
+
 st_poly_sample = function(x, size, ..., type = "random",
                           offset = st_sample(st_as_sfc(st_bbox(x)), 1)[[1]],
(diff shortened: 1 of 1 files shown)


edzer:
I can see that this simplifies life in R2 cases, my reservation is the user expectation when doing this on the sphere:
```r
library(sf)
# Linking to GEOS 3.11.1, GDAL 3.6.4, PROJ 9.1.1; sf_use_s2() is TRUE
set.seed(13331)
st_bbox(c(xmin = 0, xmax = 45, ymin = 0, ymax = 45), crs = st_crs('OGC:CRS84')) |>
  st_as_sfc() |>
  st_sample(42) |>
  st_bbox()
#       xmin       ymin       xmax       ymax 
#  1.3055001  0.4876549 44.2484992 46.2575499 
```
and looking at `ymax`. This is expected, but can be better understood, I believe, when we force the user to leave the `st_as_sfc()` in the pipeline, and when wanting to do this on a box formed by parallels and meridians, to force them to add a `st_segmentize()` in it while s2 is switched off (sampling with s2 switched off does not give a uniform sampling over the sphere). Maybe we need examples of this in documentation / vignettes / books?

edzer:
Or we add a useful `stop("")` message in case coordinates are geographic, pointing the user to the problem?

The deeper problem is obviously the meaning / interpretation of a bounding box on the sphere.

edzer:
I'll take it from here.

JosiahParry:
Ah, thank you! Can't wait to see how you implement it.

PR #2285, opened 2023-12-08
Improve documentation

olivroy:
* Take advantage of roxygen markdown support
* Update links
* Generate [package docs](https://roxygen2.r-lib.org/articles/rd-other.html?q=keywords#packages) with roxygen2.
* Add `@keywords internal`, so that the functions that are not user-facing do not appear in the function index on the pkgdown site (nor in R help system), but still exported. copy from roxygen2 docs: There’s one exception: `@keywords internal`. It’s useful because it removes the function from the documentation index; it’s useful for functions aimed primarily at other developers, not typical users of the package.
* Fixup the defunct docs
* Have `?sf` point to `st_sf()` rather than the package docs.
* Fixup aliases: they should be separated by spaces, not commas (there still may be some problems of duplicate aliases?)
* Update `@name` vs `@rdname` [Reference roxygen2](https://roxygen2.r-lib.org/dev/articles/reuse.html?q=@rdname#multiple-functions-in-the-same-topic) (generally, `@name` should be used to override the default name of the man page, and `@rdame` should be used to document other things in the same man page.

I recommend you look primarly at the diff of Rd files, as it will be more telling of the impacts of the changes.


edzer:
Great clean-up, thanks!

PR #2288, opened 2023-12-11
Various Tests lints

olivroy:
In advance of switching to testthat 3, I thought I'd revisit the tests a little to make them more consistent.

The rationale behind 

```patch
-  expect_true(all(st_bbox(p1_norm) == c(0,0,1,1)))
+ expect_equal(unclass(st_bbox(p1_norm)), c(0,0,1,1), check.attributes = FALSE)
```

is that it accidentally fails for is `object` has length 0.

See this example

```r
# before, the test passed.
pt <- sf::st_sfc( sf::st_point( x = c(0,1,3,3)))
expect_true( all( attr( pt, "zbox" ) == c(3,3) ) )

# with the refactor, this would fail (that's why I commented it out.
pt <- st_sfc(st_point( x = c(0,1,3,3)))
# expect_equal(attr( pt, "zbox" ), c(3, 3)) # FIXME: now NULL
```


tests/testthat/test_zm_range.R
```diff
@@ -1,83 +1,90 @@
...
+
+	gc <- st_geometrycollection(x = list(pt, mpl))
+	expect_st_z_range(gc, c(-10, 10))
+	expect_equal(st_z_range(gc), st_z_range(gc))
 })
 
 
 test_that("sf::st_z_range and sf::st_z_range returns correct value from sfc objects", {
+	pt <- st_sfc(st_point( x = c(0,1,3,3)))
+	# expect_equal(attr( pt, "zbox" ), c(3, 3)) # FIXME: now NULL
```
olivroy, line 53: Hi @edzer , please let me know if you agree with those changes, they are mostly stylistics, but when doing some linting, it helped me uncover this bug (or wrong test). If you want I don't mind breaking it down into smaller PRs.

The main goal is to standardize the tests, reduce noise before switching to testthat 3e.

edzer:
Great, thanks!!

PR #2295, opened 2023-12-17
Update geos_unary docs to correct wrong function name (st_use_s2() -> sf_use_s2())

tylermorganwall:
I noticed that the geos_unary docs refer to `sf_use_s2()` as `st_use_s2()`: this PR fixes that incorrect reference.


edzer:
Thanks Tyler!

PR #2323, opened 2024-01-26
fix gdal_addo()

ethanplunkett:
Fix bug that prevented using `config_options` argument to `gdal_addo()`

Reproducible  example:
``` 
    library(sf)
    dir = file.path(tempdir(), "gdal_addo")
    dir.create(dir)
    tif = file.path(dir, "geomatrix.tif")
    file.copy(system.file("tif/geomatrix.tif", package = "sf"), tif, overwrite = TRUE)
    gdal_addo(tif, read_only = TRUE, config_options = c(COMPRESS_OVERVIEW="LZW"))
```
Previously threw error:  *Error: config_options should be a character vector with names, as in c(key="value")*

The problem was caused by calling `as.character(config_options)` which stripped the names before passing it on to `CPL_gdaladdo()`


edzer:
Great, thanks!

PR #2328, opened 2024-02-05
merge.sf renames geometry column name to 'geometry' arbitrarily

faridcher:
```r
x <- st_sf(id=1:2, geom=st_sfc(st_point(1:2), st_point(2:3)))
y <- data.frame(id=2:3, name=letters[2:3])
merge(x,y) # renames geom to geometry
```


edzer:
This results in many failures in GA; please check.

PR #2329, opened 2024-02-05
sfc crs is lost during arithmetic operations

faridcher:
```r
# example 1
st_sfc(st_point(0:1), st_point(2:3), crs=4326) - c(1,1)

# example 2:   sfc - sfc 
st_sfc(st_point(0:1), st_point(2:3), crs=4326) - st_sfc(st_point(c(1,1)), crs=4326)
```


edzer:
This is on purpose: what you do is a (re)projection, and the resulting CRS is unknown.

PR #2330, opened 2024-02-05
crs is lost in st_as_sfc.bbox

faridcher:
```r
st_as_sfc(st_bbox(st_linestring(matrix(1:4,2))), crs=4326)
```

### DIFF
--- a/R/sfc.R
+++ b/R/sfc.R
@@ -600,7 +600,8 @@ st_as_sfc.blob = function(x, ...) {
 
 #' @name st_as_sfc
 #' @export
-st_as_sfc.bbox = function(x, ...) {
+st_as_sfc.bbox = function(x, crs=NA_crs_, ...) {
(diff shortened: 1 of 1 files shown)


edzer:
Yes, but what if argument `crs` is different from the CRS of `x`? Why not use
```r
st_as_sfc(st_set_crs(st_bbox(st_linestring(matrix(1:4,2))), 4326))
```

faridcher:
i think `st_set_crs` is too verbose and redundant. Why not just use st_as_sfc? I think `st_bbox(obj, ..., crs)` should support objects other than Extent and numeric so  `st_bbox(st_linestring(matrix(1:4,2)), crs=4326)` would work then.

edzer:
You did not answer my first question.

faridcher:
You are right. I wrote this patch when st_bbox did not support crs. `st_set_crs` works here but my above side makes st_bbox more flexible. thank you

PR #2341, opened 2024-02-15
New approach to register tidyverse S3 class

olivroy:
Use new way to register S3 methods with `usethis::use_standalone("r-lib/rlang", "s3-register")`. https://r-pkgs.org/dependencies-in-practice.html#imports-and-exports-related-to-s3

~I also removed some S3 methods from docs, for which it is not very necessary. `format()`, `is.na()`, or print methods as they are obvious.~ 

As you see in the preview, the man page looks cleaner! (no more .sf suffix)
image


olivroy:
Works now. sf suggests dplyr 1.0, which is 4 years old. I think this is fine.

olivroy:
I reverted unrelated changes and just included the S3 register refactor!

PR #2346, opened 2024-02-22
update URL in README

kadyb:
The tic-db badge never complains despite the workflow fails because the link is outdated.


edzer:
Thanks!

PR #2399, opened 2024-05-20
add text method

btupper:
Hi,

It very much appeals to my inner laziness to have a `text()` method. For any `sf` object this will automatically place a label at each feature's centroid.  If somebody wanted finer control, they could make a POINT table to plot.  I know you don't want to add more stuff to `sf`, but this feels pretty unobstrusive.

What do you think?

``` r
library(sf)
#> Linking to GEOS 3.10.2, GDAL 3.5.0, PROJ 9.0.0; sf_use_s2() is TRUE
nc = st_read(system.file("gpkg/nc.gpkg", package="sf"), quiet = TRUE)
plot(nc["SID74"], reset = FALSE)
text(nc, labels = substring(nc$NAME,1,1))
```

<sup>Created on 2024-05-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


edzer:
I think this is a good idea, but please

* add `of_largest_polygon = FALSE` after the `...` and pass this on to `st_centroid`
* don't use `|>` in  the function (it would make `sf` no longer work with R versions prior to 4.1)

Maybe `row.names(x)` is a more sensible default for `labels`?

btupper:
Removing pipes gave me a good laugh - I forgot there was a time before pipes.

R/plot.R
```diff
@@ -700,6 +700,25 @@ sf.colors = function (n = 10, cutoff.tails = c(0.35, 0.2), alpha = 1, categorica
...
+#
+#' @param labels character, text to draw (one per row of input)
+#' @param of_largest_polygon logical, passed on to \code{st_centroid}
+#' @name plot
+#' @export
+#' @details Adds text to an existing graphic. Text is placed at the centroid of 
+#'   each feature in \code{x}.  Provide POINT features for full control of placement.
+#' @examples
+#' text(nc, labels = substring(nc$NAME,1,1))
+text.sf = function(x, labels = row.names, ..., of_largest_polygon = FALSE){
```
edzer, line 713: `labels = row.names(x)`

R/plot.R
```diff
@@ -700,6 +700,25 @@ sf.colors = function (n = 10, cutoff.tails = c(0.35, 0.2), alpha = 1, categorica
...
+  x = st_geometry(x) 
+  x = st_centroid(x, of_largest_polygon = FALSE)
```
edzer, line 715: `of_largest_polygon = of_largest_polygon`

edzer:
Thanks!

btupper:
Well, thank you for your patience with my embarrassing show.

btupper:
Oh, geez, I hadn't thought of that.  There is quite a menagerie that might be included...

`polygon`, `polypath`, `symbols` and `sunflowerplot`  

I have been thinking about using `sf` for a project using flow cytometry data... the geospatial tools would work just fine for gate analyses.

edzer:
too bad that `xy.coords` is not a generic...

edzer:
I think I'll stop here, for `lines` it is less trivial what it should do.

btupper:
Oooo, I see.  I assumed that all of the graphics package functions were generic.  Bummer.  Well, thanks for `text` and `points`!

PR #2421, opened 2024-08-08
Fix compilation error on gcc 14.2

djhshih:
R
> devtools::install()

Compilation error encountered with gcc 14.2:

```
In file included from gdal.cpp:1:
./gdal.h:3:1: error: ‘OGRSpatialReference’ does not name a type; did you mean ‘DEFINED_OGRSpatialReferenc
eH’?
    3 | OGRSpatialReference *handle_axis_order(OGRSpatialReference *sr);
      | ^~~~~~~~~~~~~~~~~~~
      | DEFINED_OGRSpatialReferenceH
./gdal.h:4:1: error: ‘Rcpp’ does not name a type
    4 | Rcpp::List create_crs(const OGRSpatialReference *ref, bool set_input);
      | ^~~~
```

This commit forward declares `OGRSpatialReference` to fix the first error above, and it includes `Rcpp.h` explicitly to fix the second error.

```
$ R --version
R version 4.4.1 (2024-06-14) -- "Race for Your Life"

$ gcc --version
gcc (GCC) 14.2.1 20240802

$ uname -a
Linux durus 6.10.3-arch1-1 #1 SMP PREEMPT_DYNAMIC Sun, 04 Aug 2024 05:11:32 +0000 x86_64 GNU/Linux
```

### DIFF
--- a/src/gdal.h
+++ b/src/gdal.h
@@ -1,3 +1,6 @@
+#include <Rcpp.h>
+class OGRSpatialReference;
+
 void set_error_handler(void);
 void unset_error_handler(void);
 OGRSpatialReference *handle_axis_order(OGRSpatialReference *sr);


rsbivand:
@djhshih Please indicate which version of GDAL you are using. I cannot replicate with GCC 14.2.1 (current Fedora 40) for:
```
$ gdalinfo --version
GDAL 3.9.1, released 2024/06/22
```
GDAL 3.9.2 RC was made available yesterday with a further fix for GCC 14.2: https://github.com/OSGeo/gdal/blob/v3.9.2RC1/NEWS.md

rsbivand:
@djhshih How were R and GDAL installed? I always install from source, but if you use a package manager, it may arbitrarily impose CFLAGS that max out the package manager's choice of what is felt to be secure. My CFLAGS as shown in running R CMD INSTALL (please always avoid `devtools` as it and RStudio insert layers of indeterminacy):
```
configure: Package CPP flags:  -I/usr/local/include -I/usr/include/webp -DWITH_GZFILEOP -DHAVE_PROJ_H -I/usr/local/include -I/usr/local/include
configure: Package LIBS: -L/usr/local/lib64 -lproj   -L/usr/local/lib64 -lgdal -L/usr/local/lib64 -lgeos_c
...
g++ -std=gnu++17 -I"/home/rsb/topics/R/R441-share/lib64/R/include" -DNDEBUG -I/usr/local/include -I/usr/include/webp -DWITH_GZFILEOP -DHAVE_PROJ_H -I/usr/local/include -I/usr/local/include -I'/home/rsb/lib/r_libs/Rcpp/include' -I/usr/local/include    -fpic  -g -O2   -c gdal.cpp -o gdal.o
```
Yours are?

edzer:
We're not yet testing with gcc 14 I believe, but these fixes seem odd and unnecessary to me.

rsbivand:
@edzer I did test, as F40 has 14.2.1. The problem is most likely that Arch Linux binary builds of R impose specific, possibly aggressive, C/CXXFLAGS (unknown, @djhshih did not reply) that change the behaviour of the build train. I don't see any problems with default flags. This is up to the user, I think. @djhshih needs to document the flags used here, both for building GDAL etc., and R, and those that `R CMD config` emits in `configure` when installing `sf`.

PR #2426, opened 2024-08-23
Voronoi point order

rsbivand:
It turned out to be possible only in R, adding a logical argument to `st_voronoi`, and converting the two to an integer in `st_voronoi.sfc` for passing to `CPL_geos_voronoi`. Example and test added.


edzer:
Thanks; it looks like you merged this in the `sf_aware` branch, so I'll merge this after that gets merged.

Issue #2176, opened 2023-06-09
Documentation request


olivroy:
Hello,

I am learning the sf package, and browsed through the referene https://r-spatial.github.io/sf/reference/index.html 
I find it a bit difficult to browse, and to know which functions fit together (apart from the ones documented together.

Would it be possible to improve `_pkgdown.yml`? by grouping functions and add small description in the function index

See at https://github.com/tidyverse/dplyr/blob/main/_pkgdown.yml and what the result is https://dplyr.tidyverse.org/reference/index.html

Thanks for the great package

kadyb:
BTW: In the cheatsheet the functions are grouped into 6 groups, maybe it will be useful: https://github.com/rstudio/cheatsheets/blob/main/sf.pdf

edzer:
PR welcome!

edzer:
Another grouping is found in table 1 of https://journal.r-project.org/archive/2018/RJ-2018-009/RJ-2018-009.pdf

PR #2431, opened 2024-09-01
Separate functions in smaller topics

olivroy:
This is to make clearer what operations do what, hence facilitate the selection of the right `st_*()` function.

This work will require a few iterations to continue, but thought I'd send this PR first to see if there is interest.

The end goal is to combine documentation that applies to many functions in a help topic while separating out the pieces of docs that only apply to a single function to avoid clutter in the doc.

The goal is to have multiple documentation topic and use [roxygen2 tags](https://roxygen2.r-lib.org/articles/reuse.html) to reuse documentation (like `@inheritParams`.

Also added a rough classification in the pkgdown index for #2176 (can still be improved)


olivroy:
@edzer Just a reminder! When you have time, please let me know what you think..

The goal ia to improve the prompts we get when we use autocompletion, to be able to find information more quickly by dividing topics up in more digestible pieces.  

A lot of information is present already. The goal of this PR (and the follow-up ones I am planning) is to reorganize slightly by taking advantage of roxygen2 [reusability](https://roxygen2.r-lib.org/articles/reuse.html) features. 

If you want, I can add review comments to help you follow more easily?

Issue #2436, opened 2024-09-11
sf 1.0-17 fails to install with GDAL < 3.4: gdal.cpp:713:16: error: ‘ret’ was not declared in this scope


HenrikBengtsson:
# Issue

```r
> install.packages("sf")
Installing package into '/wynton/home/cbi/hb/R/rocky8-x86_64-pc-linux-gnu-library/4.4-CBI-gcc13'
(as 'lib' is unspecified)
trying URL 'https://cloud.r-project.org/src/contrib/sf_1.0-17.tar.gz'
Content type 'application/x-gzip' length 3802276 bytes (3.6 MB)
==================================================
downloaded 3.6 MB

* installing *source* package ‘sf’ ...
** package ‘sf’ successfully unpacked and MD5 sums checked
** using staged installation
configure: CC: gcc
configure: CXX: g++ -std=gnu++17
checking for gdal-config... /usr/bin/gdal-config
checking gdal-config usability... yes
configure: GDAL: 3.0.4
checking GDAL version >= 2.0.1... yes
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables...
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether the compiler supports GNU C... yes
checking whether gcc accepts -g... yes
checking for gcc option to enable C11 features... none needed
checking for stdio.h... yes
checking for stdlib.h... yes
checking for string.h... yes
checking for inttypes.h... yes
checking for stdint.h... yes
checking for strings.h... yes
checking for sys/stat.h... yes
checking for sys/types.h... yes
checking for unistd.h... yes
checking for gdal.h... yes
checking GDAL: linking with --libs only... yes
checking GDAL: /usr/share/gdal/pcs.csv readable... no
checking GDAL: checking whether PROJ is available for linking:... yes
checking GDAL: checking whether PROJ is available for running:... yes
configure: GDAL: 3.0.4
configure: pkg-config proj exists, will use it
configure: using proj.h.
configure: PROJ: 6.3.2
checking PROJ: checking whether PROJ and sqlite3 are available for linking:... yes
checking for geos-config... /usr/bin/geos-config
checking geos-config usability... yes
configure: GEOS: 3.7.2
checking GEOS version >= 3.4.0... yes
checking for geos_c.h... yes
checking geos: linking with -L/usr/lib64 -lgeos_c... yes
[...]

PR #2437, opened 2024-09-11
address #2436

rsbivand:
Quick fix for `ret` not being allocated in-scope in `CPL_transform_bounds`, added note to NEWS, bumped version number and added test in R code so that compiled code shouldn't be reached with GDAL < 3.4.0 - #2436


edzer:
Thanks!

rsbivand:
Hope your gears are working as expected!

edzer:
They do!

Issue #1371, opened 2020-04-16
Make st_voronoi() retain the point order


jplecavalier:
I have a very simple example in which I simulate 100 points, then create the Voronoi polygons associated with this partition and assigne it to another column. My problem is that the order of the polygons created by `st_voronoi` won't have the same order than the points used in argument.

```
library(sf)
library(data.table)
library(ggplot2)

set.seed(20200415L)

border <- st_sfc(st_polygon(list(matrix(c(0, 0, 1, 1, 0, 0, 1, 1, 0, 0), ncol=2))))

data <- data.table(
  point = st_sfc(lapply(replicate(100, runif(2), simplify=FALSE), st_point))
)

data[, polygon:=st_intersection(st_collection_extract(st_voronoi(do.call(c, point))), border)]

sample_data <- data[sample(1:.N, 10)]

sample_data[]
##                             point polygon
##   1:  POINT (0.4209446 0.4083639)    <XY>
##   2:    POINT (0.0919243 0.40189)    <XY>
##   3:  POINT (0.7040549 0.1799017)    <XY>
##   4:  POINT (0.5056223 0.7427985)    <XY>
##   5:   POINT (0.2678713 0.343384)    <XY>
##   6:  POINT (0.9404894 0.6390899)    <XY>
##   7:   POINT (0.188111 0.2670979)    <XY>
##   8:  POINT (0.5343007 0.3091294)    <XY>
##   9: POINT (0.01642705 0.8337068)    <XY>
##  10: POINT (0.4475015 0.05526273)    <XY>

ggplot(sample_data) +
  geom_sf(
    mapping = aes(
      geometry = point
    )
  ) +
  geom_sf(
    mapping = aes(
      geometry = polygon
    ),
    fill = NA
  ) +
  theme_minimal()
```
![image](https://user-images.githubusercontent.com/21185978/79406468-2f499f80-7f65-11ea-9326-287de988fe6c.png)

As one could see with the image above, the point and the polygon are not associated correctly and this is due to the `st_voronoi` function which do not preserve the order of the points.

Is it possible to improve this on your side or my only solution is to wrap `st_voronoi` into a personal function that will reorder the polygon to follow the point pattern correctly?

---

```
sessionInfo()
##  R version 3.5.0 (2018-04-23)
##  Platform: x86_64-apple-darwin15.6.0 (64-bit)
##  Running under: macOS  10.15.4
##  
##  Matrix products: default
[...]

PR #2438, opened 2024-09-12
create clean voronoi point order PR #1371 #2426

rsbivand:
This is a re-creation of closed #2426 based on #2437 rather than `sf_aware` which is waiting to mature. So just the changes needed to use Voronoi point order for GEOS >= 3.12. I'm not seeing problems with GEOS 3.13, by the way.


edzer:
Thanks!

Issue #2442, opened 2024-09-18
st_simplify produces error when specifying tolerance values per simple feature on S2 code path


goergen95:
**Describe the bug**
I am trying to simplify polygons with a tolerance value based on the number of vertices per simple feature.
On the $S^2$ code path, this results in an error because `s2::s2_simplify()` seems to expect a single value for the tolerance parameter. However, the docs for `sf::st_simplify()` state:

> dTolerance 	numeric; tolerance parameter, specified for all or for each feature geometry. [..]

**To Reproduce**
``` r
library(sf)
#> Linking to GEOS 3.12.2, GDAL 3.9.1, PROJ 9.4.1; sf_use_s2() is TRUE

sf_use_s2(TRUE)  

nc <- read_sf(system.file("shape/nc.shp", package="sf"))
nc <- st_transform(nc, st_crs("EPSG:4326"))
vertices <- as.numeric(table(st_coordinates(nc)[ ,5]))
nc2 <- st_simplify(nc, dTolerance = round(vertices/10))
#> Error in if (snap_radius > 3) {: the condition has length > 1
```

<sup>Created on 2024-09-18 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.4.1 (2024-06-14)
#>  os       Ubuntu 22.04.4 LTS
#>  system   x86_64, linux-gnu
#>  ui       X11
#>  language (EN)
#>  collate  en_US.UTF-8
#>  ctype    en_US.UTF-8
#>  tz       Etc/UTC
#>  date     2024-09-18
#>  pandoc   3.2 @ /usr/bin/ (via rmarkdown)
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version date (UTC) lib source
#>  class         7.3-22  2023-05-03 [2] CRAN (R 4.4.1)
#>  classInt      0.4-10  2023-09-05 [1] CRAN (R 4.4.1)
#>  cli           3.6.3   2024-06-21 [1] RSPM (R 4.4.0)
#>  DBI           1.2.3   2024-06-02 [1] RSPM (R 4.4.0)
#>  digest        0.6.37  2024-08-19 [1] RSPM (R 4.4.0)
[...]

PR #2445, opened 2024-09-24
address #2442

goergen95:
Hi :wave:

I set up this PR to address #2442. I am unsure where to best handle this, here or in `s2`? However, I think I am currently not able to contribute to `s2`, so here goes this PR.

Using mapply to iterate over multiple dTolerance values. Also works for a vector of length 1. Needed to wrap `s2::s2_simplify()` with `sf::st_as_sfc()` since the latter seems currently to be unable to coerce a list of `s2_geography` objects to `sfc`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 * fix build failure with GDAL < 3.4.0 #2436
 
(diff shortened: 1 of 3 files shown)


edzer:
Thanks!

PR #2456, opened 2024-10-15
Fix typo in vignette 7

mthh:
While reading the vignette "Spherical geometry in sf using s2geometry" I spotted a typo, so here's a PR that fixes it.


edzer:
Thanks!

Issue #2455, opened 2024-10-11
Args from S2 are not documented in `st_buffer()`


kadyb:
We can see from the doc that arguments other than those coming from GEOS in [st_buffer()](https://r-spatial.github.io/sf/reference/geos_unary.html) are ignored, but in fact arguments from S2 work. I think `max_cells` in particular is relevant.

edzer:
as well as `min_level`, which I didn't know about!

kadyb:
https://github.com/r-spatial/sf/blob/39d5ab03c1c31ea217d480a90135606a8da92295/R/geom-transformers.R#L22

I noticed that `...` is correctly described, but it only appears once in the doc from this place:

https://github.com/r-spatial/sf/blob/39d5ab03c1c31ea217d480a90135606a8da92295/R/geom-transformers.R#L701

kadyb:
And this sentence is no longer true (if we use S2)?

> If any of endCapStyle, joinStyle, or mitreLimit are set to non-default values ('ROUND', 'ROUND', 1.0 respectively) then the underlying 'buffer with style' GEOS function is used.

PR #2458, opened 2024-10-21
improve `st_buffer()` description

kadyb:
I have improved the description of `st_buffer()` and I think it should be more explanatory now (for #2455). What do you think?


edzer:
Great, thanks!

Issue #2466, opened 2024-10-31
`sf` passes CMD check fine, but 4 packages using `sf` fail on GDAL 3.10.0 rc1


rsbivand:
- [x] https://github.com/ropensci/osmextract/issues/298
- [ ] https://github.com/Defra-Data-Science-Centre-of-Excellence/sgapi/issues/1
More serious are segfaults, possibly from `st_area`:
- [x] https://github.com/aestears/plantTracker/issues/11
- [x] https://github.com/holans/ST-COS/issues/2

I have one platform with 1.0-18 and 3.9.3, another with 1.0-18 and 3.10.0rc1.

@edzer

rsbivand:
`osmextract`, `sgapi`, `plantTracker` and `stcos` errors unchanged with GDAL 3.10.0rc2, `sf` passes all its own checks with GDAL 3.10.0rc2.

rouault:
I've seen suspicous code in CPL_area() that could explain the crash in it. Cf https://lists.osgeo.org/pipermail/gdal-dev/2024-October/059747.html for diagnostic and suggestions

rsbivand:
I'll try to check later, had to go out, thanks!

rsbivand:
Segfaults in `stcos` and `plantTracker` resolved for GDAL 3.10.0rc2 with https://lists.osgeo.org/pipermail/gdal-dev/2024-October/059747.html. PR to follow Friday afternoon unless @edzer gets there first.

edzer:
Submitted to CRAN...

edzer:
On CRAN.

PR #2470, opened 2024-11-01
address #2466, incorporate #2468 and #2469

rsbivand:
Adding @rouault 's correction from gdal-devel, now updated with #2468 and #2469. As reported in #2466 the failing packages `stcos` and `plantTracking` now pass with rc1 and rc2.

### DIFF
--- a/src/gdal_geom.cpp
+++ b/src/gdal_geom.cpp
@@ -12,14 +12,19 @@ Rcpp::NumericVector CPL_area(Rcpp::List sfc) {
 	for (size_t i = 0; i < g.size(); i++) {
 		if (g[i]->getDimension() == 2) {
 			OGRwkbGeometryType gt = OGR_GT_Flatten(g[i]->getGeometryType());
-			if (gt == wkbMultiSurface || gt == wkbMultiPolygon) {
+// PR 2468
+//			if (gt == wkbMultiSurface || gt == wkbMultiPolygon) {
+			if (OGR_GT_IsSubClassOf(gt, wkbGeometryCollection)) {
+				// will match OGRMultiPolygon, OGRMultiSurface and OGRGeometryCollection
 				OGRGeometryCollection *gc = (OGRGeometryCollection *) g[i];
 				out[i] = gc->get_Area();
-			} else {
+			} else if (OGR_GT_IsSurface(gt)) {
 				OGRSurface *surf = (OGRSurface *) g[i];
 				out[i] = surf->get_Area();
-			} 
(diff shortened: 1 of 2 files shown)


src/gdal_geom.cpp
```diff
@@ -12,14 +12,19 @@ Rcpp::NumericVector CPL_area(Rcpp::List sfc) {
 	for (size_t i = 0; i < g.size(); i++) {
 		if (g[i]->getDimension() == 2) {
 			OGRwkbGeometryType gt = OGR_GT_Flatten(g[i]->getGeometryType());
-			if (gt == wkbMultiSurface || gt == wkbMultiPolygon) {
+// PR 2468
+//			if (gt == wkbMultiSurface || gt == wkbMultiPolygon) {
```
rouault, line 16: That doesn't bring much, right ?

```suggestion
```

src/gdal_geom.cpp
```diff
@@ -63,8 +68,15 @@ Rcpp::NumericVector CPL_length(Rcpp::List sfc) {
 				}
 				break;
 			default: {
-					OGRGeometryCollection *a = (OGRGeometryCollection *) g[i];
-					out[i] = a->get_Length();
+// PR 2469
+/*					OGRGeometryCollection *a = (OGRGeometryCollection *) g[i];
+					out[i] = a->get_Length();*/
```
rouault, line 73:
```suggestion
```

rsbivand:
@rouault I leave commented-out code that may be discarded to make off-github review clearer. It can be dropped later. This is @edzer 's codebase, so I don't delete on my fork and PRs, I let him do that if he agrees. Since I (still) prefer to run all the R checks locally first (because I have control of GDAL etc., versions), I am conservative in letting github work things out. I edit in vim, as you might expect.

rouault:
> I am conservative in letting github work things out. I edit in vim, as you might expect.

I don't get it: that's git (the software and code repository format) which keeps the old versions. Not github which is just a UI over it. So there's no harm in deleting unused code... It can always been recovered through git revert, etc. That's I believe the good practice. At least in the projects I'm directly involved in ;-)

rsbivand:
Certainly good practice. I started with SCCS in 1985. Here, @edzer is lead and makes the calls, I chip in (not often, and not often right), and work in my fork. So commenting out replaced code shows the diffs "in place" for visual review. I'm unlikely to become more confident in my commits over time, I think.

edzer:
Thanks, both, for resolving this!

PR #2496, opened 2025-01-08
clarification in vignette 7 regarding `lwgeom`

kadyb:
I propose to add information in the text that the `s2` and `lwgeom` packages are compared when `sf_use_s2(FALSE)`, because I personally had the impression that this concerns the comparison of `s2` and `GEOS`.


edzer:
Makes sense, thanks!

PR #2497, opened 2025-01-11
fix bashisms

rsbivand:
See https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-fedora-gcc/sf-00check.html. First is unneeded, as R >=3.3.0 is required in DESCRIPTION. The second is needed, and the edit works, checked both with _R_CHECK_BASHISMS_=true _R_CHECK_BASHISMS_EXTRA_=true and R-devel, and against a doctored gdal-config reporting 3.6.0.


edzer:
Great, thanks!

PR #2504, opened 2025-02-08
try pkg-config before gdal-config

rsbivand:
In relation to @s-u s email today, an attempt at trying pkg-config before falling back on gdal-config. Also small change to test for `stateplane.csv` not `pcs.csv` which disappeared some time ago in the GDAL data folder.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # version 1.0-20
 
+* `st_interpolate_aw()` fixes bug when a GEOMETRYCOLLECTION contains multiple POLYGON structures; found by @mtennekes
+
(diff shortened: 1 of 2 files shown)


rsbivand:
Just added test for CURL missing from GDAL build. When run locally, it gives:
```
> aa[[114]]
$file
[1] "test-read.R"

$context
NULL

$test
[1] "GDAL was built with CURL"

$user
[1] 0.198

$system
[1] 0.001

$real
[1] 0.7

$results
$results[[1]]
<expectation_success/expectation/condition>
As expected
```

rsbivand:
@edzer, @s-u, with the proposed test, Simon would have had a chance of catching the regression when checking `sf`. Maybe we will not fall down the same manhole again with it?

PR #2511, opened 2025-03-22
print.sfg() returns the input sfg, not its formatted string

heavywatal:
The current behavior of `print.sfg()` is surprising because it returns a formatted string, which prevents the usage in a pipeline, e.g.,

```r
sf::st_point(1:2) |>
  print() |>
  sf::st_sfc() |>
  print()
# POINT (1 2)
# Error in UseMethod("st_bbox"): no applicable method for 'st_bbox' applied to an object of class "character"
```

As the vast majority of `print()` methods do, side-effect functions should return the original object (see <https://design.tidyverse.org/out-invisible.html>). Actually, `print.sf()` and `print.sfc()` already do this:

https://github.com/r-spatial/sf/blob/5567b747c8e5ad5aafc41ff0cfa103188fdb48e1/R/sf.R#L422-L445
https://github.com/r-spatial/sf/blob/5567b747c8e5ad5aafc41ff0cfa103188fdb48e1/R/sfc.R#L202-L267

### DIFF
--- a/R/sfg.R
+++ b/R/sfg.R
@@ -198,7 +198,7 @@ POLYGON2MULTIPOLYGON = function(x, dim = "XYZ") {
 print.sfg = function(x, ..., width = 0) { # avoids having to write print methods for 68 classes:
 	f = format(x, ..., width = width)
 	message(f)
-	invisible(f)
+	invisible(x)
 }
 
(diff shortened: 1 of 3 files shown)


edzer:
Thanks!

PR #2518, opened 2025-04-07
Weaken test for case of missing driver

MichaelChirico:
Alternatively, we can do a `skip_if_not()`, but that would also skip the other 3 tests here, which could still be useful even for users lacking the AVCE00 driver. We could also branch this test (`if ("AVCE00" %in% ...) expect_error() else skip(...)`).


edzer:
Thanks!

PR #2572, opened 2026-01-15
gdal.h: align declaration and definition

pjonsson:
This seems like a typo, so declare
the function that has a definition
in gdal.cpp.


pjonsson:
I have been using R for less than a week, so chances of user error are high, I hope the CI will catch any mistakes I've made.

edzer:
Great, thanks!

PR #2573, opened 2026-01-15
gdal.h: align create_crs declarations

pjonsson:
The other declaration of create_crs
has a default value for the second
parameter, so align this declaration
with the other declaration.

### DIFF
--- a/src/gdal.h
+++ b/src/gdal.h
@@ -1,7 +1,7 @@
 void set_error_handler(void);
 void unset_error_handler(void);
 OGRSpatialReference *handle_axis_order(OGRSpatialReference *sr);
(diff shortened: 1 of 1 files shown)


pjonsson:
Ok, that apparently broke something, I will continue to look at this tomorrow.

pjonsson:
@edzer having multiple declarations of the same function seems error prone, can I remove the duplicate definitions as long as this package still builds without compiler warnings, or does some third-party package depend on sf and use `gdal_sf_pkg.h`?

edzer:
I think package lwgeom does (has `sf` in `LinkingTo:`)

PR #2574, opened 2026-01-15
POC: Reimplementing vctrs methods for `sfc`

DavisVaughan:
This PR merges into your `vctrs` branch

@edzer this is what I'm imagining for what the vctrs methods would look like for some `sfc` subclasses

You were having issues in your branch due to existing `vec_proxy.sfc()` and `vec_restore.sfc()` methods, they were getting in the way for what you needed. But the ability to create a typed size 0 object is _very_ useful for unlocking correct methods here!

This is a "burn it down" approach that keeps (and greatly expands on) the tests but just rewrites everything.

Something notable here is that you _should not_ export any vctrs methods. Just lazily register them.

You can add some `@keyword internal` docs for them if you'd like, but it wouldn't be required.

This PR just implements methods for `sfc_POINT`, `sfc_LINESTRING`, and `sfc_GEOMETRY`. IIUC you would also need them for:
- `sfc_CIRCULARSTRING`
- `sfc_GEOMETRYCOLLECTION`
- `sfc_MULTILINESTRING`
- `sfc_MULTIPOINT`
- `sfc_MULTIPOLYGON`
- `sfc_POLYGON`

And that's just for the vector types. We'd then have to talk about methods for the sf data frame itself. The process would be similar but it would probably be worth it to have that in a separate file, and have this file just be about sfc methods.

But to me this feels like a good basis to start expanding from. What do you think?

### DIFF
--- a/R/tidyverse-vctrs.R
+++ b/R/tidyverse-vctrs.R
@@ -4,18 +4,23 @@
 register_vctrs_methods = function() {
 	s3_register("vctrs::vec_proxy", "sfc_POINT")
 	s3_register("vctrs::vec_proxy", "sfc_LINESTRING")
+	s3_register("vctrs::vec_proxy", "sfc_GEOMETRY")
 
 	s3_register("vctrs::vec_restore", "sfc_POINT")
 	s3_register("vctrs::vec_restore", "sfc_LINESTRING")
+	s3_register("vctrs::vec_restore", "sfc_GEOMETRY")
 
 	s3_register("vctrs::vec_ptype", "sfc_POINT")
 	s3_register("vctrs::vec_ptype", "sfc_LINESTRING")
+	s3_register("vctrs::vec_ptype", "sfc_GEOMETRY")
 
 	s3_register("vctrs::vec_ptype2", "sfc_POINT.sfc_POINT")
 	s3_register("vctrs::vec_ptype2", "sfc_LINESTRING.sfc_LINESTRING")
+	s3_register("vctrs::vec_ptype2", "sfc_GEOMETRY.sfc_GEOMETRY")
 
 	s3_register("vctrs::vec_cast", "sfc_POINT.sfc_POINT")
 	s3_register("vctrs::vec_cast", "sfc_LINESTRING.sfc_LINESTRING")
+	s3_register("vctrs::vec_cast", "sfc_GEOMETRY.sfc_GEOMETRY")
 }
 #nocov end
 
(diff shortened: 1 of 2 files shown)


DavisVaughan:
The biggest restriction right now is that there is no support for `vec_c()` across sfc class types.

The `c()` method "falls back" to the abstract `sfc_GEOMETRY` type. The vctrs methods do not do such a thing. We could add this, but it would result in a combinatorially explosive number of methods required (they'd all have the same implementation though). That's a result of the way the vctrs dispatch system works (without inheritance).

``` r
x <- st_sfc(st_point(1:2))
y <- st_sfc(st_linestring(matrix(3:4, nrow = 1)))

# "falls back" to the abstract sfc_GEOMETRY
c(x, y)
#> Geometry set for 2 features 
#> Geometry type: GEOMETRY
#> Dimension:     XY
#> Bounding box:  xmin: 1 ymin: 2 xmax: 3 ymax: 4
#> CRS:           NA
#> POINT (1 2)
#> LINESTRING (3 4)

# can't do that with the current methods
vctrs::vec_c(x, y)
#> Error in `vctrs::vec_c()`:
#> ! Can't combine `..1` <sfc_POINT> and `..2` <sfc_LINESTRING>.

# can combine "within" a single class
vctrs::vec_c(x, x)
#> Geometry set for 2 features 
#> Geometry type: POINT
#> Dimension:     XY
#> Bounding box:  xmin: 1 ymin: 2 xmax: 1 ymax: 2
#> CRS:           NA
#> POINT (1 2)
#> POINT (1 2)
vctrs::vec_c(y, y)
#> Geometry set for 2 features 
#> Geometry type: LINESTRING
#> Dimension:     XY
#> Bounding box:  xmin: 3 ymin: 4 xmax: 3 ymax: 4
#> CRS:           NA
#> LINESTRING (3 4)
#> LINESTRING (3 4)
```

edzer:
I added the x.x methods for the big 7 (the 7 geometry types used in 99.9% of the cases), and will add the x.y for them next, there's only 42 of those for each method; this should now pass check.

edzer:
> The c() method "falls back" to the abstract sfc_GEOMETRY type. The vctrs methods do not do such a thing. We could add this, but it would result in a combinatorially explosive number of methods required (they'd all have the same implementation though). That's a result of the way the vctrs dispatch system works (without inheritance).

After a lot of experimentation, and re-reading your comment, I come to the conclusion that `vctrs` doesn't let us create a `sfc_GEOMETRY` when joining a `sfc_POINT` with a `sfc_LINESTRING`, following the path you sketch with this PR. If that is the case, that I'm afraid we're at a dead end here, as one cannot meaningfully convert a POINT geometry into a LINESTRING geometry or vice versa. How do you suggest we proceed?

PR #2576, opened 2026-01-20
Fix GDAL 3.13 compatibility

pjonsson:
GDALGetMetadata has changed signature
in GDAL 3.13, so change the signature
of charpp2CV to be compatible with
that.

https://github.com/OSGeo/gdal/blob/9159b2538a36062318104e585ecc77a0c81e6de3/doc/source/user/migration_guide.rst?plain=1#L19-L23


pjonsson:
I don't have access to macOS so not sure how to fix the error. Is it possible the error is intermittent and retriggering the failed job would fix the problem?

edzer:
Thanks!

PR #2581, opened 2026-01-25
Stop casting away const

pjonsson:
The compiler might use this information
for optimization, so adjust the code
so the const isn't lost.


edzer:
> The compiler might use this information for optimization,

I don't think that this is a good argument for making the code clearly harder to read (for those who maintain).

pjonsson:
> > The compiler might use this information for optimization,
> 
> I don't think that this is a good argument for making the code clearly harder to read (for those who maintain).

Ok, I dropped the commits that added the casts since they bring most of the visual noise and the compiler isn't going to optimize that code into something unexpected.

I don't think it's a good idea to tell the compiler something is const when it isn't though, compilers are getting better at taking advantage of these things, and bugs "caused by" compiler optimizations tend to be hard to track down.

PR #2582, opened 2026-01-26
update vignette sf2

kadyb:
I updated the default value of `stringsAsFactors` in the [vignette](https://r-spatial.github.io/sf/articles/sf2.html). Generally, `stringsAsFactors = FALSE⁠` is introduced since R version 4.0, but the documentation in `sf` says that since 4.1 -- is this expected?

### DIFF
--- a/R/read.R
+++ b/R/read.R
@@ -50,7 +50,7 @@ set_utf8 = function(x) {
 #' @param promote_to_multi logical; in case of a mix of Point and MultiPoint, or
 #'   of LineString and MultiLineString, or of Polygon and MultiPolygon, convert
 #'   all to the Multi variety; defaults to \code{TRUE}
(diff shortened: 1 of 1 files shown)


kadyb:
```
options(stringsAsFactors = TRUE)
#> Warning message:
#> In options(stringsAsFactors = TRUE) :
#>   'options(stringsAsFactors = TRUE)' is deprecated and will be disabled
```

It seems that the second method needs to be removed from the description?

edzer:
Thanks!

kadyb:
@edzer, shouldn't we remove this if it is deprecated (https://github.com/r-spatial/sf/pull/2582#issuecomment-3802031304)?

https://github.com/r-spatial/sf/blob/4c7b875a7399df532b45c5358966877a5933d748/vignettes/sf2.Rmd#L130-L135

PR #2583, opened 2026-01-28
CSLConstList compat for GDAL 3.1.3

mdsumner:
GDAL 3.13 requires  `CSLConstList` for `GetMetadata`

https://github.com/OSGeo/gdal/pull/13658

```R
library(sf)
#Linking to GEOS 3.12.1, GDAL 3.13.0dev-7924dc998e, PROJ 9.8.0; sf_use_s2() is TRUE
```

🙏


edzer:
Thanks!

PR #2584, opened 2026-01-28
vctrs methods for sf and sfc

edzer:
See #2568 for motivation

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,11 +2,11 @@
 
 * `st_cast.sfc()` deals with zero-length objects; #2584
 
-* rewrite/migrate `vctrs` methods for `sf` and `sfc`; #2568, #2584, w. support from @DavisVaughan
+* rewrite/migrate `vctrs` methods for `sf` and `sfc`; #2568, #2584, w. help from @DavisVaughan
 
 * `st_agr()` and `st_agr<-()` better handle multiple geometry columns
 
-* for an `sfc` object `x`, `x[0]` retains the class of `x`; reorganize `vctrs` methods; #2568
+* for an `sfc` object `x`, `x[0]` retains the class of `x`; #2568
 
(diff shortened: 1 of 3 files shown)


DavisVaughan (approved):
Few more requests to round things out. I'll also put up a PR on the vctrs side that removes all of our methods to make sure everything is still working as expected.

DavisVaughan:
If I install `pak::pak("r-lib/vctrs#2143")` and then come back to sf and run the tests in `test-tidyverse-vctrs.R`, then they all pass except for `"vec_locate_matches() works with sfc vectors"` because that one requires the `vec_proxy_order()` method mentioned above.

DavisVaughan:
I'm not totally sure yet, but I _think_ a reasonable plan is to let sf go to CRAN with these methods, which should immediately start to override the ones conditionally registered in vctrs. Then we wait like 3-6 months or so before removing them on the vctrs side, which hopefully gives most people that care a chance to go ahead and update to the latest version of sf before we yank out the vctrs methods from under them (i.e. if they are on new vctrs without the methods, and old sf without the methods, then things won't work).

We'd probably keep this error message that encourages people to update to the latest version of sf if we detect that people are trying to use new vctrs with old sf:

https://github.com/r-lib/vctrs/pull/2143/changes#diff-fbc888a8d95ca11d7f17e5eaa27a08a3d3cb183803ae509deea6aa51f61183fbR172

After a year or so we can remove that from vctrs too.

edzer:
Thanks, great work and good suggestions! I think we can fill `vec_proxy_order` with something more meaningful, after all it allows a `vec_sort` which is not of much use this way; I think we could do points < lines < polygons, and within those groups sort on number of points (in a `MULTIPOINT`), line length, and polygon area. No priority, and I think I can figure that out.

edzer:
I've started revdep checks, with the CRAN versions of vctrs and rlang; that'll take a few days.

PR #2615, opened 2026-08-13
Use getExportedValue to avoid pulling private functions

MichaelChirico:
`getNamespace()` will include un-exported symbols; my understanding is this is intended to work on the public API of {s2}, so `getExportedValue()` will better ensure that {sf} obeys the export contract of {s2}. (it's also IMHO more readable this way)


edzer:
Thanks, Michael!

Issue #2620, opened 2026-09-04
`st_coordinates` ignores `st_precision`, is this intended?


gustavo-tm:
**Describe the bug**
When running `st_precision` and then `st_coordinates`, the returned coordinates are not rounded.

**To Reproduce**
``` r
library(sf)

point  <- st_sfc(st_point(c(0.42,1.42)))

round(st_coordinates(point), 1) == st_coordinates(st_set_precision(point, 1)) 

         X     Y
[1,] FALSE FALSE
```

**Additional context**
Related to https://github.com/ropensci/osmextract/pull/325

Essentially, the input data has a 1e-7 native precision, but when importing the data there is some unintended noise on other decimal points. Using `st_set_precision` works for snapping the points to the correct grid for operations such as `st_join`, allowing for perfect joining between points. However, `st_coordinates` seem to ignore the precision tag.

edzer:
Did you read the `Details` section of the documentation of `st_precision()`?

gustavo-tm:
I did read it. The difficulty I had is that Details describes precision in terms of where it's applied internally `st_as_binary` / `CPL_write_wkb` and I didn't know which sf functions round-trip through WKB. I expected `st_coordinates` to be one of them, and the documentation of `st_coordinates` itself doesn't say either way.

A sentence in either help page would have saved me the confusion. But you know the intended audience better than I do, feel free to close if you think it's already clear enough.

edzer:
It says that the help page of `st_as_binary` contains an example of a roundtrip of sfc geometries through WKB, in order to see the rounding happening to R data." But I will add an example here as well.

agila5:
Hi Edzer! 

I'm writing an additional message here since I would like to ask a further question. I understand that setting `precision > 0` on a `sfc`/`sf` object does not modify its coordinates and the rounding is applied only before passing the data to GEOS routines. However, in my opinion, it wouldn't be unreasonable to think that such setting should propagate to other sf functions, such as `st_coordinates()`, especially because the default precision is 0 and, thus, setting `precision > 0` is a deliberate choice of a user. What do you think? 

I understand that this is possibly a big change with a lot of unforeseen consequences and I'm fine with the current behaviour. If you want to explore the idea (and its consequences) I would be happy to work on a PR focusing on `st_coordinates()`.

edzer:
Maybe by adding an argument `round = TRUE`, false by default, to `st_coordinates()`?

PR #2622, opened 2026-09-08
Add round argument to st_coordinates

agila5:
See the discussion in #2620

### DIFF
--- a/R/sfc.R
+++ b/R/sfc.R
@@ -497,8 +497,8 @@ typed_empty = function(cls, ncol = 2, dim = "XY") {
 #' retrieve coordinates in matrix form
 #' @param x object of class sf, sfc or sfg
 #' @param round logical; if `TRUE` and input object has a `precision` attribute
-#'   greater than 0, then its coordinates are rounded to the chosen precision
-#'   level before building the output matrix. See also `st_precision()`. 
+#'   different from 0, then its coordinates are rounded to the chosen precision
+#'   level before building the output matrix. See also `st_precision()`.
 #' @param ... ignored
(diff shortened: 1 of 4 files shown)


R/sfc.R
```diff
@@ -506,18 +509,32 @@ typed_empty = function(cls, ncol = 2, dim = "XY") {
...
+st_coordinates.sfg = function(x, round = FALSE, ...) st_coordinates(st_geometry(x), round = round)
 
 #' @export
-st_coordinates.sfc = function(x, ...) {
+st_coordinates.sfc = function(x, round = FALSE, ...) {
 	if (length(x) == 0)
 		return(matrix(nrow = 0, ncol = 2))
+	
+	# See the discussion in #2620
+	if (round && st_precision(x) > 0) {
```
edzer: please check equality to zero; negative precision does a double -> float "rounding".

R/sfc.R
```diff
@@ -506,18 +509,32 @@ typed_empty = function(cls, ncol = 2, dim = "XY") {
...
+		x = st_as_sfc(st_as_binary(x, precision = st_precision(x)))
```
edzer, line 539: this drops the crs of x, and its precision. You may also want to set `recompute_bbox=TRUE` as this will no longer match; `st_as_sfc.WKB` then needs to be modified to pass ... on to `st_sfc()`.

R/sfc.R
agila5: Will do, sorry!

R/sfc.R
agila5: > this drops the crs of x, and its precision. You may also want to set recompute_bbox=TRUE as this will no longer match

But is this really relevant? More or less, the following operations in `st_coordinates` extract the coordinates of `x` (via `unlist`, `coord_2`, and so on) and, based on my understanding, they ignore CRS, precision, and bbox

edzer: you're right, sorry!

agila5:
Hi Edzer. I implemented the changes requested except the following sentence since I'm not sure what I should modify: 

> st_as_sfc.WKB then needs to be modified to pass ... on to st_sfc().

PR #2624, opened 2026-09-16
Use GEOSGridIntersectionFractions_r() directly in st_interpolate_aw

edzer:
`st_interpolate_aw()` is for area-weighted interpolation, estimating attribute values from geometry set A to an arbitrary geometry set B, by using averaging (intensive variables) or summing/redistributing (extensive variables). The naive implementation computes all A-B intersections, and uses their areas for weighting. Since GEOS 3.14, `GEOSGridIntersectionFractions_r` (migrated from @dbaston's  [exactextracter](https://github.com/isciences/exactextractr)) computes such weights much more efficiently in case A or B is (regularly) gridded; and tbh that is probably a very common, if not dominant, use case. The current PR implements this both for `x` (source geoms) being a regularly gridded `stars` object, as well as `to` (target geoms).

The  `st_interpolated_aw.stars()` has been added to `sf` in the current CRAN release for that reason, and needs to be removed from `stars` (which happened in https://github.com/r-spatial/stars/commit/50ad7e859cab09e8e38a10f3ebc132a1643d9bdc).

### DIFF
--- a/src/geos.cpp
+++ b/src/geos.cpp
@@ -38,6 +38,9 @@
 # if GEOS_VERSION_MINOR >= 12
 #  define HAVE312
 # endif
+# if GEOS_VERSION_MINOR >= 14
+#  define HAVE314
+# endif
 #else
 # if GEOS_VERSION_MAJOR > 3
 #  define HAVE340
@@ -49,6 +52,8 @@
 #  define HAVE310
 #  define HAVE3101
 #  define HAVE311
+#  define HAVE312
+#  define HAVE314
 # endif
 #endif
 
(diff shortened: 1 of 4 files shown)


edzer:
@dbaston when doing so, I discovered that when calling `GEOSGridIntersectionFractions_r`  repeatedly (once for every geometry), I had to set the `float` output buffer values all to zero before every call; cells not intersecting seem to keep their old value. it would be good if this went into the docs, or got fixed.

src/geos.cpp
```diff
@@ -1609,3 +1614,28 @@ Rcpp::List CPL_line_interpolate(Rcpp::List lines, Rcpp::NumericVector dists, boo
...
+Rcpp::NumericVector CPL_grid_intersection_fractions(Rcpp::NumericVector p, Rcpp::List geoms) {
+	GEOSContextScope hGEOSCtxt;
+	int dim = 2;
+	unsigned ncell = p[4] * p[5];
+	std::vector<GeomPtr> g = geometries_from_sfc(hGEOSCtxt, geoms, &dim);
+	std::vector<float> buf(ncell); // nx * ny * n.geoms
+	Rcpp::NumericVector ret(ncell * g.size());
+	for (int i = 0; i < (int) g.size(); i++) {
+		for (int j = 0; j < ncell; j++)
+			buf[j] = 0.0; // initialize, needed for i > 0!
```
edzer, line 1628: @dbaston found out the hard way

dbaston:
> cells not intersecting seem to keep their old value

It's actually cells outside the geometry's _bounding box_ that keep their old value. This behavior is wrong, and a fix is in https://github.com/libgeos/geos/pull/1532

Still, you may want to adjust your usage to crop the raster down to each bbox before the call. Otherwise, a lot of effort is spent copying zeros around.

src/geos.cpp
```diff
@@ -1609,3 +1614,28 @@ Rcpp::List CPL_line_interpolate(Rcpp::List lines, Rcpp::NumericVector dists, boo
...
 	return ret;
 }
+
+// [[Rcpp::export(rng=false)]]
+Rcpp::NumericVector CPL_grid_intersection_fractions(Rcpp::NumericVector p, Rcpp::List geoms) {
+	GEOSContextScope hGEOSCtxt;
+	int dim = 2;
+	unsigned ncell = p[4] * p[5];
+	std::vector<GeomPtr> g = geometries_from_sfc(hGEOSCtxt, geoms, &dim);
+	std::vector<float> buf(ncell); // nx * ny * n.geoms
```
dbaston:
```suggestion
	std::vector<float> buf(ncell); // nx * ny
```

